PR #33, opened 2015-10-20
Signin bidder via github

baccigalupi:
* User authentication creates a bidder user
* saves the user id to session and stops using github uid in session
* changes database table to be more agnostic
* edit and create action for updating the current user


baccigalupi:
Also, I sit next to Maya who gave me a tutorial on using the style guide. It turns out we don't need to add so much markup. See the edit user form!

adelevie:
Noting here that we need `bundle exec rake db:migrate` to get this running.

adelevie:
@baccigalupi, is there any system for restricting roles? It seems like there might be one, but I can't figure it out.

adelevie:
I added two small commits:
- Update the seeds.rb file to account for the Bidders table name change
- Update an anachronism from the database.yml file

baccigalupi:
@adelevie This commit is not about roles. Just the bidder login. Next story deals with roles and in particular the Admin role getting routed differently on sign in.

adelevie:
@baccigalupi can you rebase against `master` which now has a working travis.yml file? This will get tests to properly run.

adelevie:
:dragon_face: rebased :dragon_face:

adelevie:
LGTM

PR #35, opened 2015-10-20
Signin admins

baccigalupi:
This work is on top of the previous pull request for creating and editing bidders on authentication.

This is the first of likely many features that differentiate users and provide different flows. This does not force admins to edit their DUNS/SAM id after logging in.


adelevie:
:camel: rebased :camel:

PR #36, opened 2015-10-20
Code climate badge

baccigalupi:
This one is straight-forward, except that I haven't been able to verify the .codeclimate.yml, exclusion works.

Also, we should note that CodeClimate hates on Sinatra, because it based on having methods outside of classes. This is considered a smell because it is hard to get those methods into a class and hard to refactor them well.

Current scoring puts the App at an A rating, but when I added more methods with actual logic it balked and tanked the code at a D. I think this is a real measure of the code being harder to work with and not just an annoyance in dealing with CodeClimate.


adelevie:
rebased

PR #41, opened 2015-10-20
Creating `config/` path

blacktm:
Active Record will look for `database.yml` in `config/` automatically. Plus, we’ll add more config stuff in the future, like for production servers: https://devcenter.heroku.com/articles/deploying-rails-applications-with-the-puma-web-server#config


baccigalupi:
Looks great. This is the convention expected by the gem, so WIN! :+1:

PR #42, opened 2015-10-20
Signed in header

baccigalupi:
This is for this issue: https://github.com/18F/mpt3500/issues/31

I reduced scope, because CSS sucks, and I don't know our framework well enough to lineup an image/icon.

This is really only one commit:  https://github.com/18F/mpt3500/commit/92cb99d170c2ab6095e89e30c145d9f1b6759e52

BUT, all my stories depend on each other. So the commit before this is https://github.com/18F/mpt3500/issues/23


baccigalupi:
@adelevie I squashed the two commits that were really in this branch and then cherry-picked it on to a new branch on top of master. Going to push this here, and then use it to update the Rails app code.

PR #44, opened 2015-10-21
Ensure /users/:id/edit works

adelevie:
- Using Rails form helpers for users/:id/edit

### DIFF
--- a/app/controllers/users_controller.rb
+++ b/app/controllers/users_controller.rb
@@ -19,6 +19,6 @@ class UsersController < ApplicationController
   private
 
   def user_params
-    params.require(:user).permit(:duns_id)
+    params.require(:user).permit(:duns_number)
   end
 end
--- a/app/controllers/users_controller.rb
+++ b/app/controllers/users_controller.rb
@@ -10,11 +10,15 @@ class UsersController < ApplicationController
     @user = User.find(params[:id])
     raise AuthorizationError if current_user != @user
 
-    @user.update_attributes({
-      sam_id:   params[:sam_id],
-      duns_id:  params[:duns_id]
-    })
+    @user.update_attributes(user_params)
 
     redirect_to '/'
+  end
+
+
+  private
+
+  def user_params
+    params.require(:user).permit(:duns_id)
   end
 end
--- a/app/controllers/users_controller.rb
+++ b/app/controllers/users_controller.rb
@@ -10,7 +10,7 @@ class UsersController < ApplicationController
     @user = User.find(params[:id])
     raise AuthorizationError if current_user != @user
 
-    @user.update_attributes(user_params)
+    @user.update(user_params)
 
     redirect_to '/'
   end
--- a/app/views/users/edit.html.erb
+++ b/app/views/users/edit.html.erb
@@ -1,13 +1,9 @@
-<h2>Please enter your SAM and/or D.U.N.S id</h2>
+<h2>Please enter your DUNS Number:</h2>
 
-<form action="/users/<%= @user.id %>" method='POST'>
-  <input type='hidden' name='_method' value='put'>
-  <label for="sam_id">SAM id:</label>
-  <input id="sam_id" name="sam_id" type="text" value="<%= @user.sam_id %>">
+<%= form_for @user, method: :put do |f| %>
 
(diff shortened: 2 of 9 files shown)


adelevie:
rebased

adelevie:
Note to myself tomorrow:
in `users_controller_spec.rb`, change `get :update, {id: user.id, duns_id: '222'}` to nest inside of `:user`. The form built with the rails helpers uses this nesting and the tests should be consistent with this.

app/controllers/users_controller.rb
```diff
@@ -10,11 +10,15 @@ def update
...
+    @user.update_attributes(user_params)
 
     redirect_to '/'
   end
+
+
+  private
+
+  def user_params
+    params.require(:user).permit(:duns_id)
```
baccigalupi: I know this is a rails thing, but for me it is a big smell to have private methods in your controllers. So, I am not a fan. My style would be to keep it explicit ... also makes it easier to extract into a class later. The way you did it is the rails way though, so this is just a nudge. You can keep it as is.

db/seeds.rb
```diff
@@ -20,16 +20,15 @@
 
 # github_id can be discovered using the github api:
 #
```
baccigalupi: Why? And can we just delete instead of commenting out?

baccigalupi:
:+1: Some comments, but it looks good

baccigalupi:
Also, thanks for the squashing! Very easy to review and understand

adelevie:
@baccigalupi I removed the commented-out code from the seeds file. It's a little less necessary/jankier with the `User`-saving code you added. E.g. a `User` instance is created upon authentication. Having a pre-baked `User` from seeds sort of deviates from that flow.

adelevie:
> Also, thanks for the squashing! Very easy to review and understand

Love to squash. h/t to @jgrevich for teaching me back in the day.

app/controllers/users_controller.rb
jessieay: woah! not having private methods in a controller would be pretty challenging. But would be fun to try!

app/controllers/users_controller.rb
```diff
@@ -10,11 +10,15 @@ def update
     @user = User.find(params[:id])
     raise AuthorizationError if current_user != @user
 
-    @user.update_attributes({
-      sam_id:   params[:sam_id],
-      duns_id:  params[:duns_id]
-    })
+    @user.update_attributes(user_params)
```
jessieay: this isn't important but thought I'd share a recent learning: `update_attributes` is actually an alias for `update` http://api.rubyonrails.org/classes/ActiveRecord/Persistence.html#method-i-update

so you could just call `@user.update(user_params)` here

app/views/users/edit.html.erb
```diff
@@ -1,13 +1,9 @@
...
-  <label for="sam_id">SAM id:</label>
-  <input id="sam_id" name="sam_id" type="text" value="<%= @user.sam_id %>">
+<%= form_for @user, method: :put do |f| %>
 
-  <label for="duns_id">D.U.N.S id:</label>
-  <input id="duns_id" name="duns_id" type="text" value='<%= @user.duns_id %>'>
-
-  <input type='submit' value='Save'>
-</form>
+  <%= f.label :duns_id, 'DUNS Number' %>
```
jessieay: since this is called "DUNS Number" in the view I wonder if it should be called `duns_number` in the database. When I see `id`, I always assume it is a unique id for another model. (started looking for a `duns` model here). Sometimes using `xyz_id` can make sense anyway, but since it seems like users will be calling it "Duns number" anyway, why not call it that on the backend anyway!

naming is fun! :8ball:

jessieay:
@adelevie I love squashing too! I think it makes the git history for a project so much more enjoyable / useful. Had a long convo about this in the git slack channel, though, and it seems us squashers are in the minority.

Left a couple of comments, but code looks good to me! :shaved_ice:

adelevie:
> it seems us squashers are in the minority.

Ask them to rebase a 25+ commit PR and then ask their opinion on squashing :microphone: :arrow_down:

blacktm:
:shipit:

jessieay:
@adelevie :pray:

PR #45, opened 2015-10-21
Auction list view aka. Home page uses real data

baccigalupi:
The old Sinatra app was using a table to display content. This
        changes the dynamic view to use the usa-grid system. It does not
        look exactly the same as the table, so product ... please check.

        The auction controller loads all the auctions. Product needs to
        supply some details about what to show and not show. The query
        preloads bids to reduce number of queries, and then maps the
        records into a presenter for the view.

        Logic about who has the winning bid is in that auction
        presenter and there are test. Product should checkout the test
        and make sure it makes sense.

@adelevie There is a template candidate that you can take apart and use for other view realization work

### DIFF
--- a/app/assets/stylesheets/main.scss
+++ b/app/assets/stylesheets/main.scss
@@ -1,3 +1,37 @@
+$color-green: #2e8540;
+
+.issue-list-item {
+  .issue-icon {
+    padding-top: 0.75rem;
+  }
+
+  .issue-title {
+    margin-top: 1rem;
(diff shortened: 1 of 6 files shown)


adelevie:
rebased

adelevie:
screen shot 2015-10-21 at 4 46 47 pm

Love it! LGTM unless you want me to wait for any new parts.

adelevie:
:ship:

PR #48, opened 2015-10-22
Use rails12factor gem to ensure assets are served

adelevie:


baccigalupi:
Can we add asset compilation to the deploy build process? It is not a good idea to check in the compiled assets because they change so often. It bloats the git repository, but also keeps the change log from clearly showing what has changed.

adelevie:
This gem might already handle that. I'll delete the contents of public/assets, and re-push to see if it works.

adelevie:
It works.

blacktm:
:ship:

PR #53, opened 2015-10-22
Add rails admin

adelevie:
- Uses `rails_admin` to implement the admin panel
- Hooks into our current auth system (looks for presence of `current_user` and checks `current_user.github_id` against the whitelist set in `admins.yml`)
- Uses the `paper_trail` gem to provide change tracking to our models.
- Uses the `paper_trail` gem's integration with `rails_admin` to make said changes viewable in the admin panel.
- Adds a log out link in the admin panel navigation.

To try this out locally:

- start up the rails server
- visit /logout to make sure you are logged out
- visit /admin and you should see 'Not authorized' as the response
- visit /auth/github to login with github credentials
- visit /admin and you will be in the admin panel

Maybe this needs tests. `rails_admin` and `paper_trail` are pretty well-tested, established gems and we probably shouldn't have redundant tests. But I would be interested in testing the integration points between our application and these libraries. @baccigalupi, anyone, any thoughts?


adelevie:
screen shot 2015-10-22 at 6 59 15 pm

adelevie:
Closing in favor of building an admin panel. See https://github.com/18F/micropurchase/issues/54 for user stories.

Issue #51, opened 2015-10-22
As an authorized user, I want a "Logout" link in the header.


blacktm:
Given that I am authorized with GitHub,
I want to see my avatar in the Header,
and have a link to "Logout".

![screen shot 2015-10-22 at 5 39 54 pm](https://cloud.githubusercontent.com/assets/887429/10679202/f4cd98a6-78e4-11e5-91fd-806ed639c1ea.png)

blacktm:
Completed in #57

Issue #50, opened 2015-10-22
As a site visitor, I want a "Login" link in the header.


blacktm:
Given that I'm not registered / have an account,
I see a login link in the header,
which takes me to the registration / "Getting Started" page,
so that I can authorize with GitHub.

Header at home screen:
![screen shot 2015-10-22 at 5 39 34 pm](https://cloud.githubusercontent.com/assets/887429/10679075/310753bc-78e4-11e5-9e52-0ae2f50987b8.png)

Registration / Getting Started page:
![screen shot 2015-10-22 at 5 40 52 pm](https://cloud.githubusercontent.com/assets/887429/10679105/5dca631c-78e4-11e5-8dcf-2b657fc94d38.png)

blacktm:
Completed in #57

PR #57, opened 2015-10-23
Restored functionality of the login/logout links.

adelevie:
- Restored the login/logout buttons at the top of the layout
- Added a `/login` route per @blacktm's UI sketches (`LoginsController#index`)


adelevie:
Fixes https://github.com/18F/micropurchase/issues/51 and https://github.com/18F/micropurchase/issues/50.

app/assets/javascripts/logins.js
```diff
@@ -0,0 +1,2 @@
+// Place all the behaviors and hooks related to the matching controller here.
+// All this logic will automatically be available in application.js.
```
baccigalupi: Can we can probably get rid of these generated empty files?

app/controllers/application_controller.rb
```diff
@@ -1,7 +1,8 @@
 class ApplicationController < ActionController::Base
   # Prevent CSRF attacks by raising an exception.
   # For APIs, you may want to use :null_session instead.
-  protect_from_forgery with: :exception
+  #protect_from_forgery with: :exception
```
baccigalupi: I think we probably want to keep this. Was this commented out for debugging, or is there another issue going on?

app/helpers/logins_helper.rb
```diff
@@ -0,0 +1,2 @@
+module LoginsHelper
+end
```
baccigalupi: This too can probably go

app/assets/javascripts/logins.js
adelevie: Yup!

app/controllers/application_controller.rb
adelevie: good catch ... this was an artifact of debugging.

app/helpers/logins_helper.rb
adelevie: on it.

app/views/users/confirm.html.erb
```diff
@@ -0,0 +1,11 @@
+<h2>Complete your Account</h2>
```
baccigalupi: users#edit is probably a better REST representation.

spec/helpers/logins_helper_spec.rb
```diff
@@ -0,0 +1,14 @@
+require 'rails_helper'
```
baccigalupi: I would ditch this file one too.

adelevie:
rebased

PR #58, opened 2015-10-23
Puts in some basic dynamic views into /bids

baccigalupi:
According to the mockups, this should include information
        about what your bid is on each auction and what the winning bid
        is.

        Feeling like all the auctions should show this information so
        you can see wherever you are looking at an individual or list
        of auctions, you can see what your bid is.

        This is just a fast get something dynamic out the door commit
        though. It has the intention of just removing static data.


adelevie:
:+1:

PR #69, opened 2015-10-27
Added Code Climate badge

adelevie:


adelevie:
rebased

baccigalupi:
I think you are already working on that token. Seems like we should omit the coverage directory in .gitignore.

blacktm:
Looks good. Any last comments before merging, @adelevie/@baccigalupi?

Issue #20, opened 2015-10-19
As an admin, I want to create new issue/auctions


baccigalupi:
Given I am logged in as an admin
And I go to the admin/issues/new page
Then I should see a form for entering new auctions into the app
When I complete the form
Then I will have created a new auction with the details provided

Given I am a non-admin signed in user
When I try to go to the new issues page
Then I will be redirected to home

blacktm:
@adelevie and I talked about this, and admins should probably be able to do CRUD operation on any model.

PR #70, opened 2015-10-27
Admin can see and create new auctions

baccigalupi:
This feature is broken into two commits:

The first adds an protected admin section that has a slightly different look. That commit lists and shows auctions. The auction show view includes all the bids. It isn't very intuitive, but the green issue icon only shows up when the auction is live (within dates).

The second commit focuses on creating auctions. This is the first time we are parsing times from the user which lead to some issues. I made the assumption that time in the UI would be in Washington DC time. Storing time in the DB as a best practice should be in UTC. ActiveSupport does the work on timestamps, but on the way in when looking at start and end times, we need to use Chronic to parse the times in eastern time. On the way out to the view, times are converted back to DC time via a little presenter.


baccigalupi:
Also, it addresses this, with some extras: https://github.com/18F/micropurchase/issues/20

adelevie:
Awesome, reviewing now...

app/views/admin/auctions/new.html.erb
```diff
@@ -0,0 +1,20 @@
+<h1>Create a new auction</h1>
+
```
adelevie: Is there a reason `issue_url` and `github_repo` aren't included in the form? Happy to add them in if we end up wanting those fields.

Also thinking I could have named those a little better. Perhaps `github_repo` -> `repo_url`?

app/models/admin_create_auction.rb
```diff
@@ -0,0 +1,51 @@
...
+  attr_reader :auction
+
+  def perform
+    @auction = Auction.create(attributes)
+  end
+
+  private
+
+  def attributes
+    {
```
adelevie: Same question applies from my other comment on this PR. `issue_url` is not listed in this hash, but perhaps it should?

app/models/admin_create_auction.rb
```diff
@@ -0,0 +1,51 @@
...
+      title: title,
+      description: description,
+      github_repo: github_repo,
+      start_datetime: start_datetime,
+      end_datetime: end_datetime,
+      start_price: start_price
+    }
+  end
+
+  [:title, :description, :github_repo].each do |key|
```
adelevie: If we do add `issue_url` to `#params`, this would also need to be updated.

app/models/presenter/bid.rb
```diff
@@ -0,0 +1,40 @@
...
+
+    def bidder_name
+      bidder.name || null
+    end
+
+    def null
+      Null::NULL
+    end
+
+    class Null
```
adelevie: The use of this class in `Presenter::Auction` pretty neat!

app/views/admin/auctions/index.html.erb
```diff
@@ -0,0 +1,29 @@
...
+  <div class='usa-width-one-whole'>
+    <div class='usa-width-one-sixth'>
+      <% if auction.available? %>
+        <%= image_tag "issue-icon", class: 'issue-icon' %>
+      <% else %>
+        <span class=''>not live</span>
+      <% end %>
+    </div>
+    <div class='usa-width-one-fourth'>
+      <b>Starts:</b> <%= Presenter::DcTime.convert(auction.start_datetime).to_s(:long) %>
```
adelevie: Small thing, but perhaps the DC-converted times should be methods on `Presenter::Bid`?

e.g.

```ruby
module Presenter
  class Auction
   # ...
   def start_dc_datetime
     Presenter::DcTime.convert(start_datetime).to_s(:long)
   end
   # ...
   end
end
```

spec/models/admin_create_auction_spec.rb
```diff
@@ -0,0 +1,148 @@
...
+            description: 'description',
+            github_repo: 'github url',
+            start_datetime: 'Nov 3, 2015',
+            end_datetime: '11/10/2015',
+            start_price: 3500.01
+          }
+        }
+      }
+
+      it 'drops the price down to 3500' do
```
adelevie: I'm a little bit confused by these few specs. The description says the price will be dropped to `3500`, but the assertion (which passes!) says `3400.99`.

1) Is there an incrementor pushing the price down `100.01`?
2) Should the spec descriptions be updated?

adelevie:
@baccigalupi I'm done reviewing. Apologies for the wall of line comments. It's in pretty great shape and I am definitely learning a lot about good Rails practices here. The only issues were questions about the inclusions of some field names and how some of the specs work (they're all in the line comments).

baccigalupi:
@adelevie thanks for the feedback. The omission of the github_repo field was unintentional. The issue_url I just didn't understand what it was on a product level. So, I think the issue_url is a link to the github (or other??) issue for the work?

Will put together a fix for these things.

adelevie:
@baccigalupi The idea is that there's a repo (e.g. https://github.com/18f/calc) and the specific issue (e.g. https://github.com/18f/calc/issues/255).

I think the issue url is much more important than the repo url, but I thought it would be helpful to directly link back to a project page so the potential bidder could get more context. I'd say the issue url is absolutely needed while the repo url is negotiable.

adelevie:
:ship:

Issue #72, opened 2015-10-28
Add feature specs to exercise all views and articulate user flows


baccigalupi:
Currently there are controller specs that test database changes and which view is rendered. There have been easy refactoring generated bugs getting through because the views are not being exercised in the tests yet. Also, the typical user story flows are not being captured in the unit tests.

This chore aims to capture user flows and make sure all the views get exercised in the test suite. More than likely controller specs will be eliminated in this process.

adelevie:
Fixed in #73.

PR #73, opened 2015-10-29
Feature specs

baccigalupi:
This is two commits again. The first adds feature specs for the user flows that I could think through. The second does a pass over the controller specs. I expected to remove more specs, but it turned out there was a little backfilling to do.

Writing these specs, controller and feature level points out some bad user experiences.

Addresses this issue: https://github.com/18F/micropurchase/issues/72

### DIFF
--- a/app/assets/stylesheets/main.scss
+++ b/app/assets/stylesheets/main.scss
@@ -170,11 +170,13 @@ footer {
 .header-account span {
   position: relative;
   top: -8px;
+  right: 10px;
 }
 
-.header-account button {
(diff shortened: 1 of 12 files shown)


adelevie:
Thanks, @baccigalupi! I won't be able to review until later today or possibly early Friday. Big :+1: for exhaustive feature specs for user flows.

baccigalupi:
No rush, just trying to get through a few stories before I roll off

adelevie:
:ship: thanks @baccigalupi!

Issue #75, opened 2015-10-29
As an admin, I want to edit existing auctions


baccigalupi:
Currently you can create auctions, but no edit them.

adelevie:
Fixed in #76

PR #76, opened 2015-10-29
Admin can edit auction:

baccigalupi:
https://github.com/18F/micropurchase/issues/75

        Since all the same data munging needed to happen in both create
        and update, I extracted out a parser. Probably not the right
        name for the object. There is a CreateAuction and UpdateAuction
        object but these are pretty light weight. Might be that they
        should go away or a repository object to handle the upsert
        stuff.

        Also unifies the form into one location instead of copying and
        pasting in both. Feature specs of course. No real controller
        specs since the only additional logic is with the flash
        generation.

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -23,7 +23,6 @@ group :test do
 end
 
 group :development, :test do
-  gem 'capybara'
   gem 'rspec-rails'
   gem 'capybara'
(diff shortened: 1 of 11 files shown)


adelevie:
:shipit:

Issue #74, opened 2015-10-29
Use U.S. Web Design Standards gem


blacktm:
See [gem repo](https://github.com/18F/us_web_design_standards_gem/pull/2) for details.

adelevie:
Closed by #77.

PR #77, opened 2015-10-30
Use U.S. Web Design Standards gem

blacktm:
This will remove the U.S. Web Design Standards asset files from the app and use the gem instead. This addresses #74.

This is ready for review.


adelevie:
Thanks @mbland and @blacktm!

Issue #79, opened 2015-11-10
Change the bidding increments


vzvenyach:
As a bidder, I want to know what the minimum bid increment is, so that I can bid appropriately.

allalala:
@adelevie have we clarified that bidding must occur in 1-dollar increments (on, say, the bid detail or the policy documentation page)?

mtorres253:
This is done

PR #80, opened 2015-11-12
Disallow bids with penny-sized increments.

adelevie:
PlaceBid prevents increments that are less than one dollar.

Update other specs to no longer user increments less than one dollar.

Update PlaceBid and AuctionParser specs to reflect 3500.00 bid limit


adelevie:
Fixes https://github.com/18F/micropurchase/issues/79.

Issue #22, opened 2015-10-19
As a bidder, I want a clear indication of whether my bid on an issue is the winning bid


baccigalupi:
Given I am an authenticated and approved bidder
And I have posted a bid
And I am the lowest bid
Then I want to see on the issue detail page a clear indication that I am the current winner

blacktm:
This also implies users should be able to see all issues they have placed bids on.

![screen shot 2015-10-22 at 5 21 49 pm](https://cloud.githubusercontent.com/assets/887429/10678648/b2c4bd34-78e1-11e5-8592-6007d7fecf8e.png)

JoshData:
As a bidder on a particular auction, if I am not the winner, clear indication of that would be helpful as well --- because I don't remember what I bid.

JoshData:
Just for reference, this applies to `/`, `/my-bids`, `/auctions/{id}`, and `/auctions/{id}/bids`.

PR #82, opened 2015-11-12
Inform the currently winning bidder that they are the currently winning bidder.

adelevie:
Depends on https://github.com/18F/micropurchase/pull/81 (e.g. merge that first).


adelevie:
Fixes https://github.com/18F/micropurchase/issues/22.

PR #85, opened 2015-11-16
Bid history

adelevie:
`/auctions/:id/bids`

When the auction is over, display names of bidders:
screen shot 2015-11-16 at 2 40 47 pm

While the auction is still in progress, redact the names of the bidders:
screen shot 2015-11-16 at 2 41 55 pm

Added a link in bid listings to link to the bid history page:
screen shot 2015-11-17 at 3 19 36 pm


adelevie:
@blacktm or @baccigalupi this PR is ready for review.

ddotrollins:
@adelevie Small tweaks to show the winner on this page and also some responsive considerations.

screen shot 2015-11-17 at 5 06 20 pm

screen shot 2015-11-17 at 5 07 07 pm

adelevie:
Latest commit makes the page look like this:

screen shot 2015-11-18 at 10 24 13 am

ddotrollins:
Looks good on my end.

app/models/presenter/bid.rb
```diff
@@ -9,7 +9,11 @@ def bidder_duns_number
     end
 
     def bidder_name
-      bidder.name || null
+      if Presenter::Auction.new(auction).available?
+        '[Name witheld until the auction ends]'
+      else
+        bidder.name || null
```
jessieay: woah! I don't usually see `null` returned like this in Ruby -- in fact, I don't think it is a thing in Ruby (but definitely correct me if I am wrong!)

```ruby
[1] pry(main)> null
NameError: undefined local variable or method `null' for main:Object
from (pry):1:in `<main>'
[2] pry(main)> nil
=> nil
```

if `bidder.name` is nil, this will just return `nil` -- would that work for your use case?

app/views/auctions/_list_item.html.erb
```diff
@@ -22,5 +22,13 @@
     <% else %>
       <%= render partial: 'auctions/winning_bid', locals: {auction: auction} %>
     <% end %>
+    <p>
+      <%=
+        link_to(
+          pluralize(auction.bids.length, 'bids'),
```
jessieay: niiiiice

jessieay: I think somewhere there was a convo about doing `count` instead of `length

app/views/bids/index.html
```diff
@@ -0,0 +1,29 @@
...
+<table class="usa-table-borderless">
+  <thead>
+    <tr>
+      <th scope="col">Bidder</th>
+      <th scope="col">Amount</th>
+      <th scope="col">Date (Eastern Standard Time)</th>
+    </tr>
+  </thead>
+  <tbody>
+    <% @auction.bids.sort_by {|b| b.created_at}.reverse.each_with_index do |bid, i| %>
```
jessieay: What do you think about adding a scope / class method to `bids` called something like `most_recently_created_first` and using that instead of `@auction.bids.sort_by {|b| b.created_at}` -- that would be easier to read and you could also have a unit test for it!

app/views/bids/index.html
```diff
@@ -0,0 +1,29 @@
...
+      <tr>
+        <td><%= bid.bidder_name %></td>
+        <% if i == 0 %>
+          <td><%= number_to_currency(bid.amount) %> &#42;</td>
```
jessieay: so for the first bid you want to append w ` &#42;` but none others? why is that?

config/application.rb
```diff
@@ -31,5 +31,7 @@ class Application < Rails::Application
 
     # Do not swallow errors in after_commit/after_rollback callbacks.
     config.active_record.raise_in_transactional_callbacks = true
+
+    config.assets.debug = true
```
jessieay: perhaps this should go in an environment-specific config file? seems like this could slow you down http://artandlogic.com/2012/12/faster-rails-dev/

spec/controllers/bids_controller_spec.rb
```diff
@@ -67,6 +69,39 @@
...
+        expect(response).to render_template(:index)
+      end
+    end
+
+    context 'when logged out' do
+      before do
+        allow(controller).to receive(:current_user).and_return(nil)
+      end
+
+      it 'renders the template' do
```
jessieay: what do you think about not including a test for this? Since this controller is using Rails default template, it seems like this is testing Rails internals more than anything else

spec/controllers/bids_controller_spec.rb
```diff
@@ -67,6 +69,39 @@
...
+        allow(controller).to receive(:current_user).and_return(current_bidder)
+      end
+
+      it 'renders the template' do
+        get :index, auction_id: auction.id
+        expect(response).to render_template(:index)
+      end
+    end
+
+    context 'when logged out' do
```
jessieay: so people can see this view without being logged in? If that is the case, I think it would be helpful to say what http status code is returned for logged out users than to have a test for which template is returned

spec/controllers/bids_controller_spec.rb
```diff
@@ -67,6 +69,39 @@
...
+      end
+    }
+    context 'when logged in' do
+      before do
+        allow(controller).to receive(:current_user).and_return(current_bidder)
+      end
+
+      it 'renders the template' do
+        get :index, auction_id: auction.id
+        expect(response).to render_template(:index)
```
jessieay: what about using the status code in expectation instead?

app/views/bids/index.html
adelevie: We have an asterisk at the bottom of the table noting that the first row is the winning bid.

tl;dr, the first row is special.

config/application.rb
adelevie: Good call! I am going to turn this into a separate issue.

spec/controllers/bids_controller_spec.rb
```diff
@@ -67,6 +69,39 @@
     end
   end
 
+  describe '#index' do
+    let(:bids) {
+      [3400, 3000, 1800].map do |amount|
+        Bid.create({
+          auction_id: auction.id,
```
jessieay: can just do `auction: auction` :) (same for bidder below)

spec/features/bidder_interacts_with_auction_spec.rb
```diff
@@ -1,6 +1,34 @@
 require 'rails_helper'
 
 RSpec.feature "bidder interacts with auction", type: :feature do
+  scenario "Viewing auction list and navigating to bid history" do
+    create_current_auction
+
+    visit "/"
```
jessieay: what about `visit root_path` ?

app/models/presenter/bid.rb
adelevie: Good find. @baccigalupi, I think you wrote this ... any problems with `null` -> `nil`?

app/views/auctions/_list_item.html.erb
adelevie: I'll switch to `#count`. Thanks!

app/views/bids/index.html
adelevie: I'll do that. And the feature spec should help catch and regressions (it tests the visual order of the table rows).

spec/features/bidder_interacts_with_auction_spec.rb
```diff
@@ -85,4 +113,83 @@
...
   end
 
+  scenario "Viewing bid history for running auction" do
+    Timecop.scale(36000) do
+      create_running_auction
+    end
+    path = auction_bids_path(@auction.id)
+    visit path
+
+    # sort the bids so that newest is first
```
jessieay: what about removing this comments? These methods are pretty small/clean for reading without comments. Alternatively, if you wanted to make this all super easy to read, you could wrap these bits of code in private methods. 

so it would be `bids = newest_first(@action.bids)`

and then at the bottom of this spec file:

```ruby
def newest_first(bids)
  bids.sort_by {|bid| bid.created_at}.reverse
end
```

adelevie: I think I'm going to keep the comments. Otherwise, there's not an easy way to infer that a column is for "amount" (only that it is the 2nd column from the left). I will probably refactor the sorting, though.

app/views/bids/index.html
monfresh: Eventually, I would recommend moving this logic to a helper method. Logic in views is usually frowned upon.

spec/features/bidder_interacts_with_auction_spec.rb
```diff
@@ -85,4 +113,83 @@
...
+    bids = @auction.bids.sort_by {|bid| bid.created_at}.reverse
+
+    # ensure the table has the correct content, in the correct order
+    bids.each_with_index do |bid, i|
+      row_number = i + 1
+      unredacted_bidder_name = bid.bidder.name
+      bid = Presenter::Bid.new(bid)
+
+      # check the "name" column
+      within(:xpath, cel_xpath(row_number, 1)) do
```
jessieay: what about `expect_content_in_row_number(content: "[Name witheld until the auction ends]", row: 1)`

and defining 

```
def expect_content_in_row_number(args)
  within(:xpath, cel_xpath(row_number, args[:row_number])) do
     expect(page).to have_content(args[:content])
  end
end
```

jessieay: ^^ above would be reusable for expectations below, too!

spec/support/feature_helpers.rb
```diff
@@ -1,20 +1,56 @@
...
-    end_datetime: Time.now + 3.days,
+    end_datetime: end_datetime,
     title: 'Oh no, fix the world',
     description: 'it is broken!',
     issue_url: 'https://github.com/18F/calc/issues/255',
     github_repo: 'https://github.com/18F/calc'
   })
-  @bidder = User.create(github_id: 'uid', duns_number: 'duns')
+  @bidders = [
+    User.create(
```
jessieay: are you using factories in this project?

spec/support/feature_helpers.rb
```diff
@@ -33,3 +69,22 @@ def sign_in_bidder
...
+
+def show_page
+  # quickly view the page visited by Capybara
+  # requires the local server to be running
+  # for more info: https://coderwall.com/p/jsutlq/capybara-s-save_and_open_page-with-css-and-js
+  save_page Rails.root.join('public', 'capybara.html')
+  %x(launchy http://localhost:3000/capybara.html)
+end
+
+def expect_cel_to_have_content(page, row_number, column_number, content)
```
jessieay: this appears to be unused?

spec/features/bidder_interacts_with_auction_spec.rb
jessieay: oh I see! So when I saw the comment "check the amount column" I thought it was referring to the DATABASE column. Thoughts on how that could be clarified either as a comment or code?

spec/features/bidder_interacts_with_auction_spec.rb
adelevie: I did have something like that! I got rid of it when I realized I'd also need `expect_content_not_in_row_number`. And then I thought, I could pass a block to a single function where the expectation could be called. It felt a little bit too close to reinventing parts of RSpec. Currently, a new set of :eyes: could see `within` and Google "rspec" within, and know immediately how the code here works.

spec/support/feature_helpers.rb
adelevie: I don't think so. We should.

spec/support/feature_helpers.rb
adelevie: Yeah that was the thing I decided to stop using. I'll delete.

app/views/bids/index.html
adelevie: So something like `content_for_first_row` could contain the `if` logic?

jessieay: you could do something like:


```ruby
def content_for_row(row_number, bid)
  if row_number == 0
    "#{number_to_currency(bid.amount)} &#42;"
  else
    number_to_currency(bid.amount)
  end
end
```

and then in the view:

```ruby
 <td><%= bid.bidder_name %></td>
 <td><%=content_for_row(index, bid)</td>
```

app/models/presenter/bid.rb
adelevie: Going to defer to @baccigalupi here and keep everything as is, but we can alter this later if needed.

app/views/bids/index.html
adelevie: I ended up packing the bid sorting inside of `Presenter::Auction#bids`.

app/views/bids/index.html
adelevie: Implemented near-verbatim.

spec/controllers/bids_controller_spec.rb
adelevie: I'll make paring down the tests another story.

spec/controllers/bids_controller_spec.rb
adelevie: Same as above.

spec/controllers/bids_controller_spec.rb
adelevie: For a new story.

spec/controllers/bids_controller_spec.rb
adelevie: Fiiiiiine :)

spec/features/bidder_interacts_with_auction_spec.rb
adelevie: For a new story.

PR #88, opened 2015-11-17
Automated security scan

adelevie:
The branch name is a bit of a misnomer. This pull request adds three types of automated scans:

- [Brakeman](https://github.com/presidentbeef/brakeman) static code analysis.
- [Hakiri](https://github.com/hakirisec/hakiri_toolbelt) Ruby/Rails version scan.
- [Hakiri](https://github.com/hakirisec/hakiri_toolbelt) Gemfile scan.

Interestingly, this brought up one high confidence security warning. If all goes well, the warning should cause the build to break inside Travis, with a helpful error message.


Gemfile
```diff
@@ -28,6 +28,8 @@ group :development, :test do
   gem 'capybara'
   gem 'byebug'
   gem 'pry'
+  gem 'brakeman', require: false
```
jessieay: brakeman is already used by Code Climate - are you using CC in this project? http://docs.codeclimate.com/article/160-glossary-oss-tools-used-by-code-climate

adelevie: We are using Code Climate, but for "reasons" I don't see any security scanning available. Here's our Code Climate page: https://codeclimate.com/github/18F/micropurchase.

jessieay: it's possible that you're using platform instead of code climate classic. They are "in the process of implementing" the brakeman engine for the new code climate platform (as opposed to classic) http://docs.codeclimate.com/article/304-platform-analysis-comparison

adelevie:
This can be safely merged. A subsequent PR, https://github.com/18F/micropurchase/pull/90, addresses the error that breaks the build.

spec/features/security_spec.rb
```diff
@@ -0,0 +1,24 @@
...
+  end
+
+  scenario "The Gemfile does not depend on vulnerable gems" do
+    expect(@gemfile_vulns).to have_content("No vulnerabilities found. Keep it up!")
+  end
+
+  scenario "The Ruby and Rails versions have no known (CVE) vulnerabilities" do
+    expect(@ruby_rails_vulns).to have_content("No vulnerabilities found. Keep it up!")
+  end
+end
```
afeld: Do these need to be run as part of the build? I thought Hakiri does it on their side anyway?

adelevie: I haven't registered with the Hakiri.com service. I'm just using their CLI and letting Travis handle the build breakage.

afeld:
:shipit: (needs manual merge)

adelevie:
Awesome, thanks!

monfresh:
As discussed in the hallway yesterday, I would not recommend running a separate service such as Brakeman or Hakiri from within an RSpec test. These services have their own separate commands that can be run separately, so tying security tests to a specific test framework prevents portability. By running the commands separately via a Rake task or as separate scripts to run on Travis allows for portability in case you end up switching to a different test framework, for example.

I equate Brakeman to other standalone services like Rubocop. Just as you wouldn't run Rubocop from within RSpec, I wouldn't run Brakeman from within RSpec either. If you want to make sure Brakeman runs as part of the local dev suite, I would create a shell script that runs `rspec spec`, then whatever else you want to run and then make a note in the README that devs should run `script/test` (or whatever you call your script). See this as an example: https://github.com/codeforamerica/ohana-api/blob/master/script/test

monfresh:
The other reason why I do not recommend doing this is that running Brakeman from running RSpec is most likely going to be slower than running it separately. We ran into this very issue on save-ferris. I was profiling our test suite (using RSpec's [profile_examples](https://www.relishapp.com/rspec/rspec-core/docs/configuration/profile-examples) option), and determined that the embedded Brakeman specs were the slowest of all our tests. By removing them from RSpec and running them separately, we ended up with a faster-running integration suite, while still benefiting from security checks.

adelevie:
@monfresh, I appreciate the time you took to review the code and to write detailed feedback. I also look to you as a person to learn from and to defer to on all things Rails. I have a different view than yours (though I am still thinking about this). Here are my thoughts:

By pulling these different pieces out of the test suite, we lose readable, automatically-generated documentation of _all things_ that are tested. `rspec -f d > somefile.txt` generates auditable, version-tracked documentation of the application. This documentation can be read by non-Rubyists, and even non-programmers. This,

```
Security (via Brakeman static code analysis)
  The site has zero Brakeman security warnings (FAILED - 1)
  The Gemfile does not depend on vulnerable gems
  The Ruby and Rails versions have no known (CVE) vulnerabilities
```

is readable to a larger audience than a `go` script.

In a compliance-driven environment, not just testing for compliance, but _continuously documenting_ what you are testing for a non-technical audience is one technique to keep us away from waterfall.

Addressing your other concerns below:

Switching test suites 1) is not happening anytime soon and 2) if it were happening, the Hakiri and Brakeman parts would be the easiest parts to switch over.

Re slowness, [RSpec tags](https://www.relishapp.com/rspec/rspec-core/v/3-4/docs/command-line/tag-option) could just as easily allow a dev to quickly run the faster specs. 

Finally, I don't think either approach is wrong, by the way. There's a lot of room for subjectivity for developer ergonomics.

adelevie:
rebased

Issue #89, opened 2015-11-17
Fix possible unprotected redirect


adelevie:
```
  1) Security (via Brakeman static code analysis) The site has zero Brakeman security warnings
     Failure/Error: expect(@brakeman_warnings.length).to eq(0), "Expected 0 security warnings, got: \n #{@brakeman_warnings.map(&:to_s)}."
       Expected 0 security warnings, got: 
        ["(High) Redirect - Possible unprotected redirect near line 3 in /Users/alandelevie/micropurchase/app/controllers/authentications_controller.rb: redirect_to(Authenticator.new(request.env[\"omniauth.auth\"], session).perform)"].
     # ./spec/features/security_spec.rb:14:in `block (2 levels) in <top (required)>'
```

More info: http://brakemanscanner.org/docs/warning_types/redirect/

PR #90, opened 2015-11-17
Fix redirect warning

adelevie:
This pull request depends on https://github.com/18F/micropurchase/pull/88 and fixes https://github.com/18F/micropurchase/issues/89.

Even after the fix, the warning is still reported by Brakeman. I am confident that the fix actually addresses the warning, and that now it is a false positive. I would like another set of :eyes: to confirm this.

I have added this warning to `config/brakeman.ignore`.

### DIFF
--- a/config/brakeman.ignore
+++ b/config/brakeman.ignore
@@ -1,0 +1,25 @@
+{
+  "ignored_warnings": [
+    {
+      "warning_type": "Redirect",
+      "warning_code": 18,
+      "fingerprint": "1f5e56ec432800fd1896435e9cd488f3c75047cdb1855add003acf99052a04f3",
+      "message": "Possible unprotected redirect",
+      "file": "app/controllers/authentications_controller.rb",
+      "line": 4,
+      "link": "http://brakemanscanner.org/docs/warning_types/redirect/",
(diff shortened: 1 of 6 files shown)


afeld:
The diff between this and #88:

https://github.com/18F/micropurchase/compare/brakeman-security-scan...fix-redirect-warning

adelevie:
@afeld good to go after rebase or any issues to resolve?

config/brakeman.ignore
```diff
@@ -0,0 +1,25 @@
...
+      "render_path": null,
+      "location": {
+        "type": "method",
+        "class": "AuthenticationsController",
+        "method": "create"
+      },
+      "user_input": "Authenticator.new(request.env[\"omniauth.auth\"], session).perform",
+      "confidence": "High",
+      "note": "Fixed here: https://github.com/18F/micropurchase/commit/799d74393c394e016fba5eab9df1eb638a7d2b06"
+    }
```
afeld: Is this necessary, with the change to the controller?

adelevie: Yes it is. The change doesn't silence the warning. The note in this file
points to the commit which protects against the (already low low risk)
warning.

afeld:
:shipit:

adelevie:
rebased

PR #109, opened 2015-11-23
Update index.html.erb

allalala:
Added administrative and policy language.
@vzvenyach please let us know if this looks OK.

### DIFF
--- a/app/views/logins/index.html.erb
+++ b/app/views/logins/index.html.erb
@@ -1,7 +1,16 @@
 <h2>Getting Started</h2>
 
-<p>Before participating in this marketplace, you’ll first need to register your company with SAM.gov and have a GitHub account. If you encounter any issues, let us know at <%= mail_to 'micropurchase@gsa.gov' %>.</p>
+<p>Before participating in this marketplace, you’ll first need to register your company with SAM.gov to receive a DUNS number, and have a GitHub account. </p>
 
-<p>Once you’re ready, continue with GitHub authorization (only public account information will be requested).</p>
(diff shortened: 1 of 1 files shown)


app/views/logins/index.html.erb
```diff
@@ -1,7 +1,16 @@
...
+
+<p>Once you’re ready, continue with GitHub authorization (only public account information will be requested). If you encounter any issues, let us know at <%= mail_to 'micropurchase@gsa.gov' %>.</p>
 
-<p>Once you’re ready, continue with GitHub authorization (only public account information will be requested).</p>
 
 <a class='usa-button' href="/auth/github">Authorize with GitHub</a>
+<p></p>
+<p></p>
+<p>Administrative Notes</p>
+<p>None of the content in this GitHub repository is considered proprietary or procurement-sensitive information.
```
vzvenyach: @allalala I think we need to specify which GitHub repository (https://github.com/micropurchase).

Additionally, I wonder whether we should focus on having two warnings: (1) at signup; and (2) at the time of bidding.

At signup, it seems that we'd say that we will only use this information to authenticate users and to communicate on an as-needed basis.

At time of bidding, I'd think we'd want to provide a warning describing how the bid information (and the bidder information) would be used for that bid/opportunity.

Thoughts>

allalala: Sounds good. Will make the changes to the copy on the bidding page.

vzvenyach:
@adelevie any idea why there's this security warning here?

adelevie:
I _think_ the fork @allalala PR'd from is older/out-of-sync.

allalala:
yes, @adelevie I think this is a problem with my fork. Can you show me how to fix this on Monday?

adelevie:
Yep!

adelevie:
closing in favor of https://github.com/18F/micropurchase/pull/117 (merged)

PR #110, opened 2015-11-23
Add email field to users

adelevie:


app/helpers/application_helper.rb
```diff
@@ -2,4 +2,17 @@ module ApplicationHelper
   def controller_class
     controller.class.name.include?('Admin') ? 'admin' : 'bidder'
   end
+
+  def flash_errors
+    result = nil
+    if !flash.empty? && !flash[:error].nil?
```
afeld: Minor: I personally prefer `unless` to `if !`.

afeld: Also, you only need the second half of this condition.

afeld:
Just the one minor style note – otherwise :shipit:

PR #112, opened 2015-11-24
Reject invalid bids

adelevie:


adelevie:
Ready for review. cc @allalala @harrisj

spec/models/place_bid_spec.rb
```diff
@@ -75,7 +80,91 @@
...
+    let(:auction) {
+      Presenter::Auction.new(Auction.create({
+        start_datetime: Time.now - 3.days,
+        end_datetime: Time.now + 7.days
+      }))
+    }
+    let(:amount) { 405 }
+
+    it 'should raise an authorization error' do
+      allow_any_instance_of(Presenter::Auction).to receive_message_chain(:current_bid, :amount).and_return(400)
```
baccigalupi: Needing a message stub chain seems like a smell. Wondering if current bid could just return a mock object. But there might be something simpler than this too. Sorry can't be of more help from github. If you want me to download and play around happy to. I am guessing that this is where folks on the Ruby channel were having concerns?

harrisj:
I would like to echo @baccigalupi's concern that using an `allow_any_instance_of` and a `message_chain` are things you should probably only use when dealing with legacy code. I'd suggest instead that we add something like [factory_girl](https://github.com/thoughtbot/factory_girl) and setup the DB with appropriate objects before each test. It'll be a bit tedious to do at first, but we'll be able to use these factories for integration testing, etc. as well.

That said, I am concerned that this PR is already 21 days old and what I'm proposing could add more time to it. I am fine if we were to merge this now, with the idea that these tests should be replaced in a subsequent pull request. Thoughts?

adelevie:
Is there a quicker fix that eliminates the message chaining without the factory overhaul? If so, let's do that, merge, and in a new PR add factories. Otherwise, maybe just get this merged and in a new PR, do the overhaul.

harrisj:
If I understand correctly, the message chaining is here precisely because there are no test records in the database that would be found on the normal call to `Presenter::Auction#current_bid` method (and a call to `amount` on that record). So, we could also just do straight DB creates, but if we are doing that it probably makes sense to make factories.

harrisj:
Anyhow, if you all are fine with this, I am good with the pull request being merged, contingent on there being a future pull request to improve the tests

PR #113, opened 2015-11-27
Updating the UI

vzvenyach:


vzvenyach:
Current view of the front page:
 
screen shot 2015-11-28 at 9 51 08 am

vzvenyach:
We have a working bid detail page, including a flash message when you have a successful bid. Screenshot below.

screen shot 2015-11-28 at 1 45 42 pm

vzvenyach:
Ok. I think this is ready for review.

allalala:
@ddotrollins can you also take a quick look at these QX adjustments and see if you spot any glaring issues?

ddotrollins:
@allalala looks good, I would remove the decimal places on the bid amount to read "$2,000" vs. "2,000.00" to reinforce only bidding in dollar increments. Otherwise is fine from a functional standpoint, we can make CSS updates later.

vzvenyach:
@adelevie is that something enforced at the model level? If not, I can probably format it at the template level.

adelevie:
The db is capable of decimals/floats, but the app prevents it. The relevant
files would be app/models/presenter/bid.rb and app/models/place_bid.rb.

Procfile
```diff
@@ -1 +1 @@
-web: bundle exec rails server -p $PORT
+web: bundle exec rails s
```
adelevie: @vzvenyach why this change? AFAICT, Cloud Foundry needs that $PORT env var.

adelevie:
rebased

PR #116, opened 2015-12-04
Updating purpose link to point at micropurchase repo

wslack:
@NoahKunin, not sure how much this matters but flagging.


allalala:
oh yeah, that was a typo 
cc @adelevie

PR #119, opened 2015-12-11
Fix "winning bid" prompt that is also shown to the losers.

adelevie:
Currently, the "you are the winning bidder" text is
displayed to *all* users when there is at least one bid.

Previously our tests were not capturinig this error because
they only tested presence or non-presence of the "winning
bidder" text in two scenarios: when the user placed bids
and when there are no bids. The latter scenario
insufficiently covered the cases when the "winning bid"
prompt should be present.

This commit presents another scenario, where there are bids,
but from other users. Currently this spec fails and will
ensure confidence in the subsequent bug fix.

### DIFF
--- a/app/helpers/bids_helper.rb
+++ b/app/helpers/bids_helper.rb
@@ -6,4 +6,10 @@ module BidsHelper
       number_to_currency(bid.amount)
     end
   end
+
+  def user_is_winning_bidder?(user, auction)
+    return false if !auction.current_bid?
+
+    user.id == auction.current_bid.bidder_id
+  end
 end
--- a/app/views/bids/new.html.erb
+++ b/app/views/bids/new.html.erb
@@ -22,9 +22,11 @@
     <div class='usa-width-one-whole'>
       <div class="usa-width-one-half">
         <h2 class='center'>Current bid:</h2>
-        <% if @auction.bids? %>
-          <p class="current-bid-amount center"><%= number_to_currency(@auction.current_bid_amount) %></p>
+        <p class="current-bid-amount center"><%= number_to_currency(@auction.current_bid_amount) %></p>
+        <% if user_is_winning_bidder?(current_user, @auction) %>
           <h3 class="center">You are currently the winning bidder.</h3>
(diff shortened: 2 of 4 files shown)


app/helpers/bids_helper.rb
```diff
@@ -6,4 +6,10 @@ def content_for_row(row_number, bid)
       number_to_currency(bid.amount)
     end
   end
+
+  def user_is_winning_bidder?(user, auction)
+    return false if !auction.current_bid?
+
+    user.id == auction.current_bid.bidder_id
+  end
```
baccigalupi: I would say you want to stick with presenters or switch to helpers. They perform the same task in two different ways. 

* Helpers are mixins and therefore end up as bald methods in the views. They can get hard to extract, when refactoring. Because they are methods in an unknown object, their dependencies can sometimes be hard to suss out. Also their are not easily available to non-view places like models and APIs (largely due to the unknown dependencies).
* Decorators are tightly scoped to a model, AR or Poro. This is what I had going on. They are limited that way and start to feel awkward as the view expands beyond the strict CRUD land.
* When that happens another option, and this is more a naming and conceptual than anything else, is a view model. View models are responsible for presenting the needed data to your view, and are tied to the concept of the view needs rather than the model needs.

Anyway, I would go with POROs for view logic or helpers but not both.

adelevie: This is helpful. So something like:

```ruby
class Presenter::Auction
# ...
  user_is_winning_bidder?(user)
    return false if !current_bid?
    user.id == current_bid.bidder_id
  end
end
```

And just call that directly in the view? (`<%=  auction.user_is_winning_bidder?(current_user) %>`)

(I'm really enjoying the aggressive tech debt prevention approach. It keeps growing on me)

harrisj: Have you decided if you want to make this change to a Presenter instead?

adelevie: That's what I was planning on doing (just haven't had the chance).

PR #120, opened 2015-12-14
Specs cover the displayed status on /bids/:id

adelevie:
Feature helpers returning both the auction and bid
objects. This makes it easier to access the objects
from within the specs.


harrisj:
This looks good to me.

PR #131, opened 2015-12-17
Fix typo in application template

adelevie:
https://github.com/micropurchase -> https://github.com/18f/micropurchase


harrisj:
Looks good to me. Merging.

PR #133, opened 2015-12-17
Update schema.rb

adelevie:
@vzvenyach this should solve the issue you had earlier where `db/schema.rb` was updated and you weren't sure if you should check it in or not.


harrisj:
Looks good to me

PR #147, opened 2015-12-22
Eliminate mocking within the controller tests

harrisj:
Continuing on the work from the prior `factory_girl` PR #142, this pull request reworks the controller tests to have a few changes:

* Create test objects using Factory Girl instead
* Instead of mocking `current_user` within the controller, pass up the session object as part of the test requests.


adelevie:
An enthusiastic :+1:.

PR #154, opened 2015-12-22
Added Edit Profile link to header.

adelevie:
screen shot 2015-12-22 at 5 19 40 pm


harrisj:
I changed this to use Rails' built in `_path` helpers for constructing URLs from the routes instead of manually building them. Otherwise it looks good to me!

PR #155, opened 2015-12-23
Use Rails' method for building URLs in views

harrisj:
Following the work in #154 to use the URL construction methods in Rails, I decided to check some other views and fix those as well. These can be merged after #154 is.


adelevie:
rebased

adelevie:
rebased again with new changes

PR #165, opened 2015-12-23
Add markdown support

harrisj:


harrisj:
This probably could use a helper instead of those `<%= raw %>` declarations, but I need someone else to weigh in

app/models/presenter/auction.rb
```diff
@@ -70,6 +75,16 @@ def current_bid_record
       @current_bid_record ||= bids.sort_by{|bid| [bid.amount, bid.created_at, bid.id]}.first
     end
 
+    def markdown
+      # FIXME: Do we want the lax_spacing?
```
adelevie: what is `lax_spacing`?

adelevie:
screen shot 2015-12-23 at 4 42 52 pm

Seems to be missing any `h` tags.

adelevie:
screen shot 2015-12-23 at 5 47 19 pm

adelevie:
Did some rebasing.

harrisj:
We should add shared specs to eliminate the duplication in the markdown tests, but I am not sure how to do that exactly yet and I want to get this in there.

PR #167, opened 2015-12-24
Use factory_girl in the db:seed task

harrisj:
Just for completeness


adelevie:
Neat! Does this mean actual records will not be created outside of the test environment?

harrisj:
No, Factory_Girl is loaded in the development environment so it'll work there too

PR #169, opened 2015-12-28
Cleaning up factory_girl factories

harrisj:
I realized it would be better to translate some of the Auction factories
into traits so tests can more accurately use them if they need bidders
or not, so this is that (plus cleaning up a few things)


adelevie:
:+1:

PR #171, opened 2015-12-28
Wrap auction description in auctions#show view.

adelevie:
Showcasing my l33t CSS skills, I added a `white-space: pre-line;` property to prevent the auction description text from running into the auction description box in the `auctions#show` view.

In sum, this code makes 
screen shot 2015-12-28 at 10 13 22 am

Look like

screen shot 2015-12-28 at 11 22 28 am

This still leaves some extra styling to be desired: the line breaks below the `h` tags is probably too much and should be reduced. But that can happen in a subsequent PR.

Iterative improvements ftw!

### DIFF
--- a/app/assets/stylesheets/main.scss
+++ b/app/assets/stylesheets/main.scss
@@ -28,6 +28,10 @@ textarea#auction_description {
   }
 }
 
+.github-issue {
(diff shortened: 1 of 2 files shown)


harrisj:
Looks good to me

PR #175, opened 2015-12-29
Clarify wording about DUNS number and SAM.gov

kfogel:
On the login page, change wording to reflect the possibility that
the vendor may already have a DUNS number.

### DIFF
--- a/app/views/logins/index.html.erb
+++ b/app/views/logins/index.html.erb
@@ -1,6 +1,6 @@
 <h2>Getting started</h2>
 
-<p>Before participating in this marketplace, you’ll first need to register your company with <a href="https://www.sam.gov/portal/SAM/##11">SAM.gov</a> to receive a DUNS number, and have a GitHub account. </p>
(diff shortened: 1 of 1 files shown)


kfogel:
We already have a DUNS number.  When I saw the login page ("Before participating in this marketplace, you’ll first need to register your company with SAM.gov to receive a DUNS number, and have a GitHub account") I wasn't sure whether that meant we still needed to create a SAM.gov account or not.  I went through the appx 2-hour SAM.gov registration process just in case -- oh joy -- but it seems possible that that time was wasted, if all that 18F micropurchases really needed was our DNS number.

So I'm not totally sure that the new wording in this pull request is correct.  If people truly just need the DUNS number (and SAM.gov is merely one way to obtain it), then perhaps it would be better just to point people to https://fedgov.dnb.com/webform, which is probably going to take them less time than the SAM.gov process.  Conversely, if a SAM.gov registration (whether individual or entity) is truly required, then the site should say so, and make it clear that merely having a DUNS number is not enough -- that one still needs to register on SAM.gov.

adelevie:
Hi @kfogel,

Thanks for the pull request and detailed explanation. It's helped us clarify our message which we hope will remove the initial confusion our wording presented to you.

We actually need the SAM.gov registration (as opposed to just the DUNS, sans SAM). We have a PR [here](https://github.com/18F/micropurchase/pull/191) with the clarifications. We'd love your feedback!

kfogel:
That other PR clarifies everything, thanks!

PR #177, opened 2015-12-30
Display number of users in admin

adelevie:
screen shot 2015-12-30 at 12 42 42 pm

Added some blank users:

screen shot 2015-12-30 at 12 44 16 pm


harrisj:
This page will get more annoying as the number of users grows above a certain size because there is no pagination or even ordering of the user records (except by ID). But we will cross that bridge when we come to it. I'm good to merge.

PR #181, opened 2015-12-30
Added destroy button next to each auction in /admin/auctions.

adelevie:
Failing spec for admin delete auctions


harrisj:
Looks good to me

PR #183, opened 2015-12-31
Add Rubocop

harrisj:
Here is what Rubocop wants to fix. I tweaked the [default configuration](https://github.com/bbatsov/rubocop/blob/master/config/default.yml) to disable some things that were triggered all over the place (like using single ' instead of " if there is no interpolation), modify a few other things like the maximum line length, and there are a few places where I just needed to disable specific rubocop checks, but this should give you a rough idea of what is involved. Look at the attached config to this PR as well to see what it is disabling and tweaking.

To run this manually, run `bundle exec rubocop` at the root and it'll print out the issues it finds. But, to make this more effective, I recommend installing a Rubocop-compatible linter in your favorite editor:

* [Sublime Edit](https://sunlightfoundation.com/blog/2015/03/31/making-sublimelinter-work-with-rbenv-and-rubocop/)
* [Atom](https://atom.io/packages/linter-rubocop)
* [vim](https://github.com/scrooloose/syntastic)
* Emacs (who am I kidding, this is just me)

If we are cool with this, we can then enable [Hound CI](https://houndci.com/) on the project to nitpick on future commits. [This can be annoying](https://github.com/18F/myusa/pull/713) if people push up WIPs or PRs without checking the style, but it does annotate the specific problems in a PR. Alternatively, we could add a scenario to make the build fail if Rubocop flags any errors much like we do for Brakeman. Or we could just leave it unenforced and clean it up regularly.

### DIFF
--- a/.rubocop.yml
+++ b/.rubocop.yml
@@ -1,34 +1,114 @@
 AllCops:
+  Include:
+    - 'Gemfile'
+    - 'Rakefile'
+    - 'lib/tasks/*'
+    - "**/*.rake"
+    - "**/Capfile"
+    - "**/Guardfile"
+    - "**/Vagrantfile"
+    - "**/Berksfile"
+    - "**/Cheffile"
   Exclude:
-    - 'bin/**/*'
-    - 'db/**/*'
-  RunRailsCops: true
+    - 'db/schema.rb'
+
+Lint/ParenthesesAsGroupedExpression:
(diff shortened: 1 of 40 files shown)


adelevie:
This is excellent!

harrisj:
That's what you say now, but wait until Hound raps you with 23 emails for various style issues on your next pull request

adelevie:
I'll know who to ('git') blame.

PR #190, opened 2016-01-05
Validate DUNS before allowing a bid

harrisj:
I added a field and check that the user has been verified to have a SAM account before accepting a bid, but the tests are currently not great and I'm not sure the flow is correct. This is still a work in progress and needs some serious thought. DO NOT MERGE YET.


adelevie:
accidentally rebased against master. fixing ...

app/views/auctions/_list_item.html.erb
```diff
@@ -16,13 +16,12 @@
     <%= render partial: 'auctions/bids_and_form', locals: {auction: auction} %>
 
       <div class="usa-width-one-third">
-        <% if auction.available? %>
+        <% if auction.available? && (!current_user || current_user.sam_account?) %>
           <a href="<%= new_auction_bid_path(auction) %>" class='usa-button'>Bid »</a>
         <% else %>
-          <a href="#" class='usa-button usa-button-disabled'>Closed</a>
+          <a href="#" class='usa-button usa-button-disabled'>Bid</a>
```
harrisj: Should we just not show a button at all? Or is it better to show a deactivated button?

app/views/components/_no_sam_header.html.erb
```diff
@@ -0,0 +1,8 @@
+<div class="usa-alert usa-alert-error auction-alert">
```
harrisj: Text help needed

harrisj:
Rebased against develop and force pushed

Gemfile
```diff
@@ -20,6 +20,7 @@ gem 'chronic'
 gem 'email_validator'
 gem 'redcarpet'
 gem 'puma'
+gem 'samwise', github: '18f/samwise'
```
adelevie: add a branch here until I can get a new gem released

app/views/auctions/_list_item.html.erb
adelevie: I think this is fine for now.

adelevie:
TODOS:

- get failing spec for rake task working
- ensure stuff that needs to be mocked is mocked (I haven't addressed this yet)

adelevie:
This all looks great. I played with this locally and does its thing pretty well.

adelevie:
We should include in this a script that will automatically run the migration when we deploy to staging.

adelevie:
Also, is the plan then to manually run the rake task at different points in the day?

Just saying, but if we have an authed write API, we could run the same task without cf-sshing :)

harrisj:
We could probably try on-demand SAM checks too if you wanted, but yeah otherwise we have to ssh there several times a day

app/views/components/_no_sam_header.html.erb
allalala: bullet 1:  You have registered on SAM.gov 
bullet 2:  Your SAM.gov account is "active" in the system. Check your status using the SAM.gov status checker https://www.sam.gov/sam/helpPage/SAM_Reg_Status_Help_Page.html
bullet 3:  You have entered your DUNS number into your profile on this platform.

adelevie:
rebased it all into one beautiful commit.

adelevie:
let's figure out migrations before merging.

app/models/user.rb
```diff
@@ -1,4 +1,29 @@
 class User < ActiveRecord::Base
   has_many :bids
   validates :email, email: true, allow_blank: true
+  before_save :clear_sam_status_if_duns_changed
```
baccigalupi: I don't think this belongs inside User. Also it is specific to vendor users, not admins. All this is a wrapper around user data that adds to it.

spec/features/bidder_interacts_with_auction_spec.rb
```diff
@@ -102,10 +102,47 @@
...
+    # returns us back to the bid page
+    expect(page).to have_content("Current bid:")
+    expect(page).to have_content("$800.00")
+  end
+
   context "placing bids" do
     scenario "Viewing auction list and detail view as a logged out user" do
       create_current_auction
-
+      @bidder = create_authed_bidder
```
baccigalupi: let?

app/models/user.rb
adelevie: Is the fix extracting the logic of `clear_sam_status_if_duns_changed` into a different module/class but retaining the `before_save` callback ... or is there a way to encapsulate without the AR callback?

baccigalupi: There is probably only one place in one controller where this duns check needs to be done. This logic can be wrapped up there, and the User can just remain data. Tests don't get funky and have side effects. And lots of other unexpected side effect averted hopefully too

adelevie: @baccigalupi I agree, these are the changes we want. I'm have a WIP PR open here now https://github.com/18F/micropurchase/pull/196 containing my first sketches at implementing these. I don't think we're going to let that block the merging of this PR though, since we need to get some things ready for a Monday auction launch.

adelevie: Also, Thanks!!!!

spec/features/bidder_interacts_with_auction_spec.rb
harrisj: Yes, cleaning up these feature_helpers to use let and FactoryGirl more directly is in the future I think

PR #193, opened 2016-01-06
Added veiled methods for use in non-admin views

harrisj:
Renamed the existing `bidder_name` and `bidder_duns_number` to `veiled_` equivalents and provided an access for the underlying value so the admin views can just call the direct display equivalents.

After a discussion with @baccigalupi, we decided this would be a better approach than relying on polymorphism and creating an `AdminAuction` and `AdminBid` presenters. Should there be more specialized distinctions between both views down the road, we might reconsider this approach.

### DIFF
--- a/app/models/presenter/bid.rb
+++ b/app/models/presenter/bid.rb
@@ -4,20 +4,28 @@ module Presenter
       Presenter::DcTime.convert_and_format(created_at)
     end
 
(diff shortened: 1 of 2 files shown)


adelevie:
This is so elegant, thanks so much @harrisj and @baccigalupi!

PR #194, opened 2016-01-08
WIP: API prototype

harrisj:
Creating this so we can look into merging @adelevie's work soon


app/serializers/admin/auction_serializer.rb
```diff
@@ -0,0 +1,11 @@
+module Admin
+  class AuctionSerializer < ActiveModel::Serializer
+    has_many :bids, seralizer: Admin::BidSerializer
+
+    attributes :issue_url,      :start_price,
```
jessieay: I find this format of listing args to be difficult to read and update. prefer:

```ruby
attributes(
  :issue_url,
 :start_price,
 :start_datetime,
)
```

that way it is also easy to alpha order to see if anything is missing!

config/initializers/assets.rb
```diff
@@ -11,3 +11,4 @@
 # Rails.application.config.assets.precompile += %w( search.js )
 
 Rails.application.config.assets.paths << Rails.root.join('app', 'assets', 'images')
+Rails.application.config.assets.precompile += %w( us_flag_small.png )
```
jessieay: this shouldn't be required if the image is in `app/assets/images`

Gemfile.lock
```diff
@@ -44,14 +46,14 @@ GEM
       minitest (~> 5.1)
       thread_safe (~> 0.3, >= 0.3.4)
       tzinfo (~> 1.1)
-    addressable (2.4.0)
+    addressable (2.3.8)
```
harrisj: Why did we go down a version for this? Curious

harrisj: Oh, I see it's a constraint from Sawyer. Never mind me

harrisj:
I want to add more tests, but given there is a deadline for one endpoint and it's only an admin-used one, I'm happy to let it slide for the moment, but we can make tests tomorrow.

adelevie:
:+1: @harrisj cc @vzvenyach

PR #195, opened 2016-01-08
Copy migrate frequently setup from C2

adelevie:
Implements "migrate frequently" as described here: https://docs.cloudfoundry.org/devguide/services/migrate-db.html and copied from 18F/C2.

### DIFF
--- a/manifest-staging.yml
+++ b/manifest-staging.yml
@@ -8,3 +8,4 @@ applications:
   buildpack: https://github.com/ddollar/heroku-buildpack-multi.git
   services:
     - micropurchase-staging-psql
+  command: bundle exec rake cf:on_first_instance db:migrate && bundle exec puma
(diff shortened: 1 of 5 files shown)


manifest-staging.yml
```diff
@@ -8,3 +8,4 @@ applications:
   buildpack: https://github.com/ddollar/heroku-buildpack-multi.git
   services:
     - micropurchase-staging-psql
+  command: bundle exec rake cf:on_first_instance db:migrate && bundle exec puma
```
afeld: Are you running Puma? If not, you should change the second command here.

afeld:
Just the one comment – otherwise, looks good! I would try to deploy to a staging environment before merging though.

manifest-staging.yml
adelevie: we are totally running with puma!

adelevie:
@afeld Thanks! I'm writing tests for the rake task and for `ServerEnv`.

adelevie:
Tests will be in a subsequent PR.

PR #221, opened 2016-01-11
Fix comma error

adelevie:
Fixes a bug where bids for "1,000" turn into "1".

### DIFF
--- a/app/models/place_bid.rb
+++ b/app/models/place_bid.rb
@@ -56,6 +56,11 @@ class PlaceBid < Struct.new(:params, :current_user)
   # rubocop:enable Style/IfUnlessModifier, Style/GuardClause
 
   def amount
-    (params[:bid] && params[:bid][:amount]).to_f.round(2)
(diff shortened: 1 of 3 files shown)


app/models/place_bid.rb
```diff
@@ -56,6 +56,11 @@ def validate_bid_data
   # rubocop:enable Style/IfUnlessModifier, Style/GuardClause
 
   def amount
-    (params[:bid] && params[:bid][:amount]).to_f.round(2)
+    params_amount = params[:bid][:amount]
+
+    # ensures that "1,000" does not save as "1"
+    params_amount = params_amount.gsub(',', '') if params_amount.is_a?(String)
+
+    (params[:bid] && params_amount).to_f.round(2)
```
harrisj: Ugh, we should probably look to get rid of this since we don't have floats as representations anymore

adelevie: Let's get rid of this later. Removing floats will break a lot of the specs and require a deeper refactor.

PR #223, opened 2016-01-11
Test for and fix comma error

adelevie:


spec/models/place_bid_spec.rb
```diff
@@ -117,6 +117,16 @@
...
+  context 'when the amount has a comma' do
+    let(:bid)    { place_bid.bid }
+    let(:amount) { '1,000' }
+
+    it 'should disregard the comma' do
+      place_bid.perform
+      expect(bid.amount).to eq(1000)
+    end
+  end
+
```
konklone: It'd be nice to add tests that run a ton of different strange inputs through the bidding process. For example, `1.000`, `1.000.00`, `0`, `-1,000`, `-1000`, `$1000`, `$1,000`, `$-1,000`, `1,.23,...,..45...$$$`, `@#$%78367834#$@$`, `five thousand`...

And I would suggest that for the most part, only `$` and `,` characters be "allowed" by stripping them out, allowing opening or closing whitespace. So any input that, after a `#strip` call, doesn't match `/^[\d,\$]+$/` be rejected, and the tests describe that behavior.

PR #228, opened 2016-01-11
Removes white-space: pre-line css from auction details (#216)

### LINKED ISSUE #216: Fix spacing in auction details page
Currently, the vertical spacing between elements in the rendered markdown of auction descriptions is too large.

screen shot 2016-01-11 at 11 36 04 am

We need to reduce the spacing, while still ensuring that no text runs into the bid info box to the right.

bmathews:
This is causing line breaks wherever the source html includes line breaks. It was introduced by https://github.com/18F/micropurchase/commit/81f3213ab01674a57169ae980721649810daeae0, but doesn't look like it's needed. Fixes #216 

Before:

image

After:

image

### DIFF
--- a/app/assets/stylesheets/main.scss
+++ b/app/assets/stylesheets/main.scss
@@ -65,10 +65,6 @@ textarea#auction_description {
   }
 }
 
-.github-issue {
-  white-space: pre-line;
-}
-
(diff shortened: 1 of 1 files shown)


maya:
@bmathews btw LGTM

PR #233, opened 2016-01-12
Fix Link on `logins#index`

reshleman:
The link to the SAM Status Tracker appears to have been incorrectly
converted from markdown. This commit fixes the link by using HTML.


adelevie:
Thank you!

Issue #219, opened 2016-01-11
As a bidder, I would like to see which is my bid


JoshData:
After bidding, one forgets what one's bid was. On `/auctions/{:id}/bids`, the name shouldn't be withheld if it's one's own bid.

adelevie:
This is great feedback! Thank you @JoshData.

harrisj:
Fixed in #234

PR #234, opened 2016-01-12
Show the current user in open auction bids

harrisj:
When a user is logged in, they should be able to see themselves in the current bid list even when all other users are redacted. Fixes issue #219


JoshData:
+1!

PR #243, opened 2016-01-21
Expand and test the API

adelevie:
This PR is pretty big and is the culmination of some smaller PRs.

- creates a robust requests spec suite for the API
- allows controllers to be completely agnostic of request type (API or web UI is all the same)
- ensures that bidder info is properly veiled/redacted for running auctions
- adds the following routes:

GET `/admin/auctions`
GET `/admin/users`
GET `/auctions`
GET `/auctions/:id`
POST `/auctions/:id/bids` (yes, you can now bid via API).

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -27,6 +27,8 @@ gem 'active_model_serializers'
 
 group :test do
   gem "codeclimate-test-reporter", require: nil
+  gem 'webmock'
+  gem 'json-schema'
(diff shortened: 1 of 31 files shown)


adelevie:
I'm going to merge this to staging just so I can test out the bidding stuff against a live site. When I'm done, I will revert this PR pending review.

PR #244, opened 2016-01-21
Add GitHub username to users

harrisj:
I have added a field for the github_login to the `User` model and a decorator for fetching it from Github (TODO: look for an API key environment variable). I don't want this to be merged yet, because I've tried to use a Decorator approach and I'd like @baccigalupi to weigh in on what should be fixed. Once we have this settled, then we can go to try fixing the `sam_account?` field to follow a similar approach


adelevie:
LGTM so far. My open PR handles some octokit stubbing using webmock (which lets us avoid allow_any_instance_of), so I might PR/commit some updates here to harmonize.

harrisj:
I am going to make a few tweaks to bring this in line with #245

adelevie:
Great! I'll get a head start on reviewing this (even though it seems to be
a moving target).

harrisj:
I have refactored this to follow the pattern in #245 and defined a `GithubReckoner` class. Note that adding the github login requires a Rake task to be run currently. Caveats:

1. I still need to figure out how to get the Octokit::Client some credentials from environment variables or such.
2. It would be useful to have it look for the github_login when the user logs in (in `authenticator.rb`), Possibly we could also pull in the Github avatar in the future. But I haven't coded in that hook yet because it would require a bit of mocking and stubbing in some places.

I'd like @baccigalupi to look this over as well as @adelevie to make sure I'm doing it right. Thank you.

harrisj:
I have added code to pull in environment variables to init the Octokit, but I have not yet added anything to automatically reckon against github when a user logs in. Should we do that? Also, I have yet to add code to display the github login if the name is blank

harrisj:
I just rebased this off the latest on `develop`

PR #245, opened 2016-01-21
Refactor: Skinnier User Junks

baccigalupi:
* Extracts sam account related checks and user modifications to
          own class: SamAccountReckoner (?rename?)
        * Removes scope from User and moves logic into sam related rake
          task.
        * Removes save related hook from User and moves it to the
          controller update action for users. This is the only place it
          is needed.
        * Extracts update related logic from UserController and into
          UpdateUser class. Moves tests to less expensive place in
          models, stubs updater in controller tests. Also, removes
          unused redirect related private method.

@harrisj You were asking about this area and what to do in slack. Had a couple extra hours yesterday ... Tell me what you think. Also thinking still about your auction types, and it seems like there isn't enough product detail to build the right abstraction, but I am pretty in the dark about it.

@adelevie your review too please

In the air today, so do whatever you want with it!

### DIFF
--- a/app/models/sam_account_reckoner.rb
+++ b/app/models/sam_account_reckoner.rb
@@ -1,0 +1,20 @@
+class SamAccountReckoner < Struct.new(:user)
+  def clear
+    user.sam_account = false if should_clear?
+  end
+
+  def set
+    return true if user.sam_account?
+    user.sam_account = client.duns_is_in_sam?(duns: user.duns_number)
+  end
+
+  private
+
+  def should_clear?
+    user.persisted? && user.duns_number_changed?
+  end
+
+  def client
+    @client ||= Samwise::Client.new
+  end
+end
--- a/app/models/update_user.rb
+++ b/app/models/update_user.rb
@@ -1,0 +1,33 @@
+class UpdateUser < Struct.new(:params, :current_user)
+  attr_reader :status
+
+  def save
+    fail UnauthorizedError if !allowed_to_edit?
+
+    update_user
+  end
+
+  def errors
+    user.errors.full_messages.to_sentence
+  end
+
+  def user
+    @user ||= User.find(params[:id])
+  end
+
+  private
+
+  def update_user
+    user.assign_attributes(user_params)
+    SamAccountReckoner.new(user).clear
+    @status = user.save
+  end
+
+  def allowed_to_edit?
+    current_user == user
+  end
+
+  def user_params
+    params.require(:user).permit(:duns_number, :email)
+  end
+end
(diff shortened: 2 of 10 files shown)


lib/tasks/sam.rake
```diff
@@ -1,9 +1,9 @@
 namespace :sam do
   task check: :environment do
-    User.not_in_sam.each do |user|
+    User.where(sam_account: false).each do |user|
```
harrisj: Does it make sense to have a class method like `SamAccountReckoner.unreconciled` that wraps this `where` clause. That way the Rake task doesn't have to know anything about the schema structure of the User table

app/models/sam_account_reckoner.rb
```diff
@@ -0,0 +1,20 @@
+class SamAccountReckoner < Struct.new(:user)
```
harrisj: Do you think we should namespace these within some module like we do for the Presenters? Maybe something like `Decorator::SamAccountReckoner` or is that not really that big a deal?

app/models/update_user.rb
```diff
@@ -0,0 +1,33 @@
...
+
+  def user
+    @user ||= User.find(params[:id])
+  end
+
+  private
+
+  def update_user
+    user.assign_attributes(user_params)
+    SamAccountReckoner.new(user).clear
```
harrisj: Ah I see. This is also where we could put a check against the SAM API if we wanted to do that

harrisj:
This looks good to me. I do have a few minor questions (and will probably follow a similar style in the work that is in #244), but I am good to merge

app/models/sam_account_reckoner.rb
adelevie: +1 to namespace.

lib/tasks/sam.rake
baccigalupi: That seems like a good idea. 

I took out the scopes because those can lead to a lot of pain. I have actually seen spreadsheets where the numbers don't add up because if you put scopes in different orders, different results. Also, if you think of scopes as an abstraction, you don't want to extract it until you have three usages.

Adding a patch on the SamAccountReckoner to hide what is actually happening seems harmless and easy to reason about!

app/models/sam_account_reckoner.rb
baccigalupi: Namespace might be good, but this is not so much a decorator. Another alternative that I have used and Sandi Metz has talked about using in putting your AR records in a namespace. That sets up the expectation that those will be the minority of work, and that you name your other models according to what they do and are about. I think it is possible that the decorators might break up into view model, specific to a particular view representation, and areas of responsibility around a data model like this one.

baccigalupi:
Hey folks, what is the current merge process. Do you do it? Do I do it?

harrisj:
I think one of us needs to do it (since you submitted the PR), but it looks good to me. But I want to let @adelevie do the merge if he is happy

adelevie:
big :+1: on this!

PR #270, opened 2016-02-01
add meta tags

stvnrlly:
This should add meta tags for all pages. I added content for all of them, but can definitely revise as appropriate.

The use of the `twitter:label` and `twitter:data` tags is probably the least common thing. I don't know if they're used by Twitter any longer, but they are used in things like Slack. To see what it looks like, check out a Medium post on Slack.

I don't think that there's anything else required to get sites to actually use these tags, but unfortunately that can only be checked after they're up and running. Facebook might be the longest lag time, but a scan can be run through their dev console, I believe.

### DIFF
--- a/app/views/auctions/index.html.erb
+++ b/app/views/auctions/index.html.erb
@@ -1,4 +1,12 @@
+<% content_for :data_tags do %>
+  <meta name="twitter:label1" value="Active Auctions">
+  <meta name="twitter:data1" value="<%= @auctions.count { |i| i.start_datetime < Time.now && Time.now < i.end_datetime } %>">
+  <meta name="twitter:label2" value="Coming Auctions">
+  <meta name="twitter:data2" value="<%= @auctions.count { |i| Time.now < i.start_datetime } %>">
+<% end %>
+
 <div class="usa-width-full">
+
   <h1>Welcome!</h1>
   <div class="usa-font-lead">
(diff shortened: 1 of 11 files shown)


harrisj:
Yes, it looks like Twitter has [deprecated the product card that uses those labels](https://dev.twitter.com/cards/types/product), and [we should probably switch to a basic summary](https://dev.twitter.com/cards/types/summary). Although I do appreciate your efforts to put the current bid, etc. in there. What do others think?

Do we have an icon image for Micropurchase app that would could also specify for these cards?

allalala:
Re Icon/image: maybe the new 18F logo?

stvnrlly:
I did use the `summary` card: see [here](https://github.com/18F/micropurchase/pull/270/files#diff-9599427925097c3c66f26ac1e0de5cadR24). The reason that I also added the product card tags is that other non-Twitter services still look for them and display that content. For example, Medium's "Reading time" tag shows up on Slack even though the site uses a `summary_large_image` card:

product tags on slack

Re: the icon: we can definitely use the 18F logo, but I avoided it since Twitter recommends against using logos. It's probably fine to do, though. Another approach would be to take a representative screenshot from the site and use that for a `summary_large_image` card.

harrisj:
Fair enough, those are valid points. But I have one last question: should we add some checks for the meta tags in our feature tests? I can do this if you want.

stvnrlly:
Since these will determine what's shown to the outside world, testing sounds wise. I'll take a pass at it.

stvnrlly:
Okay, I added a few feature tests and think it should be ready to go.

harrisj:
Looks good to me!

PR #273, opened 2016-02-01
Don't hit the GitHub API if the api-key header is blank

adelevie:
@harrisj, this was the rather simple solution to our rate limiting problem.

If there is not api key, don't hit the github api in hopes of authenticating.

If there is an api key, we'll hit the api. We might still risk some sort of rate limiting if we have many requests with invalid keys. I'll try to test this locally.

### DIFF
--- a/app/controllers/application_controller.rb
+++ b/app/controllers/application_controller.rb
@@ -66,6 +66,8 @@ class ApplicationController < ActionController::Base
   end
 
   def github_id_from_api_key(api_key)
+    return nil if api_key.nil?
+
     client = Octokit::Client.new(access_token: api_key)
     client.user.id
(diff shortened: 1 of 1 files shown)


adelevie:
I ended up using this Python script to try to trigger the rate limit (locally):

```python 
import requests

for x in range(500):
    headers = {'Accept': 'text/x-json', 'Api-Key': 'gluglgulglguglg'}
    response = requests.get("http://localhost:3000/auctions", headers=headers)
    print(x)
    print(response.status_code)
```

This clearly made more than 60 requests in an hour--however not once did we trigger the GitHub API limit.

harrisj:
Nice, I approve of this pull request

PR #278, opened 2016-02-03
Confirmation page

harrisj:
This is still a WIP, since I need to update the tests I broke (I know, bad Jake!). But it's a start on the confirmation screen. It does something a little unconventional with the PlaceBid object to dry run validation for the confirmation screen, but I think this is better than other approaches. Also, I fixed the button issues with the Submit / Cancel buttons

### DIFF
--- a/app/assets/stylesheets/main.scss
+++ b/app/assets/stylesheets/main.scss
@@ -216,8 +216,7 @@ textarea#auction_description {
   max-width: none;
 
   .cancel, .submit {
-    float: left;
-    margin: 0;
+      display: inline-block;
   }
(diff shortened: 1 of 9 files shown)


allalala:
@harrisj can you add a screenshot of what this looks like on your local machine?

harrisj:
micro-purchase

harrisj:
Marking this as a WIP no longer since its tests now pass.

adelevie:
nice! I shall review!

adelevie:
looks phenomenal to me :dragon_face:

PR #283, opened 2016-02-09
Cleanup/Refactor Pre work on sam warning toggle

baccigalupi:
@harrisj @adelevie 

I started work on this but just had to clean up the views I was in. I didn't really have to, but I couldn't help myself. So, I shoved a lot of logic into the presenters and made some new view models.

Anyway, I have failing (@wip) cucumber specs for the first failure. And I am about ready to start js-ing away. 

In the meanwhile, can you both check this out and see what you think?


app/models/presenter/auction.rb
```diff
@@ -19,24 +21,18 @@ def current_max_bid
       end
     end
 
-    def current_bid_amount
-      current_bid.amount
-    end
+    delegate :amount, :time,
```
harrisj: Thank you. I wasn't sure if we should go to this step yet, but it makes sense

app/models/presenter/auction.rb
```diff
@@ -94,8 +94,46 @@ def html_summary
...
+      status_name = if expiring?
+                      'Expiring'
+                    elsif over?
+                      'Over'
+                    elsif future?
+                      'Future'
+                    else
+                      'Open'
+                    end
+      "::Presenter::AuctionStatus::#{status_name}".constantize
```
harrisj: This is an interesting technique for polymorphism, but I noticed this immediately. So, if I may play Devil's Advocate here, what is the specific advantage of having distinct Ruby classes for these fields that we instantiate like this vs., say,  an equivalently structured hash for looking up such information? I guess the goal is that you can test all these classes with their own unit tests?

app/models/view_model/auctions_index.rb
```diff
@@ -0,0 +1,33 @@
...
+      auctions.count { |i| i.start_datetime < Time.now && Time.now < i.end_datetime }
+    end
+
+    def upcoming_auction_count
+      auctions.count { |i| Time.now < i.start_datetime }
+    end
+
+    def header_partial
+      if current_user && current_user.sam_account?
+        '/components/sam_verified_header'
```
harrisj: Should these be in `/components/` like I had placed some of them earlier or is there a better place for them?

app/views/auctions/index.html.erb
```diff
@@ -16,34 +15,8 @@
...
-<!--   <hr class='usa-size-one-whole'> -->
-  <% if @auctions.empty? %>
-    <p class="usa-font-lead">There are no current open auctions on the site. Please check back soon to view micropurchase opportunities.</p>
-  <% else %>
-    <% @auctions.each do |auction| %>
-      <%= render partial: 'list_item', locals: {auction: auction} %>
-<!--       <hr> -->
-    <% end %>
-  <% end %>
+  <%= render partial: @view_model.auctions_list_partial %>
```
harrisj: I know you don't have to explicitly pass in `locals` to render partials, but I was wondering if you think we should here?

app/views/auctions/_list_item.html.erb
```diff
@@ -2,7 +2,7 @@
   <div class="usa-grid">
     <div class='usa-width-one-whole issue-label'>
       <h3 class='issue-title'>
-        <span class="usa-label-big <%= auction_label_class(auction) %>"><%= auction_label(auction) %></span> <a href="<%= auction_path(auction) %>"><%= auction.title %></a>
+        <span class="usa-label-big <%= auction.label_class %>"><%= auction.label %></span> <a href="<%= auction_path(auction) %>"><%= auction.title %></a>
```
harrisj: Another dumb question, but would you then wrap the auction in an `AuctionShow` class here to use that code in these partials? I'm just trying to make sure I correctly understand how we don't duplicate methods between `AuctionIndex` and `AuctionShow` when the same information is displayed on both pages.

app/models/presenter/auction.rb
baccigalupi: So the pattern is called something like 'replace conditional with polymorphism'. http://refactoring.com/catalog/replaceConditionalWithPolymorphism.html

The goal is to replace the same conditional in a bunch of places into some fairly stupid declarative objects ... and only have the conditional in one place. 

Sandi Metz did a talk at Rails conf called (I think) "Nothing is something" that takes this a bit further and tries to find the missing thing that is covered by nil or false.

app/models/view_model/auctions_index.rb
baccigalupi: I think these are view models, which is to say they are tightly bound to a single view. In this case it is /auctions or the AuctionsController#index action.

Components seems like a good place for the view things that span concepts. Sometimes I also see these partials hanging out inside the layouts directory of views.

app/views/auctions/index.html.erb
baccigalupi: The underlying partial for auctions when there are auctions passes the locals in the same way.

I think keeping it simple and not passing things that are available is probably a better way, especially when there is just one instance variable exposed in the views.

By the way I am trying to stick to these rules: https://robots.thoughtbot.com/sandi-metz-rules-for-developers

app/views/auctions/_list_item.html.erb
baccigalupi: Yeah I was wondering how these things were going to divide up too.

I put the stuff in the view models that were specific to the individual action/view. In this case we are in a list item for an auction. It is really in the realm of a single object that is potentially used in both the show and the index views. I was feeling at first that these methods seemed so very view related I didn't want to throw them into the presenter, and maybe there is a delegate class that could be extracted from the presenter and put into the view model. I just didn't see a sensible way to do that in the end.

baccigalupi:
One other thing. Moving so much to presenters and view models means that there is a lot more to test in these objects. The tests should be fast to run, but will take time to backfill. It isn't this group of commits.

harrisj:
Thank you for the education. I can help fill in tests later.

app/views/auctions/_list_item.html.erb
harrisj: I definitely get the dilemma: what would be the point of moving all that logic out of the model if you just had a really long Presenter? And so ViewModels (which are new to me) are indeed a useful approach for targeting methods to specific Controller views and avoiding the ugliness of `if-elsif-elsif-end` inside ERB files. But there are definitely some cases where we have to share the same presentation of attributes across multiple views and their corresponding API endpoints, and I guess those would have to be in a Presenter for now?

baccigalupi: if statements are consistent places of errors, but also complexity. Small methods are easier to work with not just because there is less to absorb, but because we give the smaller concept a name (and hopefully a good name). Knowing that partial based on some named criterial will always be rendered is more helpful that trying to navigate an if-else branch in a view. There shouldn't be logic in views. I used to use Mustache on the server to that kind of logic.

In terms of where things belong ideally, I try to refrain from the ideal and just get really practical. When a class gets too long, I start to feel like I can't absorb what it is doing, and it must have many responsibilities. Then I try to figure out what those are. If those responsibilities are also somewhere else in the application then yanking them up to a shared space might make sense. Or moving them into a really small class that can be used in both places.

So long as things are small and well-named and not overly abstract, moving things around and changing abstractions is easy. So, I try not to think abstractly where something should be, but how to make things smaller and well named. Then it gets a lot easier to figure out what things belong together and how to get them to a better place.

adelevie:
rebased

adelevie:
looks wonderful to me :trumpet:

PR #285, opened 2016-02-10
Fix bid overflow issue

harrisj:
After some discussion with @msecret, I have made a few changes to the baseline CSS as part of fixing this issue in addition to some specific HTML/CSS issues. These changes can be summarized as follows:

* Set the base font-size to `16px`;
* Set the font size for other elements to `1.0625rem` (17px) since works better with Source Sans Pro
* Rescale most `rem` measurements for different sizing
* Eliminate use of the `em` measurements when possible
* Eliminate the use of negative margins, since those are what cause overwriting on mobile view

Because our CSS is not necessarily scoped to single views, these changes can have effects across the whole site. But I've been looking at things and I think we're in a good place. This does not fix two issues that are solved in prior PRs or should be a new PR:

* The buttons on the submit bid page are stacked (see #278 for solution)
* The login / edit profile buttons look terrible on mobile

I can apply fixes for those as well here, but I wanted to put this here first.

### DIFF
--- a/app/assets/stylesheets/main.scss
+++ b/app/assets/stylesheets/main.scss
@@ -122,7 +122,6 @@ textarea#auction_description {
 
     p {
       padding: 1rem 3rem 0 3rem;
-      font-size: 1.7rem;
       margin: 0;
     }
 
--- a/app/assets/stylesheets/main.scss
+++ b/app/assets/stylesheets/main.scss
@@ -13,6 +13,18 @@ $color-gold: #fdb81e;
 $color-cool-blue: #205493;
 $color-purple: #4c2c92;
 
+html {
+    font-size: 16px;
+}
+
+body {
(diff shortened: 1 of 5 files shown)


app/assets/stylesheets/main.scss
```diff
@@ -150,43 +165,42 @@ textarea#auction_description {
...
-
+        margin-bottom: 0.5rem;
       }
     }
 
     a.usa-button {
-      margin-top: 1.5rem;
-      padding: 0.5rem 2rem;
-      white-space: nowrap;
+        margin-top: 0;
```
msecret: There's an extra tab/space for these properties, css should probably default to 2 spaces.

msecret:
The CSS for this looks fine.

adelevie:
looks fantastic to me

PR #289, opened 2016-02-17
Move cookies.js out of the vendor folder

adelevie:
The `vendor` folder is ignored by cloud foundry, so the js file included here wasn't making it into the deployed application (breaking the deploy). This PR copies that file into the `app/assets` folder.


harrisj:
Technically, the asset pipeline should copy it out of the vendor folder, so this is something that is probably misconfigured with that, but this is fine.

PR #292, opened 2016-02-18
Keep logs out of CF deployments

adelevie:
Logs were 500mb and were slowing deploys (especially local deploys).

This PR adds logs to the cfignore (straight copy of gitignore). We should look into symlinking.


harrisj:
I am not sure why we didn't symlink .cfignore to point to .gitignore, but we can explore that later.

PR #293, opened 2016-02-18
Use absolute path to call auction/win_header partial

adelevie:
This pull request fixes a production-only error, triggered by visiting Admin::AuctionsController#preview by using an absolute path to call a partial.


harrisj:
Looks good to me.

PR #298, opened 2016-02-18
Display auction title in auctions#show

adelevie:


monfresh:
Wanna add a test for this?

adelevie:
check now :)

spec/features/bidder_interacts_with_auction_spec.rb
```diff
@@ -20,6 +20,7 @@
       visit auction_path(@published_auction)
 
       expect(page).to have_text(@published_auction.description)
+      expect(page).to have_text(@published_auction.title)
```
monfresh: If you also wanted to make sure the title was inside an `h1` tag, you could do this:
```
have_css('h1', text: @published_auction.title) 
```

harrisj:
Looks good to me, although we should look into @monfresh's suggestion soon

PR #300, opened 2016-02-19
Cleanup 2016-02-19

harrisj:
Just some little things to cleanup here. Next step will be converting some tests into cucumber features


adelevie:
looks fantastic to me :train2:

PR #302, opened 2016-02-19
Fixed the large font issue and changed auction listing styling.

ddotrollins:


app/assets/stylesheets/main.scss
```diff
@@ -14,15 +14,15 @@ $color-cool-blue: #205493;
 $color-purple: #4c2c92;
 
 html {
-    font-size: 16px;
```
harrisj: I was advised by @msecret to do this

app/assets/stylesheets/main.scss
```diff
@@ -14,15 +14,15 @@ $color-cool-blue: #205493;
...
+    font-size: 62.5%;
 }
 
 body {
-    font-size: 1.0625rem;
```
harrisj: Would eliminating this `body` declaration be enough to solve the problem? Because for `p` it's doing 1.0625 * 1.0625 * 16px? But I guess that only pushes up the text to 18px

harrisj:
Anyhow, I am fine merging this for now and we can fix later if issue

allalala:
:+1:

PR #303, opened 2016-02-19
More cucumber features

harrisj:
Since @baccigalupi bit the bullet and added some Cucumber features for the SAM warning, I thought it might make sense to convert over other specs in `spec/feature` to full-fledged Cucumber features


harrisj:
I have replaced all of the `feature/*.spec` tests that were features with Cucumber ones. This could probably use some cleanup and review later by @baccigalupi, but it's a start. In the process of doing this I discovered a few little things:

* Travis was not running Cucumber files (just spec files)
* The `UpdateUser` model was not letting users change their names
* The `bidder_interacts_with_auctions` spec was a crazy mess

In some cases I ported the specs over to features and scenarios directly. In other cases like the meta tag specs, I just added additional checks to existing scenarios instead. This could probably use some level or reorganization and such, but it's a start.

Note that this means we could now start new features by writing the cucumber features for them!

baccigalupi:
This looks really good. Something we should do in future commits is a little refactoring. The instance variables can get really out of control, pretty fast. You can make a context object, just an OpenStruct or something simple to tuck all the instance variables.

Also, when you have a lot of similar step groups, you can build a step that combines them. They are still all separate steps, very modular, but they are combined under one more detailed step. Another option for step aggregation when all the setup steps in several scenarios are the same is to have a `background` section for the feature. Not always possible because of small differences between scenario setups.

features/admin_auctions.feature
```diff
@@ -0,0 +1,42 @@
...
+    Then I expect to see the auction deadline
+
+  Scenario: Updating an auction
+    Given I am an administrator
+    And there is an open auction
+    When I sign in
+    And I visit the auctions admin page
+
+    When I click to edit the auction
+    Then I should be able to fill out a form for the auction
```
allalala: "Then I should be able to edit the auction form"

features/admin_auctions.feature
```diff
@@ -0,0 +1,42 @@
...
+
+    When I click on the Update button
+    Then I expect to see the new title for the auction
```
allalala: Then I expect to see my edited auction content on the page.

features/admin_auctions.feature
```diff
@@ -0,0 +1,42 @@
...
+
+    When I click on the auction title
+    Then I expect to see the changes
```
allalala: When I click on the auction title link for the auction I edited.
Then I expect to see new content on the page.

features/closed_auctions.feature
```diff
@@ -0,0 +1,42 @@
+Feature: Closed Auctions
+  As a contractor
```
allalala: As a vendor

harrisj:
Thank you @baccigalupi. I have been looking for a style guide for steps and they have been hard to find (there is also a lot of pushback against Cucumber in general, which I get if it's just developers working on this project, but...). I think I would like to figure out at some point:

1. If I should be doing a layer of abstraction on top of the button names (so I would say `click on the bid button`
2. Should it be `I should see` or `I expect to see`?
3. Should I use possessive apostrophes? Been avoiding it because I don't like single quotes
4. I'd like the Auction setup to actually assign to an auctions hash so it would be something like `auctions: {open: <Auction>, closed: <Auction>} then you could dereference it by saying the `open auction title`, but this is probably overkill
5. Should we be also crafting some feature files for high-level rules/behavior? These are all pretty specific to low-level controller actions

adelevie:
squashed everything here: https://github.com/18F/micropurchase/pull/342. Closing in favor of that.

PR #306, opened 2016-02-22
Don't show bid button if auction is over

cscairns:


allalala:
Thank you for this PR @cscairns!

harrisj:
Let me know if you want an overview about how some of these classes are organized, because it's a little unusual if you are used to a conventional Rails app

cscairns:
thx @harrisj...i'm picking it up as a work through some dev items...how presenters and viewmodels are being used within the context of this app are a little new to me (e.g., the presenters actually seem more like decorators)

jessieay:
This problem was solved elsewhere. Closing!

see https://18f.slack.com/archives/micropurchase/p1461013049000182 for more details

PR #308, opened 2016-02-22
Single bid auctions

adelevie:
This PR does a lot to accomplish one high level goal: implement single bid auctions.

Single bid auctions have three main differences than multi bid auctions:
- users can bid no more than once
- new bids need not be the lowest
- all bid information (including the existence of any bids) is veiled until the auction ends

The single bid limitation required changes and tests around `place_bid.rb` to disallow repeat bidding, as well as disabling the min bid validations.

The veiling required changes to the UI and the API.

In the UI, we needed to bifurcate behavior depending on whether the auction was single bid or multi bid. If multi bid, the behavior should be the same. If single bid, all bid info must be veiled during the auction.

In the API, we needed to rethink a bit how the AuctionSerializer worked. Before, `bids` came from a `has_many` relationship. This worked fine for multi bids, as the veiling occurred at the _attribute_ level. For single bids, however, we needed veiling to occur at the object level. This meant we eliminated the association and defined a method in the AuctionSerialzier, `#bids`, which explicitly passed `#veiled_bids` into the `BidSerializer`.

This PR is well-tested, but we're going to need to do some cleanup and refactoring afterwards.


adelevie:
[noting that @harrisj reviewed this code]

PR #309, opened 2016-02-22
Update admins.yml

batemapf:
adding myself as an admin --- and doing a happy dance that i don't have the highest github id!


adelevie:
welcome, fellow admin.

PR #311, opened 2016-02-23
JSON schema specs

pkarman:


jessieay:
@pkarman there's a CI failure here - want to update? should I take over?

pkarman:
@jessieay the test is intentionally failing. this was in response to https://18f.slack.com/archives/dev/p1456254837000302 and the preceding conversation.

jessieay:
@adelevie what do you want to do with this?

adelevie:
I'm ok with closing and prioritizing some sort of improvement/fix to the way we use json schema.

jessieay:
ok created an issue that references this PR so we can prioritize working on this tech debt item https://github.com/18F/micropurchase/issues/437

PR #327, opened 2016-03-02
Refactoring authentication

harrisj:
This has broken a few tests - most of which depended on mocking the internals of the application_controller class - and I could use some guidance on how to avoid passing in a controller object to the constructors, but this is an initial version for comment.


baccigalupi:
This is going in a cool direction. I would delegate more to the objects for example if each object were responsible for getting and holding on to the current user regardless of how that is achieved, no procs needed. Also, instead of two query methods that determine the request type, what about a factory that build the right object HTML or API auth depending on the request, then all other method are polymorphically sent down.

One more thing, it is a really great, zen-like practice to keep all your tests passing as you refactor. You have to move things out one by one. That is a practice that Katrina Owens champions

harrisj:
Yep, I would like to have the tests pass, but many of them were based on mocking internals of the ApplicationController that were no longer there so I will have to reimplement them in a different way.

I think I can refine it to inline those other methods (and make the call for set_current_user regardless of the authentication method). I still have a few issues I'd like to figure out with the code though (will annotate as comments on the lines)

app/controllers/application_controller.rb
```diff
@@ -9,40 +9,28 @@ class ApplicationController < ActionController::Base
...
 
+  delegate :require_authentication, :current_user, :github_id, :set_current_user,
+           to: :authenticator
+  
   rescue_from 'UnauthorizedError::MustBeAdmin' do |error|
     message = error.message || "Unauthorized"
     handle_error(message)
   end
 
+  rescue_from 'UnauthorizedError::RedirectToLogin' do |error|
```
harrisj: This is a case where we have an Authenticator raise an exception to have the controller take an action

app/controllers/application_controller.rb
```diff
@@ -53,56 +41,31 @@ def require_admin
...
+    @authenticator
   end
 
-  # rubocop:disable Style/AccessorMethodName
-  def set_current_user_to_api_user!(raise_errors: false)
-    user = User.where(github_id: github_id).first
 
-    fail UnauthorizedError::UserNotFound if user.nil? && raise_errors
-    @current_user = user
+  def html_request?
```
harrisj: These methods are just used in the `authenticator` factory above so I might just inline them there

app/controllers/application_controller/api_authenticator.rb
```diff
@@ -0,0 +1,39 @@
+class ApplicationController
+  class ApiAuthenticator < Struct.new(:controller)
+    attr_reader :current_user
+
+    def require_authentication
+      set_current_user(raise_errors: true)
+    end
+
+    def api_key
+      controller.request.headers['HTTP_API_KEY']
```
harrisj: I would like to figure out a better way to do this without passing the controller object to the authenticator. Or am I just overthinking this?

app/controllers/application_controller/web_authenticator.rb
```diff
@@ -0,0 +1,17 @@
+class ApplicationController
+  class WebAuthenticator < Struct.new(:controller)
+    def current_user
+      @current_user ||= User.where(id: controller.session[:user_id]).first
```
harrisj: Here I need the controller just so I can look at its session hash. Maybe I should just initialize with the session

spec/controllers/application_controller_spec.rb
```diff
@@ -1,140 +1,141 @@
 require 'rails_helper'
 
 RSpec.describe ApplicationController, controller: true do
-  describe '#require_authentication (via API)' do
```
harrisj: These tests need to be completely rewritten because they mock internals that are no longer there. Maybe get rid of this and replace with functional tests of the authentication

harrisj: This is now fixed

app/controllers/application_controller.rb
```diff
@@ -9,40 +9,20 @@ class ApplicationController < ActionController::Base
...
-      # otherwise, if it's an API request, we don't need to redirect.
-      require_authentication
-    end
-    is_admin = Admins.verify?(github_id)
-    fail UnauthorizedError::MustBeAdmin unless is_admin
-
-    is_admin
-  end
+  delegate :require_authentication, :require_admin, :current_user, :github_id, :set_api_current_user,
+           to: :authenticator
```
baccigalupi: Is there a reason why set_api_current_user is different that the other current user flow?
Also, do we need to expose the github_id?

app/controllers/application_controller.rb
```diff
@@ -53,57 +33,30 @@ def require_admin
...
+  def authenticator
+    if @authenticator.nil?
+      @authenticator = if html_request?
+                         WebAuthenticator.new(self)
+                       elsif api_request?
+                         ApiAuthenticator.new(self)
+                       else
+                         fail UnauthorizedError, "Unable to authenticate"
+                       end
+    end
```
baccigalupi: This is cool, what I was talking about with the one true place where the questions about request types were made. For that reason I would get rid of the #html_request? and #api_request? methods. Otherwise that exposure will encourage new code to use those methods instead of the authenticators you have setup.

app/controllers/users_controller.rb
```diff
@@ -11,7 +11,7 @@ def edit
   end
 
   def update
-    require_authentication and return
+    require_authentication or return
```
baccigalupi: If we are switching to raising errors when authentication fails, the return statement shouldn't be needed. So you should be able to drop the return statement

app/models/admins.rb
```diff
@@ -15,4 +15,9 @@ def self.github_ids
   def self.verify?(uid)
     new.verify?(uid)
   end
+
+  def self.verify_or_fail!(uid)
+    fail UnauthorizedError::MustBeAdmin unless verify?(uid)
+    true
+  end
```
baccigalupi: Also here and in the other authenticators if you don't have the return statement, you don't have to worry about returning true

baccigalupi:
Looks good! Think about the comments I threw in the mix, but they are more like finishing touches that real blockers

app/controllers/application_controller.rb
harrisj: I have to investigate this further, but current_user is lazy-evaluated for web requests and it seems like the API needs it to be actively loaded

```
  # for actions that don't require_authentication or require_admin,
  # we still need to set_current_user_to_api_user so that users can
  # optionally have authenticated access to public routes.
  # for example, /auctions/:id/bids will unveil bidder info about
  # the authenticated user. but the page also works fine sans authentication.
```

app/controllers/application_controller.rb
harrisj: I made this change in prior version, but I found there were a few places in other controllers using this method and it seemed one more thing to change for now. But I wouldn't be opposed to making this a future PR

app/models/admins.rb
harrisj: Good catch! Fixed.

adelevie:
Looks fantastic! Thanks @harrisj and @baccigalupi!

PR #328, opened 2016-03-02
Specify micro-purchase delivery deadline as a number of business days

sharjeelaziz:
We have used the following gems to complete this requirement:

https://github.com/bokmann/business_time
https://github.com/holidays/holidays

The code uses the business_time gem to calculate the delivery deadline date. The holidays configuration for the business_time gem is configured through the holidays gem. The delivery deadline day closes at 5:00 pm which the default time for the business_time gem.

### DIFF
--- a/lib/extensions/date.rb
+++ b/lib/extensions/date.rb
@@ -1,0 +1,8 @@
+# Blindly calculates end of workday using BusinessTime settings.
+# Does not consider weekends or holidays.
+class Date
+  def end_of_workday
+    cob = Time.parse(BusinessTime::Config.end_of_workday)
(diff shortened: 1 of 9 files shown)


adelevie:
Hi @sharjeelaziz, we accept the requirements. I'll contact you offline tomorrow about payment.

lib/extensions/date.rb
```diff
@@ -0,0 +1,8 @@
+# Blindly calculates end of workday using BusinessTime settings.
+# Does not consider weekends or holidays.
+class Date
+  def end_of_workday
```
adelevie: I think we should remove the monkey-patching before merging. Thoughts @harrisj, @baccigalupi?

PR #333, opened 2016-03-10
Ensure header is legible on mobile devices

jeremiak:
The site's header is hard to read on mobile devices because the logo, page title, and login button are on top of each other.

The small CSS changes here are to ensure that the header is legible on all devices

*Before*
screen shot 2016-03-09 at 7 47 55 pm

*After*
screen shot 2016-03-09 at 7 49 10 pm


adelevie:
Thanks @jeremiak!

jeremiak:
:+1:

PR #334, opened 2016-03-10
WIP: Auction type models

harrisj:
Refactoring the current auction logic to be cleaner and replace `if-else` with polymorphism and other techniques for cleaning up code. This is still early. **DO NOT MERGE**


harrisj:
So, I should elaborate a bit more on the plan. The general goal is to move the logic that changes depending on the auction type into separate models, so `Presenter::Auction` will still be used for basic general display functions (like showing a date properly, etc.) while things that might vary depending on the auction type (`user_can_bid?`, `bid_count`, even status) will be instead located in an appropriate subtype of `Policy::Auction` that wraps the `Presenter::Auction` and `Presenter::User` models and delegates down to them as appropriate. My next steps will be removing some functions from `Presenter::Auction` and moving them into the `Policy` classes. Then it will be time to make a similar determination for `PlaceBid` and any view partial methods.

harrisj:
So, this is taking much longer than I wanted because I perhaps am going about it all wrong. The general idea was:

1. I created `Policy::Auction` subtypes for `MultiBidAuction` and `SingleBidAuction` classes that wrapped a tuple of <Auction, User> and basically delegated display methods down to a Presenter but took methods for determining the auction status, bid validation, and such. I think this generally worked pretty well (see the code for [Policy::Auction](https://github.com/18F/micropurchase/blob/auction-types/app/models/policy/auction.rb) and [Policy::SingleBidAuction]
(https://github.com/18F/micropurchase/blob/auction-types/app/models/policy/single_bid_auction.rb), but then I fell into the weeds.
2. Right now for simplicity, I have the SingleBid and MultiBid auctions inherit/override a parent Policy::Auction type that in turn is a wrapper/delegates to Presenter::Auction and Presenter::User. This might not be the best approach.
2. Since I moved a lot of the logic around auction status into this Policy Object, this mean I had to create a corresponding `Policy::Bid` for wrapping a Bid presenter with the policy auction. This got a little awkward and it means [changing a lot of references all over the place](https://github.com/18F/micropurchase/blob/auction-types/app/controllers/bids_controller.rb#L33). So many tests still broken that need to be fixed, although I do have the PlaceBid, etc. working.
3. I think my Policy::Auction is a little too heavy and it might be better to instead keep the Policy objects for representing basic rules about auction availability and bid validation, but these would not wrap the User. Instead I would create an `AuctionSession` object that would do the delegation to the User and valid checking methods in the auction without needing to wrap a user in a bunch of contexts that don't need it.

Anyhow, this is where I am. Help appreciated in feedback. I'm going to take a little break from this PR to think it out.

app/controllers/bids_controller.rb
```diff
@@ -31,13 +30,14 @@ def new
   end
 
   def confirm
```
baccigalupi: There should really only be rest methods in a controller. Usually when you need non-rest methods, there is a missing concept that has its own subset of rest needs.

app/controllers/bids_controller.rb
```diff
@@ -31,13 +30,14 @@ def new
...
-    @auction = Presenter::Auction.new(Auction.find(params[:auction_id]))
-    @bid = Presenter::Bid.new(PlaceBid.new(params, current_user).dry_run)
+    @auction = Policy::Auction.factory(Auction.find(params[:auction_id]), current_user)
+    @bid = Policy::Bid.new(PlaceBid.new(@auction, params).dry_run, @auction)
```
baccigalupi: The PlaceBid#dry_run in the middle of a new bid seems funky. It reads as really complex ... making a new bid policy out of a fake place bid activity??? Probably some code is is the wrong place, and rearranging code is what design is all about!

app/controllers/bids_controller.rb
```diff
@@ -31,13 +30,14 @@ def new
...
   end
 
   def create
     fail UnauthorizedError, "You must have a valid SAM.gov account to place a bid" unless current_user.sam_account?
-    @bid = Presenter::Bid.new(PlaceBid.new(params, current_user).perform)
+    @auction = Policy::Auction.factory(Auction.find(params[:auction_id]), current_user)
+    @bid = Policy::Bid.new(PlaceBid.new(@auction, params).perform, @auction)
```
baccigalupi: This seems a little easier to understand, but is complexity that we don't need to see at this level. It could be that the bid policy takes in an auction and params to generate the bid, or .... not sure. Just seems like things are not quite in the right place.

app/models/place_bid.rb
```diff
@@ -35,65 +32,26 @@ def unsaveable_bid
 
   private
 
-  def auction
-    @auction ||= Auction.find(params[:auction_id])
-  end
-
   def auction_available?
-    Presenter::Auction.new(auction).available?
+    auction.available?
```
baccigalupi: These seem a lot easier to understand! We don't need to distinguish between an AR auction and a presenter at this level. We just get an object that can respond and tell us whether it is available.

app/models/policy/auction.rb
```diff
@@ -0,0 +1,263 @@
+class Policy::Auction
+  BID_LIMIT = 3500
+  BID_INCREMENT = 1
+
+  include ActionView::Helpers::NumberHelper
```
baccigalupi: If you can avoid includes like this, it would be good. When you include something this big into the object, it is no longer a PORO. I usually try to find a way to either directly call the module method, but Rails sometimes makes this very difficult. Or, I put all the inclusions in a separate class who can be used to isolate POROs from Rails dependencies. 

The place you can first feel the non-PORO pain is in tests. This object can't be tested with just rspec alone. It needs ActionView, and likely a lot of Rails to run. 

With a direct call to the module, you can mock out that object in pure specs. With an dedicated object that deals with those inclusions, you can mock out the whole object in tests.

app/models/policy/auction.rb
```diff
@@ -0,0 +1,263 @@
...
+    end
+
+    if amount > BID_LIMIT
+      fail UnauthorizedError, 'Bid too high'
+    end
+
+    if amount <= 0
+      fail UnauthorizedError, 'Bid amount out of range'
+    end
+  end
```
baccigalupi: This has become a really long class. Sandi rules cap the class at 100 lines and I have found 75 to be easier on comprehension. Now would be the moment to look for areas of cohesion and start pulling out smaller objects that will likely be reused outside this context.

app/models/policy/multi_bid_auction.rb
```diff
@@ -0,0 +1,57 @@
...
+  def display_type
+    'Multi-bid'
+  end
+  
+  def rules_href
+    '/auctions/rules/multi-bid'
+  end
+  
+  def info_box_partial
+    'auctions/multi_bid/info_box'
```
baccigalupi: Should this be absolute too?

app/models/policy/single_bid_auction.rb
```diff
@@ -0,0 +1,110 @@
...
+  # general bidding rules
+  def user_can_bid?
+    super && !user_is_bidder?
+  end
+
+  def bids?
+    return user_is_bidder? if available?
+    super
+  end
+
```
baccigalupi: Use of super is sometimes considered a smell. Usually there is a way to add another method to encapsulate the original behavior that super is doing. For example:

```
class Policy::Auction
  def can_user_bid?
    some_other_concept
  end

  def some_other_concept
    # original logic from can_user_bid?
  end

  # ...
end

class Policy::SingleBidAuction < Policy::Auction
  # ...
  def user_can_bid?
    some_other_concept && !user_is_bidder?
  end
  # ...
end
```

The main point is not to avoid super as much as to name missing concepts

app/models/policy/single_bid_auction.rb
```diff
@@ -0,0 +1,110 @@
...
+  def winning_bidder_name
+    winning_bid.name
+  end
+
+  def winning_bid_amount
+    winning_bid.amount
+  end
+
+  def bid_is_winner?(bid)
+    return false if bid.nil? || available?
```
baccigalupi: I would break out methods to name some of these conditionals

`return false if ___`

app/models/policy/single_bid_auction.rb
```diff
@@ -0,0 +1,110 @@
...
+    'auctions/single_bid/win_header'
+  end
+
+  def bid_input_partial
+    'auction_input'
+  end
+
+  def bid_alert_partial
+    'bids/single_bid/bid_alert'
+  end
```
baccigalupi: wondering about the mix of relative and absolute names. Also this class is tipping over the limit on size too.

app/models/presenter/auction.rb
```diff
@@ -269,7 +88,7 @@ def markdown
                                             fenced_code_blocks: true,
                                             lax_spacing: true)
     end
-
+    
     def model
       __getobj__
     end
```
baccigalupi: It looks like this class shrank a lot. Wondering what criteria was for moving code into the policy object?

app/models/presenter/bid.rb
```diff
@@ -81,6 +82,10 @@ def bidder_name
       def amount
         NULL
       end
+
+      def raw_amount
+        nil
```
baccigalupi: ??

Are there nil checks elsewhere that deal with this? If so might be better to find the default value. The goal of a Null object is to be able to answer all the questions you would ask to a normal object and get back reasonable answers.

app/models/presenter/user.rb
```diff
@@ -8,6 +8,14 @@ def admin?
       Admins.verify?(model.github_id)
     end
 
+    def sam_account?
+      model.sam_account
+    end
+
+    def to_param
+      model.to_param
```
baccigalupi: Is this needed for a Rails form?

app/models/view_model/auction_show.rb
```diff
@@ -3,62 +3,49 @@ class AuctionShow < Struct.new(:current_user, :auction_record)
...
       else
-        return number_to_currency(current_bid_amount)
+        return number_to_currency(highlighted_bid_amount)
       end
     end
 
-    def current_bid_amount
-      current_bid.amount rescue nil
+    def highlighted_bid_amount
+      highlighted_bid.amount rescue nil
```
baccigalupi: I know this code preceded this change set, but rescuing here is kind of messy. Is this dealing with the possibility that `highlighted_bid` may not exist? A guard clause would be better. Also, if you have a few methods dealing with the absence of a highlighted bid, you have a highlighted bid null class.

app/views/admin/auctions/show.html.erb
```diff
@@ -19,13 +19,13 @@
 <p><%=h raw @auction.html_description %></p>
 
 <div class='usa-grid'>
-  <% if @auction.current_bid? %>
+  <% if @auction.highlighted_bid? %>
```
baccigalupi: Null object? Keeping `if` statements out of the view is pretty important. You can move this to the presenter/view model by creating a null partial to handle the false case.

app/views/auctions/_auction_status.html.erb
```diff
@@ -1,4 +1,4 @@
-<span class="bid-amount"><%= @view_model.auction_current_bid_amount_as_currency %></span>
+<span class="bid-amount"><%= number_to_currency(@view_model.highlighted_bid_amount) %></span>
```
baccigalupi: Number helpers are already in lots of these internal classes. So, I would move this helper call out to the class responsible.

app/views/auctions/_list_item.html.erb
```diff
@@ -7,7 +7,7 @@
       <p class='issue-description'>
         <%=h raw auction.html_summary %><a href="<%= auction_path(auction) %>">View details »</a>
       </p>
-      <p><%= auction.formatted_type.capitalize %></p>
+      <p><%= auction.display_type %></p>
```
baccigalupi: :+1:

app/views/auctions/_list_item.html.erb
```diff
@@ -17,7 +17,7 @@
     <%= render partial: 'auctions/bids_and_form', locals: {auction: auction} %>
 
       <div class="usa-width-one-third">
-        <% if auction.user_can_bid?(current_user) %>
+        <% if auction.show_bid_button? %>
```
baccigalupi: Also an improvement! I would get the `if` statement out of the view too.

app/views/auctions/_winning_bid.html.erb
```diff
@@ -1,7 +1,7 @@
 <p>Bidding is closed</p>
 <% if auction.bids? %>
   <p>Winning Bid:</p>
-  <p><%= number_to_currency(auction.current_bid_amount) %></p>
+  <p><%= number_to_currency(auction.winning_bid.amount) %></p>
```
baccigalupi: This is better in the original form: law of demeter.
Also, again if/else in the view.

app/views/auctions/multi_bid/_win_header.html.erb
```diff
@@ -1,12 +1,12 @@
 <% if auction.over? %>
   <% if auction.bids? %>
     <% if current_user %>
-      <% if auction.user_is_winning_bidder?(current_user) %>
+      <% if auction.user_is_winning_bidder? %>
```
baccigalupi: This is an incredible amount of logic for a view. This was clearly here before, but if this massive change isn't improving situations like this, it is likely the wrong abstraction.

baccigalupi:
I am not convinced that this is going the right direction. With code like with product, I always ask, what is the pain point this is addressing. From looking at this change set (fully acknowledging that I haven't dived into the code in a while on my machine), it seems like the place of highest complexity and debt is in the views. This is a really theoretical change set that doesn't improve that complexity enough. 

The wrong abstraction is much more damaging than duplication, so I would recommend slowing down, rethinking and figuring out how to reduce the complexity already present.

I think if I tackled a general refactor, I would start with the massive if/else statements in the views and try to figure out what the missing abstractions are.

I know that we talked about polymorphism as a way to handle different types of things, but the objects that have been abstracted have a lot of if/else or case logic still, which says to me that the wrong concepts are being extracted.

I have worked too many hours this week, but ping me Monday if you want more in person information. Writing is a hard medium, and I don't want to come off as super-critical of your work, which I know was done is earnest!

app/controllers/bids_controller.rb
harrisj: Indeed. This was just a way to dry-run the user bid for a Confirmation, but it's a bit weird now

app/models/place_bid.rb
harrisj: Yes, I want the PlaceBid to just get an object. It was a `Policy::Auction`, but I think it makes more sense to define an `AuctionSession` that wraps the policy for a specific Auction type and the current user and can do the bid validation

app/models/policy/auction.rb
```diff
@@ -0,0 +1,263 @@
...
+  def self.factory(auction, user)
+    case auction.type
+    when Policy::MultiBidAuction::TYPE_CODE
+      Policy::MultiBidAuction.new(auction, user)
+    when Policy::SingleBidAuction::TYPE_CODE
+      Policy::SingleBidAuction.new(auction, user)
+    else
+      fail "Unsupported auction type: #{auction.type}"
+    end
+  end
```
harrisj: This is ugly, but I didn't want to create a separate Factory class for returning the appropriate `Policy::Auction` subtype. Should I instead not be using subclassing and mix-in base methods?

app/models/policy/auction.rb
harrisj: Indeed. I basically wanted to pull out all the of the methods from `Presenter::Auction` that would vary depending on the auction (replacing if-else with polymorphism), but I also fell into user-specific bid information and bid validation. Going to figure out a better way to do that.

app/models/presenter/auction.rb
harrisj: My main metric was if it would change depending on the type of the auction, it should be moved into a `Policy::Auction` subtype (or the parent class). This meant moving the following types of methods:

1. Do I show the bids?
2. What bids do I display in various places?
3. How do I denote winning bids?
4. The auction status

These vary depending on if the auction is single-bid vs. multi-bid, or they call a method like `available?` which might vary for future auctions with a "Buy It Now" button for instance.

Since some of the bid display and other options are dependent on the current user (ie, you show the current user's bid if the user has made one), I decided to make the `Policy::Auction` class encapsulate a reference to a user. This meant I also moved over:

1. Methods that returned a user's bids
2. The `user_can_bid?` method
3. I added a new `show_bid_button?` method

So that's what happened. There are probably better ways of structuring this, but I would like the `Presenter::Auction` method to jsut contain helper methods for rendering auction fields and move the other logic to more appropriate objects.

app/models/presenter/bid.rb
harrisj: Well, I thought it would be advantageous to return a Presenter::Bid object in several places, but I don't think `Presenter::Bid::Null.nil?` returns true and it probably shouldn't?

app/models/presenter/user.rb
harrisj: Exactly, It's needed by the `url_for` helper used in building routes.

app/views/admin/auctions/show.html.erb
harrisj: I think this would be a future refactoring indeed. I just wanted to rename `current_bid` to something a bit more descriptive

harrisj:
Abandoning this pull request

adelevie:
:pull-request-viking-funeral:

PR #345, opened 2016-03-21
SAM check on login, user edit of DUNS number

harrisj:
Makes SAM checking part of the login path, also if the user edits their DUNS number


adelevie:
Thanks, @harrisj!

PR #346, opened 2016-03-22
Added a Presenter::AuctionUser class (and Null model version)

harrisj:
Following the excellent (as usual) advice of @baccigalupi, I added a subsidiary presenter class for cleaning up the user_bids and other methods. In addition, I included a cleanup Kane suggested for the date presenters. Future work to be in other PRs

### DIFF
--- a/app/models/presenter/auction.rb
+++ b/app/models/presenter/auction.rb
@@ -191,25 +191,28 @@ module Presenter
       bids.sort_by(&:amount).first.amount
     end
 
+    def auction_user(user)
+      return Presenter::AuctionUser::Null.new if user.nil?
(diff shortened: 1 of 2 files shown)


adelevie:
@harrisj hate to be a stickler, but what would be a good way to test these? I think all the code is :+1:

harrisj:
I guess I assumed that current tests would cover this refactoring, but I will check on the coverage to be sure.

adelevie:
That's a good point. We can add additional tests, if we decide to, in a later PR.

harrisj:
Well I am about to go check on coverage so I will check

PR #351, opened 2016-03-23
Change path to codeclimate-batch on Travis

harrisj:


harrisj:
Forgive me, but I am merging this because I think this might be the answer

PR #354, opened 2016-03-24
Admin Auction Presenter

harrisj:
At @baccigalupi's suggestion, looking into a separate Presenter for AdminAuction so I can move some admin-only methods to that presenter (but there don't seem to be many). Might consider making Presenter::Auction an explicit delegator since there are some methods it's handling that are meant for admins only

### DIFF
--- a/app/controllers/admin/auctions_controller.rb
+++ b/app/controllers/admin/auctions_controller.rb
@@ -3,7 +3,7 @@ module Admin
     before_filter :require_admin
 
     def index
-      @auctions = Auction.all.map {|auction| Presenter::Auction.new(auction) }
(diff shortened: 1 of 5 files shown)


baccigalupi:
This looks pretty small and benign. I have run into trouble choosing delegation over inheritance around maintenance in these cases where most of the messages are delegated down. Keeping up with changes is just tedious.

If the whole point of the AdminAuction presenter is to separate out those methods that are specific to the admin view, what about inheriting from the Auction presenter?

harrisj:
That's a good point. I've been resisting my OOP-imposed tendency towards inheritance but that probably makes more sense here for those very reasons. Will revise.

harrisj:
I have changed Presenter::AdminAuction to be a subclass of Presenter::Auction

PR #357, opened 2016-03-28
Meta tags

stvnrlly:
We had removed the meta tags shortly after adding single-bid auctions due to information leakage. This should account for that and show the bidding as `Sealed` for those auctions.


adelevie:
:trumpet:

PR #359, opened 2016-03-29
A few more tweaks to the coverage

harrisj:
Working on debugging the test coverage some more


harrisj:
Trying one last thing before I kick it back to CC support

PR #385, opened 2016-04-11
Add Michael Torres to admins.yml

adelevie:


harrisj:
Looks good to me. Merging

mtorres253:
Hi @adelevie and @harrisj I see that I've been added as an admin. Should I have admin privileges when I login via github? If not, what are the steps for me to login to see how to create an auction (on staging) etc.

PR #408, opened 2016-04-13
Current bid amount not showing if not logged in

### LINKED ISSUE #407: Winning bids do not display on the main page
The template is looking for `current_bid_amount`, but that was lost at some point.

harrisj:
Refactoring some code exposed a latent bug that would prevent non-logged-in users from seeing current and winning bid amounts. Fixed the bug and added some tests as well.

Fixes #407 cc @stvnrlly


stvnrlly:
Checked it out locally and it looks good.

PR #413, opened 2016-04-14
Eliminate current_bid

harrisj:
I decided it's time to get rid of `current_bid` since it's usually used for two distinct purposes:

* The lowest bid (ie, the one that will win when the auction is over)
* The bid we display as a summary of the auction in views

This was done by actually overriding `current_bid` in the views with a version of the method that looked at the auction type and returned nil if the auction was a sealed-bid one and still running. I copied this code over to `ViewModel::Auction` but it was buggy and confusing. This PR does a few little things:

1. Rename `Presenter::Auction#current_bid` to `lowest_bid` and clean up some related method names
2. Rename `Presenter::Auction#current_max_bid` to `max_allowed_bid`
3. Define `ViewModel::Auction#highlighted_bid` for the bid to be displayed on the index or auction page
4. Replace the `rescue nil` checks in Presenter::Auction with polymorphism on the Presenter::Bid classes
5. Eliminate the `winning_bidder` method since nobody is using it
6. Removing the `winning_bidder` line in AdminAuctions#show since it was confusing


app/models/presenter/auction.rb
```diff
@@ -37,24 +28,20 @@ def current_max_bid
...
     def bids?
       bid_count > 0
     end
 
     def bids
-      model.bids.to_a
-           .map {|bid| Presenter::Bid.new(bid) }
-           .sort_by(&:created_at)
-           .reverse
+      @bids ||= model.bids.to_a
```
baccigalupi: Why the change to caching? Was there an issue?

app/models/presenter/auction.rb
```diff
@@ -126,21 +113,17 @@ def expiring?
...
 
-    def winning_bidder
-      winning_bid.bidder rescue nil
-    end
-
     def winning_bid
-      return single_bid_winning_bid if single_bid?
-      return multi_bid_winning_bid  if multi_bid?
+      return Presenter::Bid::Null.new if single_bid? && available?
+      lowest_bid
```
baccigalupi: Cool!

app/models/presenter/auction.rb
```diff
@@ -126,21 +113,17 @@ def expiring?
...
     end
 
     def winning_bidder_id
-      winning_bid.bidder_id rescue nil
+      winning_bid.bidder_id
```
baccigalupi: Love the null!

app/views/auctions/_winning_bid.html.erb
```diff
@@ -6,7 +6,7 @@
 <% if auction.bids? %>
   <div class="issue-bids-info-item"><% if defined? for_winners_page %><% else %>
   	<div class="issue-bids-info-item">Winning bid: </div>
-	<% end %><%= number_to_currency(auction.current_bid_amount) %></div>
+	<% end %><%= auction.highlighted_bid_amount_as_currency %></div>
```
baccigalupi: Yay form decreasing the complexity of views and removing these method calls!

app/views/bids/multi_bid/_highlighted_bid_info.html.erb
```diff
@@ -1,5 +1,5 @@
 <h2 class="center">Current bid:</h2>
-<p class="current-bid-amount center"><%= number_to_currency(@auction.current_bid_amount) %></p>
+<p class="current-bid-amount center"><%= number_to_currency(@auction.highlighted_bid_amount) %></p>
```
baccigalupi: Couldn't this use the #highlighted_bid_amount_as_currency method too?

baccigalupi:
Very cool. Love the renaming for clarity!

It would be easier to review if each refactor part had a different small branch/pull request. I am specifically thinking about the null change for winning bidders was great, but didn't seem related.

app/models/presenter/auction.rb
harrisj: I just felt like caching it since I use it a few places

app/views/auctions/_winning_bid.html.erb
harrisj: Yes it should. I need to more thoroughly hunt them down everywhere

harrisj:
I agree on more atomic PRs but the issue is that eventually the necessary latency of PR review means I get blocked if I go too small. I guess I could make each change PR assume the prior PR is accepted, but that might have its own issues.

PR #419, opened 2016-04-15
Auction Rules objects

harrisj:
So, I thought it might be interesting to explore getting rid of the `single_bid?` and `multi_bid?` flags and use polymorphism on that. This let me clean up the `PlaceBid` object a bit but I'm still in the point of needing those helper method to pass tests and figuring out a good way to handle the partials.

What do you think @baccigalupi and @adelevie, is this a good direction or a bridge too far? Would we have to move the partial helpers into these classes?

### DIFF
--- a/app/controllers/application_controller.rb
+++ b/app/controllers/application_controller.rb
@@ -11,8 +11,14 @@ class ApplicationController < ActionController::Base
   # the authenticated user. but the page also works fine sans authentication.
   before_action :set_api_current_user
 
-  delegate :require_authentication, :require_admin, :current_user, :github_id, :set_api_current_user,
-           to: :authenticator
+  delegate(
+    :current_user,
+    :github_id,
+    :require_admin,
+    :require_authentication,
+    :set_api_current_user,
+    to: :authenticator
+  )
 
   rescue_from 'UnauthorizedError::MustBeAdmin' do |error|
     message = error.message || "Unauthorized"
--- a/app/models/presenter/auction.rb
+++ b/app/models/presenter/auction.rb
@@ -59,6 +59,9 @@ module Presenter
       to: :auction_status
     )
 
+    delegate :winning_bid, :veiled_bids, :single_bid?, :multi_bid?, :formatted_type,
+             to: :auction_rules
+        
     def bids?
       bid_count > 0
     end
@@ -68,21 +71,6 @@ module Presenter
         .map {|bid| decorated_bid(bid) }
         .sort_by(&:created_at)
         .reverse
-    end
-
-    def veiled_bids(user)
-      # For single bid auctions, we reveal no bids if the auction is running
-      # For multi bid auctions, we let the bids go through, but depend on
-      # Presenter::Bid to veil certain attributes.
-
-      # redact all bids if auction is still running and type is single bid
-      if available? && model.single_bid?
-        return [] if user.nil?
-        return bids.select {|bid| bid.bidder_id == user.id}
-      end
-
-      # otherwise, return all the bids
-      bids
     end
 
     def bid_count
@@ -97,14 +85,14 @@ module Presenter
       Presenter::DcTime.convert_and_format(model.end_datetime)
     end
 
-    def formatted_type
-      return 'multi-bid'  if model.type == 'multi_bid'
-      return 'single-bid' if model.type == 'single_bid'
-    end
+    # def formatted_type
+    #   return 'multi-bid'  if model.type == 'multi_bid'
+    #   return 'single-bid' if model.type == 'single_bid'
+    # end
 
-    def type
-      model.type
-    end
+    # def type
+    #   model.type
+    # end
 
     def starts_in
       time_in_human(model.start_datetime)
@@ -116,10 +104,6 @@ module Presenter
 
     def delivery_deadline_expires_in
       time_in_human(model.delivery_deadline)
-    end
-
-    def winning_bid
-      decorated_bid(model.winning_bid)
     end
 
     def lowest_bid
@@ -151,6 +135,21 @@ module Presenter
 
     private
 
+    def auction_rules_class
+      type = case model.type 
+             when 'single_bid', 'sealed-bid'
+               'SealedBid'
+             else
+               'Basic'
+             end
+  
+      "::Rules::#{type}".constantize
+    end
+
+    def auction_rules
+      @auction_rules ||= auction_rules_class.new(self)
+    end
+    
     def markdown
       # FIXME: Do we want the lax_spacing?
       @markdown ||= Redcarpet::Markdown.new(Redcarpet::Render::HTML,
--- a/app/models/presenter/auction.rb
+++ b/app/models/presenter/auction.rb
@@ -12,19 +12,43 @@ module Presenter
       @auction = auction
     end
 
-    delegate :title, :created_at, :start_datetime, :end_datetime,
-             :github_repo, :issue_url, :summary, :description,
-             :delivery_deadline, :start_price, :published, :to_param,
-             :model_name, :to_key, :to_model, :type, :id,
-             :read_attribute_for_serialization,
-             to: :model
+    delegate(
+      :created_at,
+      :delivery_deadline,
+      :description,
+      :end_datetime,
+      :github_repo,
+      :id,
+      :issue_url,
+      :model_name,
+      :published,
+      :read_attribute_for_serialization,
+      :start_datetime,
+      :start_price,
+      :summary,
+      :title,
+      :to_key,
+      :to_model,
+      :to_param,
+      :type,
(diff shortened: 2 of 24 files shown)


app/models/presenter/auction.rb
```diff
@@ -33,6 +25,10 @@ def max_allowed_bid
     delegate :bidder_name, :bidder_duns_number,
              to: :lowest_bid, prefix: :lowest
 
+    delegate :winning_bid, :veiled_bids, :single_bid?, :multi_bid?, :formatted_type,
```
jessieay: this is a small stylistic thing, but I find delegations and other long lists of attrs easiest to read when they are on their own lines and alphabetized, eg:

```ruby
delegate(
  :formatted_type
  :max_allowed_bid,
  :multi_bid?,
  :single_bid?,
  :user_can_bid?,
  :veiled_bids,
  :winning_bid,
  to: :auction_rules
)
```

what do you think?

jessieay:
this general approach looks good to me!

mtorres253:
Did we want to migrate from single_bid -> sealed_bid now or would that be too much

harrisj:
By migrating, I assume you mean in how we're naming the partials and elements on the screen. I think that could be part of this, but right now I'm just trying to keep the tests passing and the PR from getting too large, but let me look into it.

mtorres253:
Let's not do the migration now then. I think this is just a refactoring story and not a user story.

harrisj:
Okay, I rebased the code for this now

app/controllers/application_controller.rb
```diff
@@ -11,8 +11,14 @@ class ApplicationController < ActionController::Base
   # the authenticated user. but the page also works fine sans authentication.
   before_action :set_api_current_user
 
-  delegate :require_authentication, :require_admin, :current_user, :github_id, :set_api_current_user,
-           to: :authenticator
+  delegate(
+    :current_user,
```
jessieay: 😭  (happiness)

app/models/rules/sealed_bid.rb
```diff
@@ -0,0 +1,53 @@
+module Rules
+  class SealedBid < Rules::Basic
+    def winning_bid
+      return Presenter::Bid::Null.new if auction.available?
```
jessieay: I would not expect an AR model to return a Presenter. Could we have this return a `NullBid` instead?

app/models/view_model/auction.rb
```diff
@@ -103,25 +95,23 @@ def auction_type
...
-        'bids/multi_bid/highlighted_bid_info'
-      end
+      auction.partial_path('highlighted_bid_info', 'bids')
     end
 
-    delegate :status, :label_class, :label, :tag_data_value_status,
-             :tag_data_label_2, :tag_data_value_2,
-             to: :status_presenter
-
+    delegate(
```
jessieay: could we throw these to the top of the file while we're here? I usually expect delegated methods to be near the top of ruby classes (not sure if this makes me Normal or not)

app/models/view_model/auction_status/open.rb
```diff
@@ -25,7 +25,8 @@ def tag_data_label_2
       end
 
       def tag_data_value_2
-        if auction.single_bid?
+        # Do we still need this?
```
jessieay: I don't know...but I am not sure the comment is helping me figure that out ;)

jessieay:
this is looking amazing! ✨ 

I think there's still more work to do with our presenters, but this is def a step in the right direction.

One change I'd like to see before merging is the returning of a presenter from one of the new bid classes. I would expect an `AuctionPresenter` to return a `BidPresenter` or `NullBidPresenter`, but not a Ruby class that is not view-logic related. Perhaps a good time to introduce `NullBid` ?

app/models/rules/sealed_bid.rb
harrisj: Well, this is wrapped in Presenter::Auction so it's not in the AR model level if that helps.

harrisj: But I will admit I'm not entirely sure of the difference between `Presenter::Bid::Null` and `NullBid`

jessieay: I imagine NullBid could be a plain Ruby object that is used to prevent nil
Vals in calculations rather than views.

We don't have one of these NullBids yet.

Reason I am suggesting change here is that I think it makes sense for view
objects or presenters to return presenters, but not things in "app/models",
which I would assume to contain "business logic" rather than "view logic".

Does my thinking make sense to you? Maybe we can have a quick chat about it
if not bc I am curious in your thoughts too!

harrisj: After a talk, we decided to defer that for another PR, but it makes sense to maybe have a NullBid and NullAuction for AR/model layers with corresponding presenters above them

app/views/auctions/single_bid/_auction_status.html.erb
```diff
@@ -1,3 +1,7 @@
-<span class="bid-amount">
-<%= @view_model.auction_user_bid_amount_as_currency %>
-</span>
+<% if @view_model.auction_won? %>
+  <%= render partial: "../multi_bid/auction_status" %>
```
harrisj: This is indeed ugly, but it's to avoid a bunch of functions instead of template_path, but we might reverse it.

harrisj:
We are almost there, but going to add some specific unit tests for the Rules classes first.

spec/models/presenter/auction_spec.rb
```diff
@@ -5,6 +5,136 @@
...
+        expect(lowest_bids).to be_an(Array)
+        expect(lowest_bids.first).to be_a(Presenter::Bid)
+        expect(lowest_bids.count).to eq(ar_lowest_bids.count)
+        expect(lowest_bids.map(&:id)).to eq(ar_lowest_bids.map(&:id))
+        expect(lowest_bids.map(&:bidder_id)).to eq(ar_lowest_bids.map(&:bidder_id))
+      end
+    end
+  end
+
+  # describe "#winning_bid" do
```
jessieay: remove commented tests?

spec/models/presenter/auction_spec.rb
```diff
@@ -5,6 +5,136 @@
   let(:ar_bids_by_amount) { ar_auction.bids.order('amount ASC, created_at ASC') }
   let(:auction) { Presenter::Auction.new(ar_auction) }
   let(:user) { FactoryGirl.create(:user) }
+  
+  describe 'internal bid methods' do
```
jessieay: We don't have to do this now, but for the future: it's easier for me to find / read unit tests if they are grouped by the method they are testing. eg:

```ruby

describe '#lowest_bid' do
  context 'when there are no bids' do
    # tests for that method
  end

  context 'when there are multiple bids for the same amount' do
    # tests for that method
  end
end
```

etc etc

spec/models/rules/basic_spec.rb
```diff
@@ -0,0 +1,66 @@
+require 'rails_helper'
+
+RSpec.describe Rules::Basic, type: :model do
```
jessieay: we shouldn't have to specify type bc we have `config.infer_spec_type_from_file_location!` in `rails_helper`

spec/models/rules/sealed_bid_spec.rb
```diff
@@ -0,0 +1,114 @@
...
+    context 'when the user has not placed a bid' do
+      let(:user) { FactoryGirl.create(:user, sam_status: :sam_accepted) }
+
+      it 'should return true' do
+        expect(subject.user_can_bid?(user)).to be_truthy
+      end
+    end
+  end
+
+  describe 'max_allowed_bid' do
```
jessieay: nice! loving this spec organization

spec/models/auction_spec.rb
```diff
@@ -30,34 +30,4 @@
       end
     end
   end
-
-  describe "#winning_bid" do
```
jessieay: can we resurface these tests for each of the rules types?

spec/models/presenter/auction_spec.rb
harrisj: Oh I hear you, it just is one of those cases where I have a bunch of complementary methods and I feel like making that determination to organize the other way, but right.

spec/models/presenter/auction_spec.rb
harrisj: Ooops, will fix

spec/models/auction_spec.rb
harrisj: I have done that, right?

jessieay: ah I wasn't sure you had! probably related to the "organization" comment. looks like these test s *are* here and I can see that much more clearly now. THANKS!

PR #420, opened 2016-04-15
Document the techniques we are using to deter technical debt

harrisj:
This should be ready to go. Please merge with a squashed commit.

### DIFF
--- a/TECHNICAL_DEBT.md
+++ b/TECHNICAL_DEBT.md
@@ -212,9 +212,14 @@ This isn't really worth it for a little repetition in two methods, but
 it can be very convenient if you find yourself repeating the same
 conditionals in many methods.
 
+Similarly, we have a few places where we branch depending on the
+auction type. This will likely make a good candidate for similar
+refactoring in the future.
+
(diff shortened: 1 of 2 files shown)


TECHNICAL_DEBT.md
```diff
@@ -0,0 +1,272 @@
...
+this project.
+
+If you don't now what a Decorator is, it's an object that wraps
+another object inside of it and can delegate some methods to the
+internal object directly and define new methods of its own.
+
+I'm going to look at the Auction model and the various classes around
+it as an example of how we are applying this approach:
+
+1. [Auction](app/models/auction.rb): this is the ActiveRecord class
```
jessieay: cool that this link works!

jessieay:
Content here looks great! My one piece of feedback is that it might make sense to include this doc in a `doc` dir rather than in the root dir. I am adding some onboarding docs in a separate PR and I am putting them in a `doc` dir. This way we can have a central place for all of our documentation.

adelevie:
Love this doc!

Issue #374, opened 2016-04-07
Payment URL validation


mtorres253:
As a vendor, I want the system to tell me if there is a problem with my credit card URL before it gets associated with my account, so if there is something wrong with it, the system should not save my entry and should notify me of the problem

Scenario: Vendor cannot save credit card URL with invalid format
Given I am an authenticated vendor
And I have entered a value for credit card URL on my profile page
But it is not in the form of a URL (http://www.domain.com for example)
When I click "Save"
Then the URL I entered should not be saved to my account
And I should see an error message "The URL you have entered is not valid"

Note: Let's discuss what is possible in addition to this. Do we want to require https:?

harrisj:
Can we have it just note the error on a Submit rather than when we tab out of it?

PR #433, opened 2016-04-19
Implement validation for the credit_card_form_url

harrisj:
This implements the functionality described in issue #374


app/models/user.rb
```diff
@@ -1,4 +1,25 @@
...
+  def validate_credit_card_form
+    return if credit_card_form_url.blank?
+
+    uri = URI.parse(credit_card_form_url)
+
+    case uri
+    when URI::HTTPS
+    # nothing to do here
+    when URI::HTTP
+      errors.add(:credit_card_form_url, 'must use HTTPS')
```
adelevie: We should communicate this requirement to vendors fairly soon, so everyone who doesn't have an HTTPS-backed form has a chance to fix it. Thoughts, @mtorres253 @vzvenyach?

adelevie:
LGTM

app/models/user.rb
```diff
@@ -1,4 +1,25 @@
...
   has_many :bids
-  validates :email, email: true, allow_blank: true
+
+  validates :email, allow_blank: true, email: true
+  validate :validate_credit_card_form
+
+  private
+
+  def validate_credit_card_form
+    return if credit_card_form_url.blank?
```
jessieay: what about checking `if credit_card_form_url.present?` instead ? I think that would be more easy to understand

features/credit_card_url.feature
```diff
@@ -30,3 +30,12 @@ Feature: Credit Card Form URL
 
     When I visit my profile page
     Then the new value should be stored as my Credit Card Form URL
+
+  Scenario: Vendor cannot save credit card URL with invalid format
```
jessieay: I don't usually write feature-level specs for "sad path" cases. Since feature specs are slower, filling a test suite with feature specs will inevitably lead to a super slow suite. We do want coverage for validations, but I think model-level specs should be enough.  What do you think?

spec/models/update_user_spec.rb
```diff
@@ -1,37 +1,44 @@
...
-  let(:updater) { UpdateUser.new(params, current_user) }
-  let(:current_user) { User.create(duns_number: '12341', sam_account: true) }
+  let(:user) { FactoryGirl.create(:user) }
+  let(:updater) { UpdateUser.new(params, user) }
 
+  let(:user_id) { user.id }
+  let(:user_name) { Faker::Name.name }
+  let(:user_email) { Faker::Internet.email }
+  let(:user_duns) { user.duns_number }
+  let(:user_credit_card_url) { 'https://random-example.com/pay' }
```
jessieay: I generally like to avoid `let` statements - leads to a mystery guest problem where you're referencing these vars in a long file. More here: https://robots.thoughtbot.com/lets-not

app/models/user.rb
```diff
@@ -1,4 +1,25 @@
 class User < ActiveRecord::Base
   has_many :bids
-  validates :email, email: true, allow_blank: true
+
+  validates :email, allow_blank: true, email: true
+  validate :validate_credit_card_form
```
jessieay: did you consider using a lib for this? https://github.com/perfectline/validates_url

That way, this could just be a one-liner. Might feel kind of heavy-handed to introduce a lib for this small thing (and I'd check on whether the lib itself is high quality first) but my guess is that, like with email addresses, there are a good number of edge cases here that our little method below might not cover. This is why I always use a lib for validation format of email addresses. Let someone else deal with those edge cases (in a well-tested way)!

app/models/user.rb
harrisj: I decided to add that myself. Should we relax this?

features/credit_card_url.feature
harrisj: I just wanted to take what was on the card, but you are right that I have also added some unit tests for this already.

app/models/user.rb
adelevie: I think it's worth doing. But it also means we should let vendors know way in advance. We could also relax the rule for now, let the vendors know, and in a month or so, put the rule in place.

We might also want to consider not preventing someone from bidding if they don't have that URL. We just make it clear we won't be able to pay them until a proper URL is submitted.

features/credit_card_url.feature
jessieay: yeah let's talk about the strategy! I know the team is using cucumber and I am OK with doing that but perhaps we should define when to use cucumber and when to use unit tests / controller tests only.

app/models/user.rb
harrisj: I might do this. I had looked at another lib and it was just a wrapper around URI.parse which doesn't do much. The only reason to not do this might be if we wanted to blacklist certain URLs, but I think we'd cross that bridge later with another validator.

harrisj:
Okay, I decided to use validates_url and to allow HTTP form URLs, so let's keep it simple for now.

jessieay:
`validate_url` stuff looks good! let me know what you think re: my remaining comments above. happy to discuss more or merge as-is

harrisj:
I think I want to keep a few of those lets there (but I took a few out). It might be good to look through our tests and see if it's too insane in some places.

jessieay:
there's a commented out test here -- was that an accident?

harrisj:
No, I just decided to relax the HTTPS requirement for now, but left that there if we wanted to bring it back

jessieay:
@harrisj do you mind removing it then? I find commented our pending tests to be very confusing because I never know if they *should* pass but don't or if they represent behavior that was not implemented (as in this case). IF we want to implement HTTPS in future, we can write a test then.

harrisj:
Done.

jessieay:
@harrisj thanks much! will merge as soon as CI passes 🍒

Issue #425, opened 2016-04-18
Add DUNS number validation on profile page


mtorres253:
As a registered vendor, I want to know whether the DUNS number I've entered for my profile is not formatted correctly, so that I can correct it before it gets saved to my profile, so the system should check to see if it is correctly formatted

Scenario: System checks that value for DUNS number is actually a number
Given that I am a registered vendor
And I have entered a DUNS number on my profile page
And the DUNS number is not formatted properly (it needs to be a number)
When I click the "Save" button
I should see an error message "The DUNS number you have entered is invalid: The value for this field has to be a number")
And my DUNS number should not be saved to my profile

jessieay:
One issue I am having with this is that we have a trait in our factories:


```
trait :with_duns_in_sam do
  duns_number { '130477032' }
end
```

I see that this duns_number is also in `data/auctions.json` -- what is the history behind this duns number? Looks like the rest of the `duns_number`s do start with 0 but that one does not and so it is invalid according to the rules in this story. 

@adelevie @harrisj do you know the history behind this special ❄️  duns? Is it the only one that won't start with a 0?

jessieay:
I take that back, there are other DUNs numbers in there that do not start with 0. Perhaps they were entered incorrectly in pre-validation times?

jessieay:
ah ha! it looks like formatting is slightly more complicated than we thought: https://github.com/18f/samwise/tree/micropurchase-compatibility#format-a-duns-number

must start with 0 if it is 8 digits!

jessieay:
my WIP branch https://github.com/18F/micropurchase/tree/jy-duns-number-validation

harrisj:
Oh nice. @jessieay, I will talk about Cucumber tomorrow with you

jessieay:
@mtorres253 if you see my comment above, the formatting referenced in the original GH issue description is not accurate. Will move forward with formatting validation but error msg and rules will be different.

mtorres253:
@jessie Let's just check to see if the number is at least 8 digits. I don't want to build duplicate validation logic in our system when there is an API we can use to validate the number. I'll add a separate story for that integration.

adelevie:
Did the docs/code in `samwise` clear this up, or is there any way I can be helpful here?

mtorres253:
Maybe @adelevie. According to `samwise`, the DUNS number format is 0XXXXXXXX0000 with the Xs being numbers. So a properly formatted DUNS number is:

1. A number
[...]

PR #438, opened 2016-04-19
Validate format of DUNs number

jessieay:
* Validate that it contains only integers
* Validate that it is 8,9, or 13 chars long (` Samwise::Util.format_duns(duns: value)`)

Closes https://github.com/18F/micropurchase/issues/425

### DIFF
--- a/app/models/user.rb
+++ b/app/models/user.rb
@@ -1,4 +1,5 @@
 class User < ActiveRecord::Base
   has_many :bids
   validates :email, email: true, allow_blank: true
+  validates :duns_number, format: { with: /\A\d{8,13}\z/ }
 end
--- a/spec/models/update_user_spec.rb
+++ b/spec/models/update_user_spec.rb
@@ -2,19 +2,20 @@ require 'rails_helper'
 
 RSpec.describe UpdateUser do
   let(:updater) { UpdateUser.new(params, current_user) }
-  let(:current_user) { User.create(duns_number: '12341', sam_account: true) }
+  let(:current_user) { FactoryGirl.create(:user, sam_account: true) }
+  let(:valid_duns_number) { '012345678' }
 
   let(:params) do
     ActionController::Parameters.new(
       id: user_id,
       user: {
-        duns_number: 'new-new'
+        duns_number: valid_duns_number
       }
     )
   end
 
   context 'when current user is not the same as the user being edited' do
-    let(:user) { User.create }
+    let(:user) { FactoryGirl.create(:user) }
     let(:user_id) { user.id }
 
     it 'raises UnauthorizedError' do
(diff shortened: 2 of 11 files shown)


app/models/user.rb
```diff
@@ -1,4 +1,5 @@
 class User < ActiveRecord::Base
   has_many :bids
   validates :email, email: true, allow_blank: true
+  validates :duns_number, format: { with: /\A\d{8,13}\z/ }
```
adelevie: https://github.com/18F/samwise/tree/micropurchase-compatibility#validate-the-format-of-a-duns-number

^ could be used instead

mtorres253:
The second validation was not in the story. And not entirely correct. The number should be 8, 9 or 13 digits (so not 10, 11 or 12 digits, for example). There is another story for this that has to be prioritized. However, I'm wondering how much more validation we will need if we can successfully tell the user whether their DUNS number exists in Sam.gov or not.

jessieay:
@mtorres253 

In this validation we are calling `Samwise::Util.format_duns(duns: '08-011-5718')`, which removes any dashes and adds zero padding if the DUNs number is 8 or 9 chars: https://github.com/18F/samwise/tree/micropurchase-compatibility#format-a-duns-number

If the number submitted is 10, 11, or 12 digits long, the samwise lib raises a `Samwise::Error::InvalidFormat` error and we return our error msg.
 
If the DUNs is the correct # of chars (8, 9, or 13), we then make sure that they are all integers. Unfortunately, the samwise lib does not do this for us so we are doing it here with a regex.

re: 

> wondering how much more validation we will need if we can successfully tell the user whether their DUNS number exists in Sam.gov or not.

I think this validation is helpful in the case that someone submits a typo or something.

spec/models/update_user_spec.rb
```diff
@@ -45,10 +46,44 @@
...
+          }
+        )
+
+        updater = UpdateUser.new(params, current_user)
+        updater.save
+
+        expect(updater.errors).to eq(
+          'DUNS number format is invalid'
+        )
+      end
```
harrisj: Can we also check that sam_account is set to false (even if it was true?)

harrisj:
This mostly looks good to me, but I think you're going to need to rebase it

spec/models/update_user_spec.rb
jessieay: ok!

jessieay: ah wait -- this is not hitting the SAM API -- we are validating the format. So we are not actually saving this new duns when someone submits it (we keep old value until they submit a potentially valid duns number, at which point we hit the API.

added spec expectation to clarify this behavior at 20b605d166498e1aace44de2668bd342f72e5103

jessieay:
@harrisj rebased!

spec/models/update_user_spec.rb
```diff
@@ -45,10 +46,46 @@
...
+          user: {
+            duns_number: nil
+          }
+        )
+
+        updater = UpdateUser.new(params, current_user)
+        updater.save
+
+        expect(updater.errors).to eq('')
+      end
```
harrisj: Should this also include a test that sam_account is reset to false?

app/models/authenticator.rb
```diff
@@ -16,7 +16,9 @@ def find_or_create_user
   end
 
   def update_user
-    user.update_attribute(:name, name) if user.name.blank?
+    if user.name.blank?
+      user.update(name: name)
+    end
```
harrisj: I'm fine with this, but we should probably change the Rubocop rules in another PR to be more lenient as well

jessieay: I refer to avoid trailing conditionals; is rubocop mad at me?

spec/models/update_user_spec.rb
jessieay: we have a test for that in `spec/models/sam_account_reckoner_spec.rb` (L65) -- seems that is more the concern of the reckoner.

jessieay:
Closing in favor of https://github.com/18F/micropurchase/pull/445

PR #439, opened 2016-04-19
Remove confusing `sam`-related traits from User factory

jessieay:
* Having a trait called `with_duns_in_sam` leads you to believe that
  there is some sort of mocked API that is returning that record
* We do not currently have a mock for the SAM API, this trait was
  therefore confusing / a red herring
* Also had problem of being a "magic number" https://en.wikipedia.org/wiki/Magic_number_(programming)


harrisj:
This looks on a good path and I will approve once we've fixed these tests...

spec/spec_helper.rb
```diff
@@ -1,100 +1,18 @@
...
-  # Setting this allows you to use `--seed` to deterministically reproduce
-  # test failures related to randomization by passing the same `--seed` value
-  # as the one that triggered the failure.
-  # Kernel.srand config.seed
-
-  # see https://docs.travis-ci.com/user/code-climate/#Common-Problems
-  config.after(:suite) do
-    WebMock.disable_net_connect!(allow: 'codeclimate.com')
-  end
+  config.order = :random
```
harrisj: I needed to bypass Webmock in the past so that it could report coverage to codeclimate. I am not sure if we still need this since we have a different flow now, but if this starts crashing on Travis after merged into develop that will be why.

jessieay:
@harrisj thanks for the input! these same specs are failing on `develop` - trying to figure out why now!

spec/spec_helper.rb
jessieay: good point, updated to `WebMock.disable_net_connect!(allow_localhost: true, allow: "codeclimate.com")` :)

jessieay:
merging, since we paired on this

PR #440, opened 2016-04-20
WIP: Setting up ActiveJob (with DJ backend) to check DUNS

adelevie:


app/jobs/check_duns_job.rb
```diff
@@ -0,0 +1,7 @@
+class CheckDunsJob < ActiveJob::Base
+  queue_as :default
+
+  def perform(user)
+    SamAccountReckoner.new(user).set!
```
jessieay: defining jobs can be helpful if you have complex logic -- since all we're doing here is calling a method, we can instead call it where it was being called before and update to read:

`SamAccountReckoner.new(user).delay.set!`

this tells DJ to handle it in the background. see: https://github.com/collectiveidea/delayed_job#queuing-jobs

adelevie:
@jessieay just implemented your suggestion.

jessieay:
there might be other ways to do it, but what I've done in the past to get DJ to work on Cloud.gov is add the worker to the Procfile and have Cloud.gov run that via foreman

see

https://github.com/18F/C2/blob/master/Procfile
https://github.com/18F/C2/blob/master/manifest.yml#L7
https://github.com/18F/C2/blob/master/script/start

features/step_definitions/page_steps.rb
```diff
@@ -1,3 +1,7 @@
+Then(/^I refresh the page$/) do
```
jessieay: this step appears to be unused

jessieay: nevermine I am wrong

jessieay:
closing in favor of https://github.com/18F/micropurchase/pull/446

PR #443, opened 2016-04-20
Parse VCAP_SERVICES on cloud foundry

stvnrlly:


adelevie:
Whoah

harrisj:
Holy cow. Does this work on deploy? And if so, should we remove the VCAP stuff in authenticator.rb or keep it there just as a belt-and-suspenders thing?

stvnrlly:
It _should_ work on deploy (i tested it with `cf-ssh`), and should make it possible to remove the CF-specific code. I'll see if this is something that I can do in this PR.

jessieay:
Would you mind adding something to `doc/deployment.md` about this? Seems like it might be a change in deploy behavior? 🙌

adelevie:
Do we load `sed` or `jq` anywhere? Or are these implicitly available? I worry that this could make the codebase a bit more brittle since it relies on dependencies we don't really control.

stvnrlly:
@jessieay: 👍 
@adelevie: `sed` comes with Linux and `jq` is loaded in the Cloud Foundry stack: https://github.com/cloudfoundry/stacks/blob/master/cflinuxfs2/build/install-packages.sh#L40

adelevie:
Nice. I'm gonna squash these.

adelevie:
Just created a `data-dot-gov` service for both staging and production environments and added the keys. I'll update the docs in a separate PR.

Issue #381, opened 2016-04-08
The system checks Sam.gov to validate a vendor's DUNS number


mtorres253:
As a vendor, I want signup to be as easy as possible, so if the system can validate profile data for me, it should.

Scenario: System checks Sam.gov to validate a vendor DUNS number
Given I am a registered vendor
When a new DUNS number is associated with my account
Then the system checks Sam.gov to validate my DUNS number (valid means that the number exists and is associated with the correct profile)
And checks every hour until it gets a response

e.g.
https://api.data.gov/sam/v1/registrations?qterms=legalBusinessName:Battle%20Plan%20Marketing%20Group&api_key=DEMO_KEY

PR #446, opened 2016-04-21
Rebased: Setting up ActiveJob (with DJ backend) to check DUNS

adelevie:
This would negate: https://github.com/18F/micropurchase/pull/440

- Sets up ActiveJob/Delayed job
- Delays a call to the `SamReckoner`

Fixes https://github.com/18F/micropurchase/issues/381

### DIFF
--- a/features/step_definitions/page_steps.rb
+++ b/features/step_definitions/page_steps.rb
@@ -1,3 +1,7 @@
+When(/^I refresh the page$/) do
+  visit page.current_path
+end
+
 When(/^I click on the "?([^"]+)"? button$/) do |button|
   # will click the first if there are two with same text!
   first(:link_or_button, button).click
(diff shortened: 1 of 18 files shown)


features/step_definitions/page_steps.rb
```diff
@@ -1,3 +1,7 @@
+Then(/^I refresh the page$/) do
```
jessieay: I think this should be a `When` rather than `Then`

jessieay:
this looks great to me! wanna try pushing it to staging manually before we merge to `develop` ?

adelevie:
Sure thing. We're fixing the env var situation first so new deploys have access to the right API keys.

adelevie:
Ready after:
- https://github.com/18F/micropurchase/pull/448 is merged
- this branch is rebased

jessieay:
this now has a merge conflict @adelevie

adelevie:
rebased

PR #448, opened 2016-04-21
Refactor GitHubCredentials into UserProvidedService

adelevie:
This PR refactors some common VCAP parsing code into its own class. It also creates a small PORO for loading getting the data.gov api key from the `data-dot-gov` user-provided-service

TODO: 

- ~~load the data.gov api key into `ENV['DATA_DOT_DOV_API_KEY']` in an initializer~~
- ~~test `UserProvidedService` and/or `DataDotGovCredentials`~~

### DIFF
--- a/app/models/data_dot_gov_credentials.rb
+++ b/app/models/data_dot_gov_credentials.rb
@@ -1,0 +1,11 @@
+class DataDotGovCredentials
+  extend UserProvidedService
+
+  def self.api_key(force_vcap: false)
+    if use_env_var?(force_vcap)
+      ENV['DATA_DOT_GOV_API_KEY']
+    else
+      credentials('data-dot-gov')['api_key']
+    end
+  end
+end
--- a/spec/models/data_dot_gov_credentials_spec.rb
+++ b/spec/models/data_dot_gov_credentials_spec.rb
@@ -1,0 +1,20 @@
+require 'rails_helper'
+
+RSpec.describe DataDotGovCredentials do
+  let(:vcap_json) do
+    JSON.parse(File.read('spec/support/vcap_services.json'))
+  end
+  let(:data_dot_gov_api_key) do
+    data_dot_gov = vcap_json['user-provided']
+      .find {|service| service['name'] == 'data-dot-gov'}
+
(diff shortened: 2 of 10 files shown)


adelevie:
I'm going to manually deploy this to staging to test it out (once tests pass).

jessieay:
@adelevie let us know when this is ready for review / merge!

adelevie:
Ready for review.

app/models/data_dot_gov_credentials.rb
```diff
@@ -0,0 +1,8 @@
+class DataDotGovCredentials
+  extend UserProvidedService
+
+  def self.api_key(force_vcap: false)
+    return ENV['DATA_DOT_GOV_API_KEY'] if use_env_var?(force_vcap)
```
jessieay: I personally find this format more readable for conditions:

```ruby
if use_env_var?(force_vcap)
  ENV['DATA_DOT_GOV_API_KEY']
else
  credentials('data-dot-gov')['api_key']
end
```

what do you think?

app/models/user_provided_service.rb
```diff
@@ -0,0 +1,22 @@
+module UserProvidedService
```
jessieay: since this is a module, what do you think about making it a concern? (concerns are always modules)  http://wendybeth.ghost.io/adding-modules-to-rails-4-models/

spec/models/data_dot_gov_credentials_spec.rb
```diff
@@ -0,0 +1,10 @@
+require 'rails_helper'
+
+RSpec.describe DataDotGovCredentials do
+  describe '::api_key' do
+    let(:api_key) { DataDotGovCredentials.api_key(force_vcap: true) }
+    it 'returns a string' do
+      expect(api_key).to be_a(String)
```
jessieay: can we update this to look for the exact string in `spec/support/vcap_services.json` ?

jessieay:
this looks great! just a few small comments from me

adelevie:
👍 to both suggestions. I'll implement them.

adelevie:
Ready again.

PR #449, opened 2016-04-21
Refactor logic around `winning_bid`

jessieay:
* This represents business logic and therefore makes sense to include in
  `auction`
* Related `status` logic was hard to parse, moved that into service
  class

### DIFF
--- a/app/models/auction.rb
+++ b/app/models/auction.rb
@@ -1,11 +1,29 @@
 class Auction < ActiveRecord::Base
   has_many :bids
   has_many :bidders, through: :bids
-  enum result: {not_applicable: 0, accepted: 1, rejected: 2}
-  enum type: {single_bid: 0, multi_bid: 1}
-  enum awardee_paid_status: {not_paid: 0, paid: 1}
-  enum published: {unpublished: 0, published: 1}
+  enum result: { not_applicable: 0, accepted: 1, rejected: 2 }
+  enum type: { single_bid: 0, multi_bid: 1 }
+  enum awardee_paid_status: { not_paid: 0, paid: 1 }
+  enum published: { unpublished: 0, published: 1 }
 
   # Disable STI
   self.inheritance_column = :__disabled
+
+  def winning_bid
+    if single_bid? && AuctionStatus.new(self).available?
+      nil
+    else
+      lowest_bid
+    end
+  end
+
+  def lowest_bid
+    bids.select {|b| b.amount == lowest_amount }.sort_by(&:created_at).first
+  end
+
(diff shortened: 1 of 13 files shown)


app/models/auction.rb
```diff
@@ -1,11 +1,29 @@
 class Auction < ActiveRecord::Base
   has_many :bids
   has_many :bidders, through: :bids
-  enum result: {not_applicable: 0, accepted: 1, rejected: 2}
-  enum type: {single_bid: 0, multi_bid: 1}
-  enum awardee_paid_status: {not_paid: 0, paid: 1}
-  enum published: {unpublished: 0, published: 1}
+  enum result: { not_applicable: 0, accepted: 1, rejected: 2 }
```
harrisj: I would like to convert these from enums to strings, but not urgent. Just noting this is a vague to-do of mine

app/models/auction.rb
```diff
@@ -1,11 +1,29 @@
...
   # Disable STI
   self.inheritance_column = :__disabled
+
+  def winning_bid
+    if single_bid? && AuctionStatus.new(self).available?
+      nil
+    else
+      lowest_bid
+    end
+  end
```
harrisj: I don't think winning_bid should be in the AR model because it depends on the auction type and I'd rather keep basic methods only in here

harrisj:
I feel like this and #419 are trying to do similar things and I'm not sure if we should hash out some hybrid approach between the two

app/models/auction.rb
jessieay: I like them as enums (get free methods like `not_applicable?`) -- what do you see as the benefit of strings?

app/models/auction.rb
jessieay: I think of `winning_bid` as a basic model -- this app at a basic level holds auctions and determines a winner, so I think elevating winner beyond view logic makes sense. We can move out of `auction`, but let's discuss !

jessieay:
I think we can def do a hybrid! What I am seeing in 419 is a breaking up of how bids are placed for single vs multi bids. I think we can also do that for how we determine winning bids. The main point of this PR was to highlight that `winning_bid` is as business rules concept, not something purely for the view / decorator. In the process of seeing how the presenter determined `winning_bid`, I saw all of the `status` stuff around auctions and decided that could go in its own service class, so that became part of this as well.

harrisj:
Looks good to me. I want to move `winning_bid` to Rule objects as in #419 eventually, but this is needed for a feature to be implemented and I don't want to block that.

PR #451, opened 2016-04-22
Make `start_datetime` and `end_datetime` fields non-null

jessieay:
* We should strive to have non-null data when possible
* At db and AR level

### DIFF
--- a/app/models/auction.rb
+++ b/app/models/auction.rb
@@ -9,6 +9,9 @@ class Auction < ActiveRecord::Base
   # Disable STI
   self.inheritance_column = :__disabled
 
+  validates :end_datetime, presence: true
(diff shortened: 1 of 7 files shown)


adelevie:
Should we get a PR that sets up auto-migrations in place before merging this? Or just merge and manually migrate?

jessieay:
@adelevie the `start` script in https://github.com/18F/micropurchase/pull/446 auto runs migrations! so we can merge this after that for sure

adelevie:
Whoah, I did add this (ctrl+c ctrl+v from CAP I think), but we don't actually have that `rake cf:on_first_instance` implemented. That would have been bad. I'll add it here.

PR #458, opened 2016-04-22
New VCAP parsing method

stvnrlly:
Rather than rely on some weird yml hack, this uses a new buildpack to do the parsing without having to touch the manifest.

Currently, it provides the user-provided credentials in this format: [name of service]_[key]=[value]. For example: `micropurchase_github_client_id=1234567890`. It's long, but prevents name overlap. We can either change the name throughout the app, or add `MPT_3500_GITHUB_KEY: $micropurchase_github_client_id` to the manifest.


adelevie:
This is pretty sweet.

jessieay:
@stvnrlly what do we need to do to get this merged? Docs would be helpful, I can add some if you can point me to where I can learn about this new parsing method.

stvnrlly:
The only thing remaining is to change the existing environment variables to match the way the buildpack sets them up. I think it was on hold while @adelevie set up the data.gov api keys so that the merge conflicts could be resolved all in one fell swoop.

More information about how things work should be in the buildpack's repo: https://github.com/18F/cf-env-parser-buildpack. I'll update the docs along with the name changes to mention the naming convention once I get the go-ahead.

jessieay:
this looks good to me! just needs a rebase and fix failing tests

stvnrlly:
I've rebased and fixed as many tests as I can. There are two failures left that I'm not sure about. It might make more sense for somebody familiar with the tests to fix those, if they're quick fixes. If not, let me know and I'll dig in.

jessieay:
Hey @stvnrlly this should be passing now

One failure was related to a super weird thing we saw earlier where setting email from via ENV var was resulting in a string in an array, whereas setting via VCAP json was setting it as JUST a string. So now tht we are using ENV vars we can look for the array version (either works...still scratching my head over why there was a difference. That update at 990e040

Second failure was a spec file left over in `spec/models` that was an outdated dupe of a spec file in `spec/credentials`  -- removed the old spec at bb272f8

This PR overall looks 💯 to me! Is there a reason we cannot save the json keys within vcap services as all caps? I am not sure there is a *technical* reason for this, but there is a strong convention for ENV vars (in Rails apps, at least) to be all caps. Having them lowercase might be confusing to people who are used to always seeing them as all caps.

stvnrlly:
The lowercasing is just what the buildpack does. I think I went that route because the credentials keys are lowercase in `VCAP_SERVICES`, but I agree that uppercase is the expectation. I'll change that in the buildpack next week.

jessieay:
this looks AMAZING! and ready to merge...looks like we need to update our user provided service on CF before merging / deploying. Otherwise 💯

stvnrlly:
If the update would be to change the case, that shouldn't be necessary: the buildpack should do the capitalization automatically. Still wouldn't hurt to update the user-provided service, though.

jessieay:
Oh wow so if the keys in VCAP services are lowercase this will still work?
COOL!

stvnrlly:
Hopefully! 😃

jmcarp:
Sorry for the drive-by, but I noticed this and had a few questions:

* It looks like buildpack-multi is being deprecated, and I don't think it's been vetted by the cloud.gov team.
* If you do use buildpack-multi, I would recommend pinning all the buildpacks in `.buildpacks` to a specific commit or tag. One of my apps got hosed a while back when the node buildpack got upgraded, and I had to pin to the last working release.
* I see that you're already using https://github.com/cloudfoundry/cf-app-utils-ruby, which seems to have some useful helpers for parsing `VCAP_SERVICES` and other cf envvars. Would that be a viable alternative to a custom buildpack?

By the way, I know literally nothing about this project, so apologies if my questions are off-base.

stvnrlly:
@jmcarp Pinning the versions is a great point. I'd also love to avoid using the multi buildpack in general and have been thinking about scripts to build the env vars at deploy time, but I think there's still too much potential for human error there. I also tried [this wacky thing](https://github.com/18F/micropurchase/blob/f2a242cfc6b9397068d6a4c34dd6b60646921533/manifest.yml) which didn't quite work.

As for the `cf-app-utils-ruby` gem: the idea of using this buildpack is to be able to keep `cf`-specific stuff confined to the manifest and associated files and out of the app logic itself, so that wouldn't quite be the same thing.

jmcarp:
@stvnrlly: would `CF::App::Environment.set!` or `CF::App::Environment.set_from_yaml!` help here? Looks like the idea is to copy creds from `VCAP_SERVICES` to top-level envvars, so that you wouldn't have to litter the code with calls to `CF::App::Credentials.find_by_service_name`.

stvnrlly:
I think we're trying to avoid the gem in general, so that wouldn't help, though in the grand scheme it may not be a big deal.

I've also been bugging @adelevie this morning with some other ideas on how to handle this with any kind of app, and may try some new stuff soon.

Issue #450, opened 2016-04-21
Pull in email address from github oauth?


jessieay:
As a user who is signing up for a micropurchase account through github, I want to have my profile information populated for me if possible, so if there is a way to get my email address from github, the system should do that.

Scenario: I sign up for a micropurchase account
Given I am a new micropurchase user
And I have an email address specified in github
When I signup via github
Then my email address should be assigned to my profile
And when I view my profile page I should see my email address pre-populated in the email address field

Scenario: I login to micropurchase using my github account
Given I am a micropurchase user (admin or vendor)
And I have an email address specified in github
And I do not have an email address specified in my micropurchase profile
When I signin via github
Then my email address should be assigned to my profile
And when I view my profile page I should see my email address pre-populated in the email address field

Omniauth requires "email" scope in order to pull email directly from github.

Currently, we do not request this scope

Is there a reason why we do not request this already?

https://github.com/intridea/omniauth-github

cc @adelevie @harrisj

jessieay:
from Slack:

Yeah, let’s request the email scope. Can we also get their their github login?

[5:02] 
I think we try to always request minimal info

Jessie Young (SF, she/her) [5:03 PM] 
what you mean login?

[5:03] 
these are available scopes https://developer.github.com/v3/oauth/#scopes

[5:03] 
I think email is a pretty good minimum

[5:03] 
it’s just read access

jessieay:
should look into storing github username, too. currently we do this as part of a separate call to github

jessieay:
Pretty sure we can pull both in during oauth process 

see `create_with_omniauth` method https://www.natashatherobot.com/rails-omniauth-github-tutorial/

mtorres253:
If this can help with the email story, let's do this now. I can also tweak the acceptance criteria if we can't do it exactly as described

jessieay:
PR for this here: https://github.com/18F/micropurchase/pull/460

PR #460, opened 2016-04-22
Pull github username, email from GitHub auth

jessieay:
* Add `email` scope to get access to email on login
* Also redirect to HTTP referrer on login rather than hard coding redirect
* Closes https://github.com/18F/micropurchase/issues/450


adelevie:
Looks good to me! I'll take another closer look either later today or Monday.

app/models/user.rb
```diff
@@ -4,4 +4,21 @@ class User < ActiveRecord::Base
   validates :credit_card_form_url, url: {allow_blank: true, no_local: true, schemes: ['http', 'https']}
   validates :duns_number, duns_number: true
   validates :email, allow_blank: true, email: true
+
+  def from_oauth_hash(auth_hash)
+    set_if_blank('name', auth_hash)
+    set_if_blank('email', auth_hash)
+    self.github_login = auth_hash[:info][:nickname]
```
jessieay: note: `nickname` is github username https://github.com/intridea/omniauth/wiki/Auth-Hash-Schema (super confusing)

adelevie: I feel like you're doing better at code review of your own code than I am ... I'm looking for things to point out!

jessieay: hahah never! I left this comment bc I found it to be super confusing when writing the code. oauth hash has a bunch of fields with `name` in the them and it took be a bit to figure out which one was `github_login` -- octokit response used `github_login` key but omniauth does not return that key!

app/models/api_authenticator.rb
```diff
@@ -0,0 +1,42 @@
+class ApiAuthenticator < Struct.new(:controller)
```
harrisj: When we reorganize things, these should probably become Policy models (I had scoped them before to only be accessible from controllers because I wanted it to be clear)

app/controllers/authentications_controller.rb
```diff
@@ -1,11 +1,22 @@
...
   end
 
   def destroy
     reset_session
     redirect_to '/'
   end
+
+  private
+
+  def path
```
harrisj: Do you think we should just have this method do the redirect_to as well and name it redirect_to_logged_in?

app/controllers/authentications_controller.rb
```diff
@@ -1,11 +1,22 @@
...
   def create
-    path = Authenticator.new(request.env['omniauth.auth'], session).perform
+    session[:return_to] ||= request.referer
+    Authenticator.new(request.env['omniauth.auth'], session).perform
     redirect_to(path)
   end
 
   def destroy
     reset_session
     redirect_to '/'
```
harrisj: Let's change this to `redirect_to root_path` as well?

app/models/authenticator.rb
```diff
@@ -1,24 +1,14 @@
 class Authenticator < Struct.new(:auth_hash, :session)
```
harrisj: This class was there before I added the ApiAuthenticator and WebAuthenticator models, and it doesn't function in quite the same way. I wonder if we should rename it to alleviate some potential confusion, especially since all three models are in the same directory. Maybe `LoginUser`?

app/models/github_reckoner.rb
```diff
@@ -1,27 +0,0 @@
-class GithubReckoner < Struct.new(:user)
```
harrisj: So happy we can get rid of the GithubReckoner. I clearly should do a better job of reading the docs. Thank you, Jessie!

app/models/authenticator.rb
```diff
@@ -1,24 +1,14 @@
...
-
-  def update_user
-    if user.name.blank?
-      user.update(name: name)
-    end
+  def user
+    @_user ||= User.find_or_create_by(github_id: auth_hash[:uid])
   end
 
   def sign_in
```
harrisj: I wonder if we should consider setting a timeout on the session cookie?

harrisj:
This is awesome and I am glad you were able to kill GithubReckoner. Unfortunately, I'm asking to fix a few other little things while you are in there because they were to-dos I never to-did.

app/models/api_authenticator.rb
jessieay: do you mind of we do that separately? I could create a card for it...

Seems like we might want to use something more detailed like `pundit` which would create a lot of noise in here... or maybe you just thought we should rename to `ApiAuthentcatorPolicy` ? (the latter could def be done in the context of this PR)

app/controllers/authentications_controller.rb
jessieay: I updated the logic / testing for this here: fc7b499 what do you think?

app/controllers/authentications_controller.rb
jessieay: done! at eec32ac (good call)

app/models/authenticator.rb
jessieay: love it! renamed at deacdde

app/models/authenticator.rb
jessieay: I myself am not too worried about it... maybe we should open a GH issue for this?

jessieay:
@harrisj think I've responded to all of your comments. Let me know what you think!

app/models/api_authenticator.rb
harrisj: Yeah, definitely make that step a future PR. Sorry for the confusion!

harrisj:
I think it (and you) are awesome. Merging!

app/models/api_authenticator.rb
jessieay: cool! added issue here https://github.com/18F/micropurchase/issues/468

PR #465, opened 2016-04-23
Add javascript linter

stvnrlly:
This adds a the AirBNB `.jshintrc` so that Code Climate will have something to use, and applies that style to most of the javascript. It doesn't catch _everything_, in part to see what score it gets without being perfect and in part because ¯\_(ツ)_/¯.


harrisj:
LGTM. Merging

Issue #426, opened 2016-04-18
System shows pending message when DUNS number is being verified with Sam.gov


mtorres253:
As a registered vendor, I want to know the status of my profile while it is being validated, so I should see a pending message while my profile is being validated

Scenario: System shows pending message while vendor DUNS number is being checked in Sam.gov
Given that I am a registered vendor
And I have entered a DUNS number in my profile
And the system is checking Sam.gov to verify my DUNS number
When I go to my profile page
Then I should see a message on my profile page "Your profile is pending while your DUNS number is being validated. This typically takes less than one hour" 

Note: Check with Alan to make sure the estimate is accurate

harrisj:
So, it seems like we have two ways of handling this, both of which would involve some DB alteration:

1. Change the sam_account column from a boolean to something like `sam_status` that is either `new`, `pending` or `found` (I guess we could have an option for explicitly rejecting too)
2. Add a `sam_checked` datetime column. If it's nil that means it's never been checked, otherwise it's the time of the last check for that user.

We could also of course do both if we think that makes sense. Thoughts?

jessieay:
@harrisj I like the idea of doing a `datetime` column. That also serves as a `true` / `false` checker for whether something has been checked against sam. We wouldn't have pending, but assuming this is running in a delayed job that is kicked off as soon as someone enters their same acct number, these things would not be pending for long anyway.

adelevie:
I like the proposal. Let's think through the statuses:

- `new`. this is the default. The DUNS has never been validated by SAM.gov. (alias/display name could be `pending`--conceptually they seem the same to me)
- `accepted`. SAM.gov accepts the DUNS.
- `rejected`. SAM.gov rejects the DUNS.

`sam_checked` is great. Ideally we should run these against all DUNS numbers fairly regularly. Someone could who was once accepted could become rejected and we want to catch that.

harrisj:
Three more questions then:

1. Did you want a nil state in addition to the other states? Or should we have `new, pending, accepted, rejected`?
2. Does SAM.gov API actually ever reject a DUNS number or is this something we would apply if the DUNS has not been accepted and it's been a long enough time?
3. We should probably change `sam_checked` to `sam_last_checked` to make it more clear it's a datetime.

jessieay:
PR for this here 

https://github.com/18F/micropurchase/pull/467

[...]

PR #467, opened 2016-04-25
System shows pending message when DUNS number is being verified with …

jessieay:
…Sam.gov

* Closes https://github.com/18F/micropurchase/issues/426
* Move from `true/false` field to active record enum
* 3 states: pending, accepted, rejected

It looks like this:

screen shot 2016-04-25 at 11 34 06 am


app/models/sam_account_reckoner.rb
```diff
@@ -24,8 +26,12 @@ def should_clear?
     user.persisted? && user.duns_number_changed?
   end
 
-  def user_in_sam?
-    client.duns_is_in_sam?(duns: user.duns_number)
+  def sam_status
+    if client.duns_is_in_sam?(duns: user.duns_number)
```
harrisj: This is why I get a little bit leery of enums, especially if we are inspecting the DB at some point. In any event, can we map these to constants or something?

harrisj:
Looks good to me, but it looks like we'll need a rebase.

app/models/sam_account_reckoner.rb
jessieay: updated to use symbols instead of integers at 8cbae1d

better?

I think the only place we need to ref integers rather than symbols is in AR queries

eg

```ruby
User.where(sam_status: :sam_pending)
```

does not work. So you must do:

```ruby
User.where(sam_status: 0)
```

this will be fixed in Rails 5.  https://github.com/rails/rails/issues/19964#issuecomment-98591859

jessieay:
@harrisj rebased!

features/sam_check.feature
```diff
@@ -26,8 +28,10 @@ Feature: Automatically checking a user's SAM status
   Scenario: Pending SAM check on DUNS edit
     Given I am a user without a verified SAM account
     And I am signed in
-    When I enter a new DUNS in my profile
+    When I visit my profile page
+    And I enter a new DUNS in my profile
     And I click on the "Submit" button
+
```
harrisj: This test was failing because the SAM pending message only shows up on the profile edit page and submit redirects to the homepage. Should we consider tweaking the homepage SAM warning in another PR?

jessieay: yeah, feels weird to have the SAM warning in two places / slightly different. Add an issue?

harrisj: Will do

jessieay:
@harrisj thanks for the update! I guess merge when ready?

harrisj:
MERGING AWAY!

PR #473, opened 2016-04-26
Add docs for how to set env vars

jessieay:
* Move credentials classes into their own dir
* Move docs to "docs" dir


harrisj:
Looks good to me

PR #474, opened 2016-04-26
Use `example.com` email addresses in factories

jessieay:
* Best practice to use reserved domain name for testing
* Ref: http://www.faqs.org/rfcs/rfc2606.html
* This became clear today when we were testing emails on staging,
  which uses seed data via Factory Girl and had some weird / possibly
  real email addresses for users
* Also speeds up test suite by a couple of seconds


harrisj:
Okay, I also made some tweaks to some Cucumber steps as well. I know that it shouldn't make a difference ideally (because mocks and all that), but I wanted to just be consistent. Also fixed an error from the test runner.

jessieay:
good catches, @harrisj ! this looks ready to go to me ;)

jessieay:
considering this a pair effort w Jacob and mergin!

PR #475, opened 2016-04-27
[WIP] Simple form for datetime fields

jessieay:
Just spiked this out. I was having issues with datetimes not working as I wanted to when creating auctions. This makes the process much more explicit (no chronic to parse)

Looks like this:
screen shot 2016-04-26 at 5 59 27 pm

We could also make it a fancy jquery ui selector but didn't want to spend too much time on it before we decide that this is a direction we want to move


jessieay:
@mtorres253 feedback on this concept? I know we don't want to put too much time into the forms before we do the user research necessary..this change would take a small amount of time to finish and would clarify that the form inputs are in UTC. They also make it more obvious *how* you enter in an exact date / time since there is a drop down for month / day / year / hour / minute.

jessieay:
just spoke w @mtorres253, he wants to think more about UX before making this type of change. closing.

PR #476, opened 2016-04-27
CodeClimate suggested adding these files

harrisj:
But I think I need @stvnrlly to look these over


stvnrlly:
I would say that we should remove the `css` linter since the stylesheets seem to all be `scss`, and use https://raw.githubusercontent.com/18F/frontend/18f-pages-staging/.scss-lint.yml as the style guide.

It looks like they don't support `jshint`, so I guess we don't need the `.jshintrc`. We can get the same result using https://raw.githubusercontent.com/airbnb/javascript/master/linters/.eslintrc for `eslint` (instead of what they recommended).

Issue #453, opened 2016-04-22
Include winning bidder in public and private REST API responses


adelevie:
As a consumer of the Micropurchase API, I want to know who the winner of a given auction is and I don't want to have to munge data or have to interpret the API response to see who won. So, when an auction has ended, and I make a request to get information for that auction, the winner of that auction should be easily identifiable in the response.

Scenario: Auction data is requested after winner is identified
Given that there is an auction where a winner has been identified
When a GET request is made for that auction data
Then the response should identify the winner of that auction
And the winner should be easy to identify (front-end consumer should not have to 'figure anything out')

Responding to @gemfarmer's need to know who the winner is given the auctions.json data. It would be redundant to have to recalculate that in JS, so we should just ensure the API returns it.

mtorres253:
@adelevie @gemfarmer Can you be more specific as to why this data is being used? Do we need anything specific returned?

gemfarmer:
@mtorres253 I'll take a stab.

#### User stories:

_As a front end developer_ on this project I want to be able to know who won an auction, when they won it, and how much they won it for in order to present this information on the [previous winners dashboard](micropurchase.18f.gov/auctions/winners/).

_As someone interested in collecting data_ about the micropurchase platform, I want to be able to know who won an auction, when they won it, and how much they won it for in order to better understand the micropurchase platform.

Why this is useful for the dashboard
- *Accuracy/consistency*: the dashboard does quite a bit of data munging/transforms to determine the "winning bid" and "winning bidder" for a particular auction. Currently, given the `auctions/` api endpoint, the winner is unclear and therefore open to misinterpretation (e.g. is the winner the last bidder? Is it the lowest bidder? ). 
- *Redundancy*: it doesn't make sense to be performing these data transform in back-end AND front-end
- *Performance*: less munging = charts show up faster

adelevie:
As for points, this feels like a 2 or a 3. We already have the API setup and [the methods](https://github.com/18F/micropurchase/blob/develop/app/models/presenter/auction.rb#L103-L105) for calculating the winners. We just need to connect the two.

mtorres253:
Perfect @gemfarmer. I updated the story. And, I pointed it as a 3 based on @adelevie. Since there are no users yet for the dashboard, this is not an urgent priority. However, we should be returning the winner in the response as it is one of the more important data points.

vzvenyach:
[...]

PR #484, opened 2016-04-28
Add `winning_bid` to API 

jessieay:
* properties: `bidder_id` and `amount`
* this is in both the output for a single bid and the output for all bids
* also refactored tests to clarify dependencies
* closes https://github.com/18F/micropurchase/issues/453

If there is no winning bid available, this is what the output looks like:

```
{"auction"=>
  {"bids"=>[],
   "created_at"=>"2016-04-28T00:27:04+00:00",
   "description"=>
    "Audacia patria aestas peior accusamus deprimo aiunt. Spoliatio tremo est quia sint bene accipio animi. Admoveo omnis stabilis similique utrimque. Spoliatio anser victoria vel suggero vox condico earum.\n\nCura vinum quam amiculum. Sollicito vado vomer caute texo auxilium vinco. Crustulum volo victoria uxor armarium defleo.\n\nCaelum cetera studio audio. Aer certe certus demoror. Alii expedita tolero culpa cursim ut.",
   "end_datetime"=>"2016-04-30T00:27:04+00:00",
   "github_repo"=>"https://github.com/18F/calc",
   "id"=>176,
   "issue_url"=>"https://github.com/18F/calc/issues/255",
   "start_datetime"=>"2016-04-26T00:27:04+00:00",
   "start_price"=>3500,
   "summary"=>"Ut ut qui. Natus nesciunt quaerat. Exercitationem omnis cum iusto aliquam ratione deleniti veniam.",
   "title"=>"Future-proofed fresh-thinking array",
   "updated_at"=>"2016-04-28T00:27:04+00:00",
   "winning_bid"=>{"amount"=>nil, "bidder_id"=>nil}}}
```


jessieay:
cc @gemfarmer -- you cool with returning `nil` when there is no winning bid available? I figure that is more friendly than returning no `winning_bid` json object at all...

gemfarmer:
@jessieay yes, I would prefer an empty object than a non-existent one. :)

app/models/presenter/auction.rb
```diff
@@ -32,6 +32,7 @@ def initialize(auction)
       :to_param,
       :type,
       :updated_at,
+      :lowest_bid,
```
adelevie: just trying to wrap my head around this ... where is `lowest_bid` used in this PR where it wasn't used before?

spec/requests/admin/users_spec.rb
```diff
@@ -1,6 +1,8 @@
 require 'rails_helper'
 
-RSpec.describe Admin::UsersController do
+describe Admin::UsersController do
+  include RequestHelpers
```
harrisj: Why does this need to be included and there are no other changes in this file?

app/models/presenter/auction.rb
jessieay: Lowest bid is used to calculate winning bid. If we were using a plain Ruby object in the RulesFactory (like we do in its first two uses) we don't have to mess with the view model but since the API uses the controller and that returns a view object for auction, we need to delegate it down. We def need to reconsider how heavily we are relying on the auction view model /presenter because right now they are all tangled up!

gemfarmer:
@jessieay looks great from my end!

app/serializers/winning_bid_serializer.rb
```diff
@@ -0,0 +1,16 @@
...
+    :amount,
+    :bidder_id
+  )
+
+  def amount
+    object.amount
+  end
+
+  def bidder_id
+    object.bidder_id
```
cwarden: The current bidder is exposed during active auctions.  Looks like you need to use `veiled_bidder`.

adelevie: Thanks, Chrisrian. We just deployed a bug fix in our latest PR.

Issue #388, opened 2016-04-12
Admin can navigate to edit page for admin users


mtorres253:
As an admin user, I want to be able to edit users via the UI instead of via a Yaml file.

Scenario: Admin navigates to an edit page for other admin users
Given I am an authenticated admin user
And I have permission to edit existing admin users
When I fill am viewing https://micropurchase-staging.18f.gov/admin/users
And I click on a user name
Then I should be taken to a user edit page
And on that page, I should see the user name and email address
And I should see a checkbox (unchecked by default) labeled "Contracting Officer"
And I should see a "Save" button

PR #489, opened 2016-04-28
Admin can set user as "contracting officer"

jessieay:
* Closes https://github.com/18F/micropurchase/issues/388

form: 
screen shot 2016-04-28 at 2 45 45 pm

after submit:

screen shot 2016-04-28 at 2 46 10 pm


Gemfile
```diff
@@ -20,6 +20,7 @@ gem 'active_model_serializers'
 gem 'business_time'
 gem 'holidays'
 gem 'factory_girl_rails'
+gem 'simple_form'
```
harrisj: We don't have an issues with simple_form in our projects, do we?

It's not the easiest to tweak the style of (as I found on when trying to make it more compatible with US Web Design Standards), but it does alleviate some of the tedium of form creation so that is good.

harrisj:
LGTM. Merging

PR #490, opened 2016-04-28
Fix mailer test

jessieay:
* SMTPCredentials returns parsed JSON or ENV var depending on ENV
* Failing on CI bc weirdly, parsed JSON returns a string for `from`
  whereas env var string returns a one item array
* This makes no sense but whatevs


jessieay:
@harrisj ready for u

PR #506, opened 2016-05-02
Move / rename presenters and view models

jessieay:
* Put in own directories
* Append type to name rather than namespacing
* This is more consistent w Rails conventions, eg: `app/helpers/application_helper.rb`
* This is pre work for some refactoring I'd like to do to the view
  models to make them less dependent on presenter behavior


app/controllers/admin/auctions_controller.rb
```diff
@@ -3,7 +3,7 @@ class AuctionsController < ApplicationController
     before_filter :require_admin
 
     def index
-      @auctions = Auction.all.map { |auction| Presenter::AdminAuction.new(auction) }
+      @auctions = Auction.all.map { |auction| AdminAuctionPresenter.new(auction) }
```
harrisj: So, we want to switch to a different naming style? I don't have a preference (and this might avoid some issues with Rails' autoload), but I want to see what we all think

jessieay: This naming style is closer to what I usually see and closer to how rails does naming, eg  `app/helpers/application_helper.rb` using name `ApplicationHelper` as opposed to `app/helpers/application.rb` being named `Helper::Application` 

Existing naming wasn't bad but I wanted to bring the view models and presenters up to the same level as `models` in the dir structure rather than hiding them in there.

spec/view_models/auction_status/expiring_view_model_spec.rb
```diff
@@ -0,0 +1,25 @@
+require 'rails_helper'
+
+describe AuctionStatus::ExpiringViewModel do
```
harrisj: I'm not sure I completely understand when some things are in modules/classes like BidPresenter::Null and this and others are not

app/controllers/admin/auctions_controller.rb
harrisj: No I think that is an admirable goal and I have no issue with that, but I just want to make sure @adelevie is aware of this

spec/view_models/auction_status/expiring_view_model_spec.rb
jessieay: I put all of these `AuctionStatus` view models in that module as a way of namespacing them. I do think the null bid presenter should be named `NullBidPresenter`

jessieay: When we are dealing with objects that are *not* similar in behavior, but just happen to be the same type of thing, I think namespacing is less helpful. 

So status-type view models have namespacing of `AuctionStatus` but all view models are already logically organized by living in the `view_models` dir...am I making sense?

jessieay:
I think I've made code climate as happy as I can :)

spec/view_models/auction_status/expiring_view_model_spec.rb
jessieay: updated `NullBidPresenter` at 62c9230

harrisj:
I was thinking we should relax the class length issue in Rubocop for now. We can also tell CodeClimate to just ignore these specific issues

harrisj:
I am okay merging this if @adelevie is. I just would prefer we do this sooner rather than later, since big reorg PRs start to block progress if they languish

jessieay:
@adelevie and I talked about this today, he says he is cool w/ it! 😄

harrisj:
Go ahead and merge then

harrisj:
Alright, I will do it.

PR #509, opened 2016-05-02
Fixing the relative path issue for closed single-bid auctions

harrisj:
This is a duplicate of #508, but that branch was deleted and I'm having issues just pushing it again so I created a duplicate of it for merging into develop


jessieay:
let's merge this now but also add tests for it!

Issue #480, opened 2016-04-27
Enhancement: Remove some winners dashboard logic


gemfarmer:
Refactor or remove client-side logic in `previousWinners.js` to determine key metrics in favor of [18f/micropurchasedata](https://github.com/18F/micropurchasedata)

PR #515, opened 2016-05-03
Winners refactor

gemfarmer:
For #480 

The objective of this PR is to pull apart components of the previousWinners.js script into intelligible, more reusable and testable components.

What the PR does:
- [x] added charts.js, metrics.js
- [x] refactored logic to include addition of `winning_bid`
- [x] removed dead code in previousWinners.js
- [x] update tests, added new tests 
- [x] adds small amount of documentation to [CONTRIBUTING](CONTRIBUTING.md)

Next steps (open issue when PR closes): 
- charts.js and metrics.js should be moved to npm and managed as packages
- manage d3, c3, and underscore with npm instead of static files


gemfarmer:
@jessieay this passes, but the refactor now relies on the `winning_bid` attribute, so it either
- shouldn't be merged until that attribute is re-introduced
- pros: the page will continue to look good 
- cons: introduces new blocker

OR

- it can be merged and will just look bad until the winning_bid attribute is back in action. 
- pros: we don't have to worry about rebasing excessively
- cons: page will look bad if people happen to see it

I vote for the ladder because the page hasn't been shared, but if you think that getting `winning_bid` back up and running might take a bit, my vote is for the former

gemfarmer:
@jessieay if you think it is ready to merge I will squash.

gemfarmer:
Could someone review? @adelevie @harrisj @jessieay

db/schema.rb
```diff
@@ -72,8 +72,8 @@
     t.datetime "created_at",                           null: false
     t.datetime "updated_at",                           null: false
     t.string   "email"
-    t.string   "github_login"
     t.string   "credit_card_form_url"
+    t.string   "github_login"
```
harrisj: This is kinda weird, thanks Rails schema generation

harrisj:
Thanks! This looks good to me, but I'll let you have the honor of merging

db/schema.rb
gemfarmer: yeah that keeps happening to me. I try to stash instead of commit those schema changes, but sometimes I miss them in large commits :/

PR #519, opened 2016-05-06
Ad hotfix veil winning bidder from api

jessieay:
PRing this to develop so we can get it back in sync w master


harrisj:
Looks good to me. Merging

PR #520, opened 2016-05-06
Only show winning bid in the API if auction is closed

jessieay:
* Modification / revert-ish of https://github.com/18F/micropurchase/pull/519

### DIFF
--- a/app/models/null_bid.rb
+++ b/app/models/null_bid.rb
@@ -1,7 +1,9 @@
 class NullBid
   def amount
+    nil
   end
 
   def bidder_id
+    nil
   end
 end
--- a/app/models/null_bid.rb
+++ b/app/models/null_bid.rb
@@ -1,0 +1,7 @@
+class NullBid
+  def amount
+  end
+
+  def bidder_id
+  end
+end
--- a/app/models/rules/base_rules.rb
+++ b/app/models/rules/base_rules.rb
@@ -1,0 +1,13 @@
+class Rules::BaseRules < Struct.new(:auction)
+  def partial_path(name, base_dir = 'auctions')
+    "#{base_dir}/#{partial_prefix}/#{name}.html.erb"
+  end
+
+  def user_can_bid?(user)
+    auction.available? && user.present? && user.sam_accepted?
+  end
+
+  def partial_prefix
+    fail NotImplementedError
+  end
+end
--- a/app/models/rules/basic.rb
+++ b/app/models/rules/basic.rb
@@ -1,16 +1,10 @@
-class Rules::Basic < Struct.new(:auction)
+class Rules::Basic < Rules::BaseRules
   def winning_bid
     auction.lowest_bid
   end
 
   def veiled_bids(user)
     auction.bids
-  end
-
-  def user_can_bid?(user)
-    return false unless auction.available?
-    return false if user.nil? || !user.sam_accepted?
-    true
   end
 
   def max_allowed_bid
@@ -29,24 +23,12 @@ class Rules::Basic < Struct.new(:auction)
     true
   end
 
-  def partial_path(name, base_dir = 'auctions')
-    if partial_prefix.blank?
-      "#{base_dir}/#{name}.html.erb"
-    else
-      "#{base_dir}/#{partial_prefix}/#{name}.html.erb"
-    end
-  end
-
   def partial_prefix
     'multi_bid'
   end
 
   def formatted_type
     'multi-bid'
-  end
-
-  def rules_type
-    'basic'
   end
 
   def highlighted_bid_label
--- a/app/models/rules/sealed_bid.rb
+++ b/app/models/rules/sealed_bid.rb
@@ -1,4 +1,4 @@
-class Rules::SealedBid < Rules::Basic
+class Rules::SealedBid < Rules::BaseRules
   def winning_bid
     if auction.available?
       NullBidPresenter.new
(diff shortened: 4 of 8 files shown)


harrisj:
I'm going to ask you to make the changes that CodeClimate wants. Sorry.

app/models/null_bid.rb
```diff
@@ -0,0 +1,7 @@
+class NullBid
+  def amount
```
harrisj: I know these return nil by default, but should we add a comment that this is happening here?

app/models/rules/base_rules.rb
```diff
@@ -0,0 +1,13 @@
+class Rules::BaseRules < Struct.new(:auction)
```
harrisj: Should we remove the module parent?

app/models/rules/basic.rb
```diff
@@ -1,4 +1,4 @@
-class Rules::Basic < Struct.new(:auction)
+class Rules::Basic < Rules::BaseRules
```
harrisj: Maybe we should call the default auction something else since we have Base Rules now. Default? Reverse?

app/models/rules/sealed_bid.rb
```diff
@@ -1,4 +1,4 @@
-class Rules::SealedBid < Rules::Basic
+class Rules::SealedBid < Rules::BaseRules
   def winning_bid
```
harrisj: I approve of this change to the OOP hierarchy

app/models/rules/basic.rb
jessieay: ah yes this was a tricky naming situation -- "Reverse" is not as clear to me. Perhaps "Default' although that feels similar to "Base", too. Now that I know we used to call them "Multi" I always call it that in my head....ack idk! what do you think?

app/models/rules/base_rules.rb
jessieay: you mean just call it `BaseRules` ?

app/serializers/auction_serializer.rb
```diff
@@ -43,6 +43,15 @@ def winning_bid
   private
 
   def find_winning_bid
-    RulesFactory.new(object).create.winning_bid
+    if auction_status.available?
+      NullBid.new
+    else
+      RulesFactory.new(object).create.winning_bid
```
harrisj: I know I approved that earlier, but I am curious about when we should have a dedicated factory class like this vs. a factory method in a class

jessieay: benefit of factory class is that it brings all of the logic into one confined thing rather than adding to the "API" of an existing object, like `Auction` or `AuctionPresenter`...but yeah. Still need to think through how we're dealing with all of these view / presenter / model layers

app/models/rules/sealed_bid.rb
jessieay: been reading the Sandi Metz book and she had an example almost exactly like this so I followed her advice :)

app/models/rules/base_rules.rb
harrisj: Yeah, what do you think?

app/models/rules/basic.rb
harrisj: Yeah maybe @adelevie has a good idea. I just feel like Base and Basic are easily confused.

app/serializers/auction_serializer.rb
harrisj: Oh I'm not asking you to justify all your actions, sorry! I just grapple with this same question and was wondering when the threshold is reached.

app/models/null_bid.rb
adelevie: I'd be 👍 for explicit `nil` here. Otherwise it might make a future dev think we just forgot to fill in these methods. Explicit `nil` says, no, this is what the method should be.

app/models/rules/basic.rb
adelevie: I always felt like `multi-bid` and `single-bid` clarified this stuff pretty well in my mind :/

But I guess both of those are sub-types of Reverse. And spit-balling a bit more here, other plausible auction types:

- `Dutch` (probably not a child of `Reverse`) https://en.wikipedia.org/wiki/Dutch_auction
- `BuyItNow` or `FixedPrice` (should be composable with with any Reverse auction. e.g. something can be BuyItNow and MultiBid.
- `Lot` or `MultiWinner`: There can be many winners, maybe used for a crowdsourcing micro-purchase. Should likely inherit from `ButItNow` (assuming these would just have a fixed price).

gemfarmer:
I have already updated it in #515 for this PR, but in the future could y'all update the mock json that I am using the jasmine tests **[here](https://github.com/18F/micropurchase/pull/515/files#diff-898a2d13db01679947e5ac750b37da71R1)** when you make api adjustments?

app/models/rules/basic.rb
harrisj: Yes, I guess we should call it Reverse or Descending although neither option is particularly exciting

app/models/null_bid.rb
harrisj: Yeah let's force it to return a nil and I guess disable Rubocop from being upset by wrapping with a disable

gemfarmer:
@jessieay what is the status here?

jessieay:
I am in a ruby training today through Weds, @gemfarmer -- @harrisj said he would take over this PR

harrisj:
I will take over this PR. I just need a good name for the public descending auctions so I can make that change. Suggestions?

app/models/rules/basic.rb
harrisj: Going to defer this to a later PR, where we'll also move some files around in the view directories

app/models/rules/base_rules.rb
harrisj: Going to defer this for a later PR

PR #521, opened 2016-05-06
Updating the about.yml

harrisj:
Noticed a few team-members weren't in there so I wanted to change it, but is there anything else we should be tweaking in here?

### DIFF
--- a/.about.yml
+++ b/.about.yml
@@ -49,7 +49,7 @@ team:
 
 - github: adelevieb
   id: Alan DeLevie
-  role: Lead Developer
+  role: Developer
 
 - github: harrisj
   id: Jacob Harris
@@ -61,7 +61,7 @@ team:
 
 - github: gemfarmer
   id: Brian Hedberg
-  role: Lead Frontend Designer
+  role: Front End Designer
 
 - github: andrewmaier
   id: Andrew Maier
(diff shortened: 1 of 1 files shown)


adelevie:
I don't love being labeled as "lead developer" (like ... I'm not), so we can change that just to "developer".

gemfarmer:
@harrisj similarly I can just be `Front End Designer`.
Also, should Alla and Duane be removed if they have helped on the project in the past? Is the point here to generate a list of current devs or project participants?

harrisj:
I will make those edits, but in some cases it's just what was in there before, @adelevie

I will admit I am not entirely sure if this is meant to represent the current team of the project or everybody who has worked on the project at some time, but the about_yml cheatsheet says `Who are the team members on your project? `

It would be nice if we had an `alumni` section

gemfarmer:
should @mtorres253 's role be `lead` instead of [`Project Manager`](https://github.com/18F/micropurchase/pull/521/files#diff-5b2628df36d3e851221d184cc28347b4R44)?

harrisj:
Noticed that @gemfarmer has asked the about_yml project about how to indicate past team members https://github.com/18F/about_yml/issues/88

gemfarmer:
yes, thanks @harrisj for cross-referencing

gemfarmer:
@harrisj just noticed [this](https://github.com/18F/micropurchase/blob/amend-about-yml/.about.yml#L141). Is @allalala still the point of contact? or is it @mtorres253 now?

mtorres253:
Just made myself point of contact. Everything else looks good to me.

PR #523, opened 2016-05-08
Sometimes, things have to get worse before they get better

jessieay:
* Attempt at refactoring presenters and view objs
* Strategy, per Kane suggestion: dump everything into one class
* Let objects emerge from that class rather than trying to disentangle
  existing objects related to auctions/bids
* I realize that this commit makes everything worse. But this is the
  step in making things better!
* This is what I did on my flight from DCA to SFO on Friday. 💇 🍔 💯


harrisj:
Much as I appreciate the sentiment I think we might have to wait on this one until we have a week of refactoring clearance for the project. i suppose we could just keep tweaking on a separate branch but the we are coding against master, develop and this branch and that's hard. What are the best agile approaches for large refactors?

mtorres253:
Let's talk about this in grooming. Want to not ignore the refactoring needs
as we progress.

jessieay:
>  What are the best agile approaches for large refactors?

The tests are all passing here, so I would consider the functionality to be the same as it was pre-refactor, just with different code making it run under the hood. Deleted tests were modified / moved for new objects where applicable.

If we are not confident that the tests passing means the app functionality is still in tact, perhaps we can talk about places where our test coverage is weak so we can address that?

I tried to make this commit as small as possible. Certainly it is just one step along a path of refactors, which is why I opened it now rather than waiting for the next level of extracting objects from `Auction` and `Bid`.

Would love technical feedback from @baccigalupi as well since she suggested this approach when I mentioned wanting to refactor to View model / presenter layer but feeling overwhelmed with refactoring those abstractions.

jessieay:
@harrisj to actually answer your question about "best practices" here are my general rules for refactor PRs:

- under 500 changes (breaking that rule here, obv, but couldnt' find a way around it. Luckily, most changes are deletions so most of the code doesn't actually need to be reviewed...since it's gone)
- green test suite

I do think it makes more sense to merge this rather than maintain it is a separate branch for some indefinite amount of time. 

If we want to merge it after the existing open PRs are merged in order to prevent crazy merge conflicts, I am totally on board with that.

.rubocop.yml
```diff
@@ -23,7 +24,7 @@ Metrics/ClassLength:
   Description: Avoid classes longer than 100 lines of code.
   Enabled: true
   CountComments: false
-  Max: 100
+  Max: 200
```
baccigalupi: I wouldn't change this, just exclude this work temporarily from the rule. Also the Description is now out of date

baccigalupi:
I am with Jessie about keeping refactors small instead of scheduling a big block of time for the change-the-world bang. Also, if tests are passing, it should be good to go ... or there is a problem with test confidence that needs to be understood and fixed ... like she said.

When I was talking about collapsing all the complexity into one object (or one object per view), I was thinking of a mega view model, not collapsing things into the AR models. I am a believer that life in Rails is best when there is nothing but relationships in the AR classes.

It seems like I didn't really understand the extent to which view and domain logic was mixed up in these objects.

There might be a better first step to taking apart the increasing PORO complexity, like figuring out which things are _currently_ needed outside the view.

Sorry if my remote advice keeps making this application harder to work on :(

harrisj:
I apologize if my remarks seemed out of place or if I was impugning the motivations/quality of the work. I think these are admirable goals and I like breaking it up into smaller commits with green tests passing like you have done. But I think there is an implicit assumption that this commit would be followed very soon after by some other commits to make things better and I just want to make sure we'd have a little room to do that.

harrisj:
And yes, let's rebase and merge after the other ones.

I guess my main question is, should we have some sort of epic or label to organize the pull requests and issues around? And how much should we try to name some future technical debt cleanup in advance? (ie, renaming multi_bid/single_bid template files to something else more consistent)

jessieay:
> When I was talking about collapsing all the complexity into one object (or one object per view), I was thinking of a mega view model, not collapsing things into the AR models. I am a believer that life in Rails is best when there is nothing but relationships in the AR classes.

The issue I was seeing with the way our current code is set up is that we are already highly dependent on the `Auction` AR class, it's just less obvious because we are dependent via delegation.

This is what we current do: decorate an `Auction` instance with `AuctionPresenter`, pass `AuctionPresenter, along with the `current_user` to `AuctionViewModel`, then pass that `AuctionViewModel` instance to the view.

Even though it *looks* like we are using `AuctionViewModel`, we're actually using methods from `AuctionViewModel` + `AuctionPresenter` + `Auction` in the view. Since we're delegating down to `Auction`, it felt like dumping stuff there, at least temporarily, was the Most Obvious Thing. If we try to bring everything up to `AuctionViewModel` as a short term solution, I am not 100% clear on why that is better (neither choice feels great, so why is that less bad?)

> There might be a better first step to taking apart the increasing PORO complexity, like figuring out which things are currently needed outside the view.

Agree, this approach was a kind of "burn down the house" approach. I was feeling frustrated by my inability to reason about our current hierarchy so I went crazy deleted a bunch of stuff the code functionality more obvious. It might be more responsible / easy to review if I do it in smaller chunks.

>  should we have some sort of epic or label to organize the pull requests and issues around? And how much should we try to name some future technical debt cleanup in advance?

I first noticed the issue with this code when refactoring how we calculate `winning_bid`

harrisj:
I am the person who is responsible for that ugly `ViewModel::Auction` class. I created it because I wanted to solve a few issues:

1. There were some methods that were only used in views (like `show_bid_button?`) but not confined to a single view.
2. I wanted to bind methods to the current_user (available in the view models) so I could eliminate the need to pass in current_user in ERB files

That said, it created another level of delegation and a bit of duplication of method names (one with a user argument in the Presenter and one without). I would like a better option too. I hope we can discover it together.

jessieay:
In grooming today we decided to do more incremental refactors. Closing this for now and will open some smaller Prs based on the work here

Issue #501, opened 2016-04-29
Vendor sees auction start/end times and delivery time in ET when viewing new auction details


mtorres253:
As a vendor, before an auction goes live, I want to be able to see auction details containing the start and end date of the bidding as well as the delivery date, so that if I am interested in an auction, I can plan to bid on it.

Scenario: Vendor views auction details before the auction is live
Given I am an authenticated vendor
And there is an auction where bidding has not yet started
When I view the details page of an auction (https://micropurchase.18f.gov/auctions/[number]
Then I should see when bidding starts and ends in ET
And I should see the delivery deadline in ET

harrisj:
Should I make these separate features (do you want a whole story for them) or should I add these as assertions to our existing cucumber stories for viewing the auction details?

harrisj:
For reference, here are some stories that look at auctions already:

* https://github.com/18F/micropurchase/blob/develop/features/auctions.feature
* https://github.com/18F/micropurchase/blob/develop/features/bidding.feature

We could consider if these should be constructed better and if I should add some steps to them. Or I could add new features too.

harrisj:
Also, do we have a mockup for showing the delivery deadline on the page? Or am I just to do that?

harrisj:
Closing since this was merged

Issue #512, opened 2016-05-03
Vendor sees start/end times and delivery time for auction where bidding has ended, but has not yet been delivered


mtorres253:
As a vendor who is registered on the micropurchase platform, even though no more bids are being accepted, I still want to see the start time and end time of the auction, as well as the delivery date, so that I can get an idea of how auctions are generally set up, or, if I'm the winner, so that I can see when I need to deliver the work.

Scenario: Vendor looks at auction details page after the auction bidding ends
Given I am an authenticated vendor
And there is an auction where bidding has ended
When I go to the details page for that auction
Then I should see the bidding start and end date as well as the delivery deadline, in ET

harrisj:
This was referenced in PR #524 which has been merged

Issue #511, opened 2016-05-03
Vendor sees auctions start/end times and deliver times in ET for current auctions


mtorres253:
As a vendor who is interested in bidding on an auction, after the auction starts, I want to see when it started, when it will end and the delivery date, so that I can plan my bid, so the system should show me these dates clearly.

Scenario: Vendor views details page for an auction during its bid period
Given I am an authenticated vendor
And there is an auction where bidding has started and has NOT ended
When I view the details page of an auction (https://micropurchase.18f.gov/auctions/[number]
Then I should see in ET when bidding started and will end
And I should see the delivery deadline in ET

harrisj:
This was referenced in PR #524 which has been merged

PR #524, opened 2016-05-09
Display times as EDT on auction detail page

harrisj:
This is an implementation of #501 with a little bit of additional work:

* I did not implement additional scenarios for this issue, but just changed the implementation of existing Cucumber features.
* Renames the `starts_in` and `starts_at` methods to be clearer
* Cleans up some duplication of code for relative time display

Note there is still some duplication between `relative_end_time` and `relative_time_left` as well as the need for some tests for these time display methods. I just wanted to put this here now so @mtorres253 could see it.


adelevie:
so far so good!

harrisj:
I think this closes three issues:

* Closes #501 
* Closes #511 
* Closes #512 

But I need to check. I did not actually implement the cucumber features from there, but fixed the implementations of existing scenarios/stories. Is this acceptable, @mtorres253?

features/step_definitions/auction_steps.rb
```diff
@@ -132,14 +145,23 @@
   expect(page).to have_content("You are currently not the winning bidder.")
 end
 
-# Fix me (look for specific date in there)
 Then(/^I should see when the auction started$/) do
-  expect(page).to_not have_content("Auction ended at:")
+  expect(page).to have_text(
+                    DcTimePresenter.convert_and_format(@auction.start_datetime))
```
gemfarmer: it looks like there is some extra whitespace here.

spec/presenters/dc_time_presenter_spec.rb
```diff
@@ -32,7 +32,15 @@
 
     context 'when it is a time' do
       it 'should convert to a reasonable string' do
-        expect(converter.convert_and_format).to eq("April 01, 2001 06:15:00 AM")
+        expect(converter.convert_and_format).to eq("April 01, 2001 06:15:00 AM EDT")
```
gemfarmer: Should this be `EST` instead of `EDT`?

spec/presenters/dc_time_presenter_spec.rb
```diff
@@ -32,7 +32,15 @@
...
+      end
+
+      context 'when we are in daylight saving time' do
+        let(:time) { Chronic.parse('12/1/2010 6:15am').utc }
+        
+        it 'should handle daylight saving time correctly' do
+          expect(converter.convert_and_format).to eq("December 01, 2010 06:15:00 AM EST")
```
gemfarmer: Should this be `EDT` instead of `EST`?

features/step_definitions/auction_steps.rb
harrisj: Spacing things to make Rubocop happy

spec/presenters/dc_time_presenter_spec.rb
harrisj: Nope, it's EDT for spring/summer/fall and EST for winter

app/presenters/auction_presenter.rb
```diff
@@ -89,20 +89,24 @@ def bid_count
...
+  end
+
+  def relative_start_time
     time_in_human(model.start_datetime)
   end
 
-  def ends_in
-    time_in_human(model.end_datetime)
+  def relative_time_left
+    "#{distance_of_time_in_words(Time.now, model.end_datetime)} left"
```
jessieay: thoughts on `Time.current` (alias for `Time.zone.now`) instead of `Time.now`? The former takes into account the timezone. http://winstonyw.com/2014/03/03/time_now_vs_time_zone_now/

app/views/auctions/show.html.erb
```diff
@@ -39,6 +39,12 @@
         <p class="auction-label"><%= @view_model.auction_start_label %></p>
         <p class="auction-label-info"><%= @view_model.auction_formatted_start_time %></p>
       </div>
+      <% if @view_model.auction_has_delivery_deadline? %>
```
jessieay: can we write this method in a way that prevents the need for a conditional in the view?

features/step_definitions/auction_steps.rb
jessieay: woof that is annoying -- which rule is enforcing this spacing?

app/presenters/auction_presenter.rb
```diff
@@ -89,20 +89,24 @@ def bid_count
     bids.size
   end
 
-  def starts_at
+  def formatted_start_time
     DcTimePresenter.convert_and_format(model.start_datetime)
```
jessieay: thoughts on putting `DcTimePresenter.convert_and_format` into a private method so we can re-use it for these three methods rather than calling it each time?

jessieay:
yay very excited for this!

app/presenters/auction_presenter.rb
harrisj: We could, but I think I might consider such things for future PRs

app/presenters/auction_presenter.rb
harrisj: I didn't know about that. It looks like it doesn't affect the calculations, so sure...

app/views/auctions/show.html.erb
harrisj: Well, the method would return `&nbsp;` if the delivery_deadline is not set, but the label would still be there.

features/step_definitions/auction_steps.rb
harrisj: Line length

harrisj: Okay, my bad...

spec/presenters/dc_time_presenter_spec.rb
gemfarmer: 👍

jessieay:
look great! mergin

Issue #499, opened 2016-04-29
Vendors are marked as small businesses


mtorres253:
As an admin user, I want to know whether a vendor is a small business because some auctions will only be available to small businesses, so I should see some indication of that.

Scenario: User is in Sam.gov and is a small business
Given that I am an admin user
And there are users who are in Sam.gov that are small businesses
When I view the users page (https://micropurchase.18f.gov/admin/users)
Then I should see a column labeled "Small Business"
And in that column I should see a value of "Yes" for that user

Scenario: User is in Sam.gov and is not a small business
Given that I am an admin user
And there are users who are in Sam.gov that are not small businesses
When I view the users page (https://micropurchase.18f.gov/admin/users)
Then I should see a column labeled "Small Business"
And in that column I should see a value of "No" for that user

Scenario: User is not in Sam.gov 
Given that I am an admin user
And there are users who are not in Sam.gov
When I view the users page (https://micropurchase.18f.gov/admin/users)
Then I should see a column labeled "Small Business"
And in that column I should see a value of "N/A" for that user

Note: This will require some kind of check in Sam.gov for members who have a DUNS number and who are in Sam.gov. We talked about doing this every day?

Also this: 18F/samwise#23 - bake in a status check so that we get only small businesses who are active.

harrisj:
Shouldn't this be closed?

jessieay:
this looks like it is finished. closing

PR #530, opened 2016-05-11
Add a small_business flag to users table

harrisj:
**This is not ready to be merged**. This is an implementation of #499 only and adds a column to the table and displays in the admin. I still need to fix the cucumber feature to be complete. Note that this does not cover setting the `small_business` flag from samwise, just displaying in the admin.

### DIFF
--- a/app/presenters/user_presenter.rb
+++ b/app/presenters/user_presenter.rb
@@ -23,6 +23,14 @@ class UserPresenter < SimpleDelegator
     "components/user_nav_drawer"
   end
 
+  def small_business_label
+    if model.sam_accepted?
+      small_business? ? 'Yes' : 'No'
+    else
+      'N/A'
+    end
+  end
+
   def model
     __getobj__
   end
--- a/db/migrate/20160510180328_add_small_biz_flag_to_users.rb
+++ b/db/migrate/20160510180328_add_small_biz_flag_to_users.rb
@@ -1,0 +1,5 @@
+class AddSmallBizFlagToUsers < ActiveRecord::Migration
+  def change
+    add_column :users, :small_business, :boolean, default: false, null: false
+  end
+end
--- a/features/step_definitions/admin_users_steps.rb
+++ b/features/step_definitions/admin_users_steps.rb
@@ -44,3 +44,24 @@ end
 Then(/^I expect the page to show me the number of admin users$/) do
   expect(page).to have_text("Admins (1)")
 end
+
+Then(/^I should see a column labeled "([^"]+)"$/) do |text|
+  find('th', text: text)
+end
+
+Given(/^there are various types of users in the system$/) do
+  @users = []
+
+  30.times do
+    has_duns = rand(4)
+    in_sam = has_duns && rand(4)
+    small_biz = in_sam && rand(4)
+
+    opts = { }
+    opts[:duns_number] = nil unless has_duns
+    opts[:sam_status] = :sam_accepted if in_sam
+    opts[:small_business] = true if small_biz
+
+    @users << FactoryGirl.create(:user, opts)
+  end
+end
--- a/features/step_definitions/admin_users_steps.rb
+++ b/features/step_definitions/admin_users_steps.rb
@@ -23,6 +23,12 @@ When(/^I click the edit user link next to the first non-admin user$/) do
   end
 end
 
+When(/^I click the edit user link next to the first admin user$/) do
+  within(:xpath, '/html/body/div/div/table[1]') do
+    click_on "Edit"
+  end
+end
+
 When(/^I check the 'Contracting officer' checkbox$/) do
   check('user_contracting_officer')
 end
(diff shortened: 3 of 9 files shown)


db/migrate/20160510180328_add_small_biz_flag_to_users.rb
```diff
@@ -0,0 +1,5 @@
+class AddSmallBizFlagToUsers < ActiveRecord::Migration
+  def change
+    add_column :users, :small_business, :boolean, default: false, null: false
```
jessieay: let's index on this too, since we're going to be querying on it. See section on "partial indexes" here http://rny.io/rails/postgresql/2013/08/20/postgresql-indexing-in-rails.html

features/step_definitions/admin_users_steps.rb
```diff
@@ -44,3 +44,24 @@
...
 end
+
+Then(/^I should see a column labeled "([^"]+)"$/) do |text|
+  find('th', text: text)
+end
+
+Given(/^there are various types of users in the system$/) do
+  @users = []
+
+  30.times do
```
jessieay: I think we can get the same spec result but much faster with just 2 user records

db/migrate/20160510180328_add_small_biz_flag_to_users.rb
harrisj: Well, I don't know how much we will be querying on it, but it will be used to identify if the current_user can bid on it.

features/step_definitions/admin_users_steps.rb
harrisj: I went a little overboard I will admit...

db/migrate/20160510180328_add_small_biz_flag_to_users.rb
jessieay: makes sense!

features/support/env.rb
```diff
@@ -4,6 +4,8 @@
 require 'cucumber/rspec/doubles'
 require 'capybara/poltergeist'
 require 'webmock/cucumber'
+require 'capybara-screenshot/cucumber'
```
jessieay: I thought we turned these off? I was accidentally committing them all the time, would be cool if we could have them turned on in specific envs instead of globally

features/step_definitions/admin_users_steps.rb
```diff
@@ -44,3 +50,32 @@
...
+Given(/^there is a user who is not in SAM.gov$/) do
+  @user = FactoryGirl.create(:user, sam_status: :sam_pending)
+end
+
+Then(/^I should see "([^"]+)" for the user in the "([^"]+)" column$/) do |value, column|
+  user_admin_columns = ['Name', 'Email', 'DUNS Number', 'SAM.gov status',
+                        'Small Business?', 'Github ID']
+
+  index = user_admin_columns.index(column)
+  fail 'Unrecognized column: #{column}' if index.nil?
```
jessieay: neat!

features/step_definitions/admin_users_steps.rb
```diff
@@ -44,3 +50,32 @@
...
+
+Given(/^there is a user in SAM.gov who is a small business$/) do
+  @user = FactoryGirl.create(:user, sam_status: :sam_accepted, small_business: true)
+end
+
+Given(/^there is a user who is not in SAM.gov$/) do
+  @user = FactoryGirl.create(:user, sam_status: :sam_pending)
+end
+
+Then(/^I should see "([^"]+)" for the user in the "([^"]+)" column$/) do |value, column|
```
jessieay: should we use this snazzy step for some of our existing feature specs? Looks like it was written for a general purpose but is just being used for "small business"

app/presenters/user_presenter.rb
```diff
@@ -11,6 +11,14 @@ def in_sam?
     model.sam_status
   end
 
+  def small_business_label
```
jessieay: maybe this should go in `admin_user_presenter` ?

features/support/env.rb
harrisj: For some reason they were being saved to the root directory. I explicitly have them saved to `tmp` in the app now and I think that's in the .gitignore

jessieay: 👍

jessieay:
just a couple small comments from me. looking great! 🐯

app/presenters/user_presenter.rb
harrisj: Ha, we don't have an AdminUserPresenter yet...

features/step_definitions/admin_users_steps.rb
harrisj: That's the idea... Note that it is currently tied to a specific table ID if you wanted to parameterize it even further...

adelevie:
rebased

features/step_definitions/admin_users_steps.rb
jessieay: do you think we should do that work in this PR, or wait until the other tests are being updated?

PR #532, opened 2016-05-12
Don't Merge: Eligibility strawperson

adelevie:


app/models/rules_factory.rb
```diff
@@ -4,10 +4,18 @@ def initialize(auction)
   end
 
   def create
+    eligibility = if auction.small_business?
```
jessieay: might feel like overkill here, but it seems like it shouldn't be the responsibility of `RulesFactory` to know about eligibility -- thoughts on creating a factory class for `Eligiblity` similar to `RulesFactory` that we can call here?

app/models/eligibilities/eight_a_stars.rb
```diff
@@ -0,0 +1,8 @@
+module Eligibility
```
jessieay: I've been namespacing on one line in this project, so `Eligiblity::EightaStars` . Not hugely different but helpful for knowing that a module is for namespacing rather than including behavior, imo

app/models/auction.rb
```diff
@@ -1,5 +1,5 @@
 class Auction < ActiveRecord::Base
-  MAX_START_PRICE = 3500
+  MICROPURCHASE_THRESHOLD = 3500
```
jessieay: much better! Yay naming 👏

jessieay:
this looks awesome, I really like the direction you're taking it!

app/models/auction.rb
harrisj: Huzzah!

app/models/auction.rb
```diff
@@ -32,12 +32,12 @@ def lowest_amount
   end
 
   def start_price_equal_to_or_less_than_max_if_not_contracting_officer
-    if user && !user.contracting_officer? && start_price > MAX_START_PRICE
+    if user && !user.contracting_officer? && start_price > MICROPURCHASE_THRESHOLD
```
harrisj: Not really specific to this feature, but are we not allowing non-CO users to create/save auctions greater than the micropurchase threshold? I thought we were just preventing them from publishing them.

app/models/place_bid.rb
```diff
@@ -65,7 +60,7 @@ def validate_bid_data
       fail UnauthorizedError, 'Bids must be in increments of one dollar'
     end
 
-    if amount > BID_LIMIT
+    if amount > auction.start_price
```
harrisj: Ooops, this is a bug we should've fixed.

app/models/rules_factory.rb
adelevie: I'm good with adding another factory.

app/models/eligibilities/eight_a_stars.rb
adelevie: I'm good with that.

app/models/auction.rb
adelevie: That's a good point ... we should clear that up with @mtorres253.

app/models/place_bid.rb
adelevie: We got lucky since no auctions started at a number different than 3,500, but yay!

adelevie:
I'm going to remove the 8 a stars stuff so this will be merge-able.

app/models/eligibilities/in_sam.rb
```diff
@@ -0,0 +1,7 @@
+module Eligibility
+  class InSam
+    def eligible?(user)
```
harrisj: I think I'd prefer to clarify eligibility in terms of actions like `user_can_bid?`

app/models/rules_factory.rb
harrisj: I agree. I think we would basically have a separate column in the database so we could indicate the auction's rules and the population available to participate in it (or the second one might be derived from looking at extant fields already in the DB). But it would be a separate factory.

Instead of calling it Eligibility, maybe we should consider some other term of art like `Pool` or `Population` or such?

app/presenters/auction_presenter.rb
```diff
@@ -151,6 +151,15 @@ def auction_rules
...
 
+  def small_business?
+    start_price > Auction::MICROPURCHASE_THRESHOLD
+  end
+
+  def eight_a_stars
+    # this could be a db column
+    true
+  end
+
```
harrisj: Following the discussion above, we would probably eliminate both of these boolean methods and instead have a factory that returns the right `Eligibility` class based on one or more conditions in the auction.

adelevie: 👍

app/models/rules_factory.rb
adelevie: I've seen `Pool` in the Agile BPA context and I don't think that's what we want here. A pool would be more of a vetted/pre-approved list of vendors for a specific purpose.

FWIW I really kind of do like eligibility. It sort of fits well with the error messages we would display to a user:

```
Only small businesses are eligible to bid on this auction.
```

What are the issues you see in it, @harrisj?

jessieay: I am fine with sticking to eligiblity for now. I am guessing another term may emerge as we learn more about different types of auctions and how their rules are set

harrisj: Oh I don't know. It just feels a bit weird since it's a relatively abstract noun and I don't know if we should be considering what the class is for (although I can only think of something ugly like `VendorGroup` and in procurement these are known as `SetAside`) or something else. Never mind me

adelevie:
closing in favor of https://github.com/18F/micropurchase/pull/558, which implements eligibilities and addresses the acceptance criteria with small biz messaging

PR #542, opened 2016-05-12
Update error msg for auction start price error

jessieay:
* Had key "start price" so read "Start price You do not have permission to publish
  auctions with a start price over $3500"
* Now just "You do not have permission to publish auctions with a start
* price over $3500"

Updated:

screen shot 2016-05-12 at 4 06 10 pm


harrisj:
LGTM. Good catch.

PR #548, opened 2016-05-13
Add Ruby version to rubocop

jessieay:
* Otherwise it doesn't know about cool Ruby 2+ things like keyword args


harrisj:
LGTM. Merging

PR #550, opened 2016-05-13
Do not check SAM on each login

jessieay:
* We already check each time DUNs is updated, why check on login?


harrisj:
To provide some context, this was implemented before we decided to have a DelayedJob and cron checking SAM.

jessieay:
@harrisj should we still do it? we could throw a delay on there. not sure if SAM status changes regularly or not

harrisj:
No, we can get rid of it. If we are doing the other mechanism, we might as well just use one. I just wanted to provide some context.

harrisj:
Looks good to me. Merging

PR #557, opened 2016-05-14
Bids now record the method they were posted via

### LINKED ISSUE #526: Add a bid mechanism field to bids?
I was wondering if it would make sense to add a column named something like `mechanism` or `method` to the bids column that could be used to record if the bid was made via the `web` interface or the `api`? (and could possibly also be `auto` if we had automatic bidding)

harrisj:
It uses 'web' for HTTP access and 'api' for bids posted via the API. I figured this would be helpful for assessing how bids were being submitted.

Fixes #526

### DIFF
--- a/app/controllers/bids_controller.rb
+++ b/app/controllers/bids_controller.rb
@@ -32,14 +32,14 @@ class BidsController < ApplicationController
 
   def confirm
     @auction = AuctionViewModel.new(current_user, Auction.find(params[:auction_id]))
-    @bid = BidPresenter.new(PlaceBid.new(params, current_user).dry_run)
+    @bid = BidPresenter.new(PlaceBid.new(params, current_user, via).dry_run)
   end
 
   def create
     unless current_user.sam_accepted?
       fail UnauthorizedError, "You must have a valid SAM.gov account to place a bid"
     end
-    @bid = BidPresenter.new(PlaceBid.new(params, current_user).perform)
+    @bid = BidPresenter.new(PlaceBid.new(params, current_user, via).perform)
 
     respond_to do |format|
       format.html do
--- a/app/controllers/bids_controller.rb
+++ b/app/controllers/bids_controller.rb
@@ -32,14 +32,14 @@ class BidsController < ApplicationController
 
   def confirm
     @auction = AuctionViewModel.new(current_user, Auction.find(params[:auction_id]))
-    @bid = BidPresenter.new(PlaceBid.new(params, current_user, via).dry_run)
+    @bid = BidPresenter.new(PlaceBid.new(params: params, user: current_user, via: via).dry_run)
   end
 
   def create
     unless current_user.sam_accepted?
       fail UnauthorizedError, "You must have a valid SAM.gov account to place a bid"
     end
-    @bid = BidPresenter.new(PlaceBid.new(params, current_user, via).perform)
+    @bid = BidPresenter.new(PlaceBid.new(params: params, user: current_user, via: via).perform)
 
     respond_to do |format|
       format.html do
--- a/spec/controllers/bids_controller_spec.rb
+++ b/spec/controllers/bids_controller_spec.rb
@@ -127,12 +127,7 @@ RSpec.describe BidsController, controller: true do
     end
 
     context 'when there are no other bids' do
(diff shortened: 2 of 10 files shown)


jessieay:
tiny little code climate error on this one

harrisj:
CodeClimate issue fixed

spec/controllers/bids_controller_spec.rb
```diff
@@ -127,32 +127,47 @@
     end
 
     context 'when there are no other bids' do
-      before do
-        allow(PlaceBid).to receive(:new).and_return(place_bid)
-      end
+      # before do
```
jessieay: delete?

spec/controllers/bids_controller_spec.rb
```diff
@@ -127,32 +127,47 @@
...
       end
 
-      it "adds a flash error when the bid is bad" do
-        expect(PlaceBid).to receive(:new).with(anything, current_bidder)
-          .and_raise(UnauthorizedError.new("Bad bid, sucker!"))
-        post :create, request_params, user_id: current_bidder.id
-        expect(flash[:error]).to eq("Bad bid, sucker!")
-        expect(response).to redirect_to("/auctions/#{auction.id}/bids/new")
+      context 'when the bid is bad' do
+        let(:request_params) do
```
jessieay: thoughts on inlining since this is just used in one test?

spec/requests/bids_spec.rb
```diff
@@ -53,6 +54,10 @@
       it 'returns a 404 status code' do
         expect(status).to eq(404)
       end
+
+      it 'should not create a bid' do
+        expect(auction.bids.count).to eq(@bid_count)
```
jessieay: since ivars can be reassigned wherever, not sure I am a big fan of using them for rspec expectations.

thoughts on moving the "exercise" into each spec (`post auction_bids_path(auction), params, headers`) and then testing it like this:

```ruby
expect do
  post auction_bids_path(auction), params, headers
end.not to { auction.bids.count }
```

spec/requests/bids_spec.rb
```diff
@@ -253,6 +304,9 @@
         it 'returns a 403 status code' do
           expect(status).to eq(403)
         end
+
+        it 'should not create a bid' do
```
jessieay: remove empty spec?

spec/requests/bids_spec.rb
```diff
@@ -241,6 +290,8 @@
         it 'returns a 403 status code' do
           expect(status).to eq(403)
         end
+
+        it 'should not create a bid'
```
jessieay: remove?

spec/requests/bids_spec.rb
jessieay: an added bonus for inlining the exercise is that it can be hard to see what is being tested when the exercise just happens once per file. Similar arg to not using `let` but I do think that having the *what is being tested* line exist within each test

app/controllers/bids_controller.rb
```diff
@@ -32,14 +32,14 @@ def new
 
   def confirm
     @auction = AuctionViewModel.new(current_user, Auction.find(params[:auction_id]))
-    @bid = BidPresenter.new(PlaceBid.new(params, current_user).dry_run)
+    @bid = BidPresenter.new(PlaceBid.new(params, current_user, via).dry_run)
```
jessieay: woooo! thoughts on named args for `BidPresenter` so we can be less dependent on arg order?

jessieay:
looks very close -- back to you, @harrisj 🐶  ⚾️

spec/requests/bids_spec.rb
harrisj: Yep, going to tweak all that now.

spec/requests/bids_spec.rb
harrisj: Implementing it instead

app/controllers/bids_controller.rb
harrisj: We can try named args for PlaceBid

harrisj:
Back to you @jessieay

jessieay:
🌈  🙌

Issue #496, opened 2016-04-29
Messaging to vendors about auctions with start price over $3500


mtorres253:
As a vendor that is not a small business, I want to know when auctions are not available for me to bid on, so the system should tell me that. 

Scenario: A vendor sees small business only auctions on the auction feed
Given I am a vendor
And there are auctions that are only available for small businesses
When I view the auction homepage (https://micropurchase.18f.gov/)
Then the previews of those auctions should show that they are only available for small businesses

Note: We can discuss this but maybe we should just have a tag that shows "Small business only"

Scenario: A vendor sees small business only auctions on the auction view page
Given I am a vendor
And I am NOT a small business
And there is an auction that is only available for small businesses
When I view the view page for that auction (https://micropurchase.18f.gov/auctions/[number])
Then I should not see the bid button
And instead I should see this message "This auction is only available for small businesses."

adelevie:
@jessieay were you working on this? I moved this into the "current" column a few days ago, but forgot to assign myself.

jessieay:
@adelevie nope, all you friend 🏀

adelevie:
🎷

PR #558, opened 2016-05-14
Small business messaging to vendors

adelevie:
Fixes https://github.com/18F/micropurchase/issues/496


app/models/rules_factory.rb
```diff
@@ -4,10 +4,12 @@ def initialize(auction)
   end
 
   def create
+    eligibility = EligibilityFactory.new(auction).create
```
harrisj: We can look at this later, but I am a little leery of one Factory encapsulating another, especially since the Eligibility is really only used in `user_can_bid?` (a method I would argue should just be pulled from Rules, although there is the issue of sealed-bid auctions...) But I suppose we can fix that later.

app/presenters/auction_presenter.rb
```diff
@@ -140,6 +145,14 @@ def url
     "#{root_url}/auctions/#{id}"
   end
 
+  def eligibility
```
harrisj: This should probably be delegated to the Eligibility object itself (maybe with a name like `eligibility_label` or such

app/models/eligibility_factory.rb
```diff
@@ -0,0 +1,15 @@
+class EligibilityFactory
+  attr_reader :auction
+
+  def initialize(auction)
+    @auction = AuctionPresenter.new(auction)
+  end
+
+  def create
```
jessieay: BEAUTIFUL ✨

app/models/rules_factory.rb
jessieay: if it were me, I might stick this up in `initialize` to move it as far outside of this obj as possible. The next step would be to pass it in as an arg to `initialize`, but like @harrisj says, that can happy later

app/presenters/auction_presenter.rb
jessieay: Having an `EligibilityPresenter` for each type would be cool, too

app/services/start_price_thresholds.rb
```diff
@@ -0,0 +1,8 @@
+class StartPriceThresholds < Struct.new(:start_price)
+  MICROPURCHASE = 3500
+  SAT = 150000
+
+  def small_business?
```
jessieay: what if `SmallBusinessEligibility` had a ` small_business?` method that returned `true` and `InSamEligibility` returned `false` and we depended on those for this logic?

jessieay: seems like whether an auction is for small businesses is a cncern of eligibility more than start price threshold

app/services/start_price_thresholds.rb
```diff
@@ -0,0 +1,8 @@
+class StartPriceThresholds < Struct.new(:start_price)
```
jessieay: I think this could be in `app/models` - I'd expect most classes in `app/services` to have a `perform` method that updates or creates something

features/step_definitions/auction_steps.rb
```diff
@@ -19,11 +19,37 @@
...
+    .find(:xpath, "..")
+
+  expect(auction_div).to have_content('Small-business only')
+end
+
+Then(/^I should not see that the auction indicates it is for small business only$/) do
+  auction_div = find('a', text: @auction.title)
+    .find(:xpath, "..")
+    .find(:xpath, "..")
+    .find(:xpath, "..")
```
jessieay: we sure there isn't a simpler way to find the correct selector? do you use chrome to find the xpath selector? (that's what I do)

features/step_definitions/auction_steps.rb
```diff
@@ -35,13 +61,17 @@
...
-  visit "/auctions/#{@auction.id}"
+  visit(auction_path(@auction.id))
 end
 
 When(/^I visit the unpublished auction$/) do
   visit "/auctions/#{@unpublished_auction.id}"
 end
 
+When(/^I visit the auction bids page$/) do
+  visit(auction_bids_path(@auction.id))
```
jessieay: protip: you can pass `@auction` rather than `@auction.id` when using named routes like this. Rails automatically calls `to_param` on any object passed into a named routed, and the default `to_param` def returns `id`

features/step_definitions/auction_steps.rb
```diff
@@ -171,8 +201,26 @@
   expect(page).to have_content("Delivery deadline: #{DcTimePresenter.convert_and_format(@auction.delivery_deadline)}")
 end
 
+Then(/^I should see the bid button$/) do
+  within(:css, 'div.auction-info') do
```
jessieay: nice!

jessieay:
merge conflicts @adelevie  :(

adelevie:
rebased 🎺

adelevie:
Rebase 2: Conflict Resolution

features/step_definitions/auction_steps.rb
adelevie: I went with this approach because it lets me select the entire outer div given an instance of `Auction`/`AuctionPresenter`. Our markup doesn't otherwise make those divs uniquely selectable.

app/presenters/auction_presenter.rb
```diff
@@ -65,6 +65,11 @@ def initialize(auction)
     to: :auction_rules
   )
 
+  delegate(
+    :small_business?,
```
jessieay: how about chaining the methods here rather than adding more delegations? will help in our effort to make law of demeter violations more obvious for future refactors

app/models/start_price_thresholds.rb
```diff
@@ -0,0 +1,8 @@
+class StartPriceThresholds < Struct.new(:start_price)
+  MICROPURCHASE = 3500
+  SAT = 150000
+
+  def small_business?
```
jessieay: since this is only used in one eligiblity (`small_business_eligibility.rb`) - thoughts on moving this method there?

Issue #495, opened 2016-04-29
On demand report containing sam.gov data


mtorres253:

mtorres253:
As an admin user, I want to be able to send required reports without having to manually find data, so if Sam.gov has the data then I should be able to easily get it from the system.

Scenario: I get a .csv file containing system auction data
Given I am an authenticated admin user
And there are auctions in the system with start prices over $3500
And those auctions have ended and there was a winner
And when I click on the link to generate a .csv report
Then the file should be generated
And the file should contain the following data from Sam.gov for the winning vendor

```csv
13GG Vendor Name, [Sam.gov Vendor Name]
13JJ Vendor Address Line 1, [Sam.gov Address Line 1]
13KK Vendor Address Line 2, [Sam.gov Address Line 2]
13LL Vendor Address Line 3, [Sam.gov Address Line 3]
13MM Vendor Address City, [Sam.gov City]
13NN Vendor Address State, [Sam.gov State]
13PP Vendor Zip Code, [Sam.gov Zip code]
13QQ Vendor Country Code, [Sam.gov Country Code]
13RR Vendor Phone Number, [Sam.gov Phone Number]
13SS Vendor Fax Number, [Sam.gov Fax Number]
```

jessieay:
@mtorres253 suggested csv headers for these fields? should we just use the names in the text blob above?

mtorres253:
The headers can be "FPDS Field", "Value". Does that work?

mtorres253:
Based on today's convo, we're sticking with standard .csv format (column header, value)

PR #575, opened 2016-05-16
On demand report containing sam.gov data

jessieay:
Closes https://github.com/18F/micropurchase/issues/495

screen shot 2016-05-16 at 11 25 17 am

When DUNS is not valid, shows a flash message saying that the DUNS is not valid

### DIFF
--- a/app/models/winning_bid.rb
+++ b/app/models/winning_bid.rb
@@ -1,0 +1,17 @@
+class WinningBid
+  def initialize(auction)
+    @auction = auction
+  end
+
+  def find
+    auction_rules.winning_bid
+  end
+
+  private
+
+  attr_reader :auction
+
+  def auction_rules
+    RulesFactory.new(auction).create
+  end
+end
--- a/app/models/winning_bidder_export.rb
+++ b/app/models/winning_bidder_export.rb
@@ -1,0 +1,118 @@
+require 'csv'
+
+class WinningBidderExport
+  class Error < StandardError; end
+
+  def initialize(auction)
+    @auction = auction
+  end
+
+  def export_csv
+    CSV.generate do |csv|
+      csv << header_values
+      csv << data_values
+    end
+  end
+
+  private
+
+  attr_reader :auction
+
+  def header_values
+    [
+      "13GG Vendor Name",
+      "13JJ Vendor Address Line 1",
+      '13KK Vendor Address Line 2',
+      '13LL Vendor Address Line 3',
+      '13MM Vendor Address City',
+      '13NN Vendor Address State',
+      '13PP Vendor Zip Code',
+      '13QQ Vendor Country Code',
+      '13RR Vendor Phone Number',
+      '13SS Vendor Fax Number'
+    ]
+  end
+
+  def data_values
+    [
+      legal_business_name,
+      address_line_1,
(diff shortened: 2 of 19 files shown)


app/models/winning_bid.rb
```diff
@@ -0,0 +1,17 @@
+class WinningBid
```
harrisj: I'm not sure I understand why this needs to be a separate class since it's delegating for all of its functionality... Also, should it be in models since it isn't AR-backed?

app/models/winning_bidder_export.rb
```diff
@@ -0,0 +1,118 @@
+require 'csv'
+
+class WinningBidderExport
```
harrisj: Nice class

config/application.rb
```diff
@@ -9,35 +9,13 @@
 require "action_mailer/railtie"
 require "action_view/railtie"
 require "sprockets/railtie"
```
harrisj: Thank you, but we should probably do this kind of cleanup in separate PRs?

features/step_definitions/auction_steps.rb
```diff
@@ -35,11 +35,16 @@
 end
 
 When(/^I visit the auction page$/) do
-  visit "/auctions/#{@auction.id}"
+  visit auction_path(@auction)
```
harrisj: Thank you!

spec/rails_helper.rb
```diff
@@ -1,4 +1,4 @@
-ENV['RAILS_ENV'] = 'test'
+ENV['RAILS_ENV'] ||= 'test'
```
harrisj: Now this is a ||= again. Did we ever figure out why it was changed before? Just curious...

spec/rails_helper.rb
```diff
@@ -31,6 +31,7 @@
 
   config.before do
     mock_github
+    WebMock.stub_request(:any, /api.data.gov/).to_rack(FakeSamApi)
```
harrisj: Ooooh, I didn't know we could do this!

spec/rails_helper.rb
harrisj: Is this even needed considering this is also in the spec_helper.rb?

harrisj:
I have a few questions, but nothing that should hold up a merge...

config/application.rb
jessieay: yes agreed, got overzealous!

app/models/winning_bid.rb
jessieay: we seem to have quite a few non-AR models. I am fine with putting anything model-y in `app/models`

I made it a separate class bc I realized I was instantiating the rules factory a bunch and then finding winning bidder based on that

spec/rails_helper.rb
jessieay: oh d'oh I changed this when debugging a weird issue -- my rails console thinks it is in test env by default. can't figure out why!

spec/rails_helper.rb
jessieay: yes! @adelevie and I worked on this approach when I was in DC and now we have a nice setup for stubbing out the SAM API

PR #576, opened 2016-05-16
Remove default screenshots

jessieay:
* Were saving to root dir for some reason
* Can throw in a screenshot as a debugging mechanism without this lib
* See https://github.com/teampoltergeist/poltergeist#taking-screenshots-with-some-extensions


harrisj:
LGTM. Merging

PR #577, opened 2016-05-16
Small tweak to fix CSV Reporting for SealedBids

jessieay:
* Error on staging: 'undefined method `available?` for `auction`
* SealedBid` rules looks at `auction.available?` in several methods
* `available? `is a method we delegate from `app/view_models/auction_view_model.rb`
  to `app/models/auction_status.rb`
* so you basically need a view model in order to look at the rules
* This tweak fixes for now, but going to update further in following PR


adelevie:
Looks good!

PR #579, opened 2016-05-16
Be explicit when querying AuctionStatus

jessieay:
* This should help uncover where we might want to present auctions differently based on status

Should be merged after https://github.com/18F/micropurchase/pull/577


adelevie:
Looks good! RIP to those `let` calls.

Issue #494, opened 2016-04-29
On demand report contains system data


mtorres253:
As an admin user, I want to be able to send required reports without having to manually find data, so if the system has the data then I should be able to easily get it from the system.

Scenario: I get a .csv file containing auction data that is stored in the system
Given I am an authenticated admin user
And there is an auction
And the auction has ended (closed auctions)
When I click on the link to "Generate a .csv Report" on the admin auction show page (https://micropurchase.18f.gov/admin/auctions/[number])
Then the file should be generated
And the file should contain the following default data (In addition to the Sam.gov data for the vendor)

```csv
2B Effective Date, [Auction End Date] (the starting date of the contracts requirements) - Date (YYYYMMDD)
2C Current Completion Date, [Auction Delivery Date] (the completion date of the contract) - Date (YYYYMMDD)
3A Base And All Options Value, [Award Amount - if there is a winner]
3B Base And Exercised Options Value, [Award Amount - if there is a winner]
3C Action Obligation, [Award Amount - if there is a winner]
6M Description of Requirement, Auction Description
6N Purchase Card as Payment Method, Y (until we move to some other type of payment)
6R National Interest Action, None
9A DUNS Number, [DUNS number of winning vendor - if there is a winner]
9B Contractor Name from Contract, [Business name from Sam.gov]
10J Commercial Item Test Program, N
10M Solicitation Procedures, SP1
```

https://www.acquisition.gov/far/html/Subpart%2013_1.html#wp1125361

PR #580, opened 2016-05-17
Add more fields to CSV report

jessieay:
* Only show link to generate when auction is over
* Closes https://github.com/18F/micropurchase/issues/494

screen shot 2016-05-16 at 5 13 45 pm


harrisj:
I am okay with the CodeClimate issues

jessieay:
@harrisj cool - me too. Wish there were a way to exclude arrays from code climate method length alerts

jessieay:
@adelevie this is ready for review / merge when you are

app/models/winning_bidder_export.rb
```diff
@@ -3,6 +3,11 @@
 class WinningBidderExport
   class Error < StandardError; end
 
+  PURCHASE_CARD_AS_PAYMENT_METHOD = 'Y'
+  NATIONAL_INTEREST_ACTION = 'None'
+  COMMERCIAL_ITEM_TEST_PROGRAM = 'N'
+  SOLICITATION_PROCEDURES = 'SP1'
+
```
adelevie: CodeClimate wants `#freeze` called on these strings.

app/models/winning_bidder_export.rb
```diff
@@ -29,7 +34,19 @@ def header_values
...
+      '3A Base And All Options Value',
+      '3B Base And Exercised Options Value',
+      '3C Action Obligation',
+      '6M Description of Requirement',
+      '6N Purchase Card as Payment Method',
+      '6R National Interest Action',
+      '9A DUNS Number',
+      '9B Contractor Name from Contract',
+      '10J Commercial Item Test Program',
+      '10M Solicitation Procedures'
```
adelevie: CodeClimate does not love the size of this class. Maybe stick the above array into its own class or module called something like `WinningBidderExportFields`.

adelevie:
Just the two minor comments from CodeClimate. @jessieay if you think they're not worth doing (or just worth doing in a later PR), let me know and I'll merge.

app/models/winning_bidder_export.rb
jessieay: cool I can do that

app/models/winning_bidder_export.rb
jessieay: I am also against long classes, but when the majority of the class is just long arrays I tend think it's worth it to have the longer class rather than deal with additional indirection of throwing arrays in a new class. let's keep this one long for now and move stuff out of we feel like it gets too big/complicated

jessieay:
@adelevie updated! 📈

adelevie:
errors! 😿

jessieay:
argh should be fixed now, thanks @adelevie

jessieay:
review me, please!

PR #583, opened 2016-05-17
Use `Guest` rather than `nil` when no user

jessieay:
* OO!
* Duck typing!
* All the things!
* Through this refactor, also discovered an entire view partial that was
  rendered but did not show on page...weird


app/views/components/_nav_drawer.erb
```diff
@@ -1,34 +0,0 @@
-<drawer class="usa-width-one-fourth drawer" aria-hidden="true">
```
jessieay: I think `aria-hidden` is why this was not showing...but in any case, we didn't need it!

adelevie:
amazing

Issue #538, opened 2016-05-12
Index page default sorting


mtorres253:
As a vendor or unknown user, I want to see information in a particular order, so I know how to find things on the micropurchase homepage

Scenario: Default ordering for the micropurchase homepage
Given I am a registered vendor or an unknown user
When I look at the micropurchase homepage
Then the auction previews should be in descending order by end date timestamp
And the order should go left to right and top to bottom

Note: There are now two columns in the layout so the order will be left->right and top->bottom. https://pages.18f.gov/micropurchase/. This story contemplates just the new grid, not the actual styling

gemfarmer:
@mtorres253 to clarify, descending order by date created, _with the most recently created auction in the top left_?

@adelevie @harrisj @jessieay how is this currently ordered? If/when we add additional filtering, I assume that we would prefer this to be done in javascript?

jessieay:
@gemfarmer currently, index contains published auctions in reverse chron order (using end datetime timestamp)

PR #584, opened 2016-05-17
New auctions grid

gemfarmer:
For #538 

I made the grid 2 wide to emulate the [micropurchase prototype](https://pages.18f.gov/micropurchase/).

* used flexbox to create same-heighted cards
* did not change style of auction cards
* wrote rspec tests to make sure that they are ordered by `date_endtime`

### DIFF
--- a/spec/controllers/auctions_controller_spec.rb
+++ b/spec/controllers/auctions_controller_spec.rb
@@ -43,9 +43,9 @@ describe AuctionsController do
         auction_3 = auctions[2]
 
         expect(auction_1).to be_a(AuctionViewModel)
-        expect(auction_1.end_datetime).to eq(date_latest)
-        expect(auction_2.end_datetime).to eq(date_middle)
-        expect(auction_3.end_datetime).to eq(date_first)
+        expect(auction_1.end_datetime).to be_within(0.1).of(date_latest)
+        expect(auction_2.end_datetime).to be_within(0.1).of(date_middle)
+        expect(auction_3.end_datetime).to be_within(0.1).of(date_first)
       end
     end
   end
--- a/spec/controllers/auctions_controller_spec.rb
+++ b/spec/controllers/auctions_controller_spec.rb
@@ -15,10 +15,10 @@ describe AuctionsController do
 
         Timecop.freeze
 
-        date_start = Time.now - 3.days
-        date_latest = Time.now + 3.days
-        date_middle = Time.now + 2.days
-        date_first = Time.now + 1.days
+        date_start = Time.current - 3.days
+        date_latest = Time.current + 3.days
+        date_middle = Time.current + 2.days
+        date_first = Time.current + 1.days
 
         auction_record_1 = create(
           :auction,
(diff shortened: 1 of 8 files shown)


gemfarmer:
@jessieay this is passing locally. Any idea why this would be failing [here](https://travis-ci.org/18F/micropurchase/builds/130954767#L571)? The times look slightly different, but I am storing those values in variables, so I'm not sure why those would be different...

spec/controllers/auctions_controller_spec.rb
```diff
@@ -9,6 +9,45 @@
...
+
+        get :index
+        auctions = assigns(:view_model).auctions
+
+        auction_1 = auctions[0]
+        auction_2 = auctions[1]
+        auction_3 = auctions[2]
+
+        expect(auction_1).to be_a(AuctionViewModel)
+        expect(auction_1.end_datetime).to eq(date_latest)
```
jessieay: why not check that `expect(auction_1.end_datetime).to eq(auction_record_2)` ? checking datetime equality is annoying bc there is so much formatting stuff.

spec/controllers/auctions_controller_spec.rb
```diff
@@ -9,6 +9,45 @@
       expect(auction).to be_a(AuctionViewModel)
       expect(auction.id).to eq(auction_record.id)
     end
+
+    context 'the list of auctions is sorted' do
+      it 'renders them descending by datetime' do
+
+        Timecop.freeze
+
+        date_start = Time.now - 3.days
```
jessieay: you want to use `Time.current` instead of `Time.now` -- might be reason for CI failures  http://winstonyw.com/2014/03/03/time_now_vs_time_zone_now/

spec/controllers/auctions_controller_spec.rb
gemfarmer: do you mean just check to make sure that the auction records equal each other instead of one of their underlying attributes? Like this:

`expect(auction_1).to eq(auction_record_2)`

spec/controllers/auctions_controller_spec.rb
gemfarmer: hmm it looks like the difference between `Time.current` and `Time.now` is that `current` has a timezone fallback. Not sure how that would help with millisecond differences like I am seeing here (only on travis, not locally):

```
expected: 2016-05-20 20:55:16.235439941 +0000
     got: 2016-05-20 20:55:16.235439000 +0000
```

spec/controllers/auctions_controller_spec.rb
gemfarmer: Interesting article. http://batasrki.github.io/blog/2015/01/29/rspec-matcher-and-time-value/

Looks like `be_within()` is an acceptable solution.

jessieay: yeah I mean look at records rather than underlying attributes. If you name the records according to their time attributes, then the spec expectations will still be helpful in telling us the ordering. eg:

```
expect(auctions[0]).to eq(most_recent_auction)
expect(auctions[1]).to eq(middle_auction)
expect(auctions[2]).to eq(oldest_auction)
```

gemfarmer:
This ain't a WIP any longer, folks! @jessieay mind giving it a final review?

cc @mtorres253 and @andrewmaier for visibility. Feel free to take a look/review as well!

jessieay:
woo hoo 💭  👍

gemfarmer:
woot. Thanks @jessieay !

Issue #571, opened 2016-05-16
Vendors that are not small businesses cannot bid on auctions > $3500 via API


mtorres253:
As an admin user, I want to prevent vendors that are not small businesses from bidding on an auction that is only open to small businesses via API, so the system should prevent that vendor from bidding on that auction and should communicate the reason

Scenario: Vendor is not small business and tries to bid on an auction >$3500
Given I am a vendor
And there is an auction with start price >$3500
And I am not a small business
When I try to place a bid on that auction using the API
Then my bid should not be accepted
And I should get an error message [403 not authorized - You must be a small business to bid on this auction]

PR #585, opened 2016-05-18
Ensure vendors that are not small businesses cannot bid on auctions > $3500 via API

adelevie:
fixes https://github.com/18F/micropurchase/issues/571

I'm not sure we should have cucumber features for an API story. If I'm wrong, let me know and I'll try to add some.

You'll notice this PR doesn't contain any new application code, just a new request spec (tests the API) and specs on `PlaceBid` (controls all bid placement) to ensure that vendors who aren't small businesses cannot bid on small-business-only auctions.

### DIFF
--- a/spec/models/place_bid_spec.rb
+++ b/spec/models/place_bid_spec.rb
@@ -31,6 +31,37 @@ RSpec.describe PlaceBid do
                        end_datetime: Time.now + 7.days)
   end
 
+  context 'when the auction start price is above the micro-purchase threshold' do
+    let(:auction) { create(:auction, :between_micropurchase_and_sat_threshold) }
+
+    context 'and the vendor is not a small business' do
+      let(:current_user) { create(:user, :not_small_business) }
+
+      it 'should reject the bid' do
+        expect { place_bid.perform }.to raise_error(UnauthorizedError)
+      end
+
+      it 'should not create a new bid' do
+        # need to rescue otherwise the raising of UnauthorizedError
+        # causes the test to error before the change can be evaluated
+        expect do
+          begin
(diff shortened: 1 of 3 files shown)


harrisj:
We haven't done stories for the other API requests. If we do, I'd suggest we at least tag them with `@api` tags

adelevie:
The `PlaceBid` and request specs seem to me to provide enough assurance. Maybe also add a controller spec to `expect(PlaceBid).to receive(:new)` to describe `BidsController#create`?

spec/models/place_bid_spec.rb
```diff
@@ -31,6 +31,26 @@
...
   end
 
+  context 'when the auction start price is above the micro-purchase threshold' do
+    let(:auction) { create(:auction, :between_micropurchase_and_sat_threshold) }
+
+    context 'and the vendor is not a small business' do
+      let(:current_user) { create(:user, :not_small_business) }
+
+      it 'should reject the bid' do
+        expect { place_bid.perform }.to raise_error(UnauthorizedError)
```
harrisj: Should we add a check that it did not change the bid count?

jessieay:
I like the idea of doing request specs for API tests (rather than cucumber or controller specs). 

The reason for request specs over controller, imo, is that in request specs we can inspect the request body so we can confirm that the JSON matches what we we expect. I think we are already doing this.

harrisj:
I think if we can just add a quick check that the bids count did not change for the error case then I'm happy

adelevie:
Done!

harrisj:
Thank you. Just waiting for Travis to give me the go-ahead and I will merge

PR #586, opened 2016-05-18
Moved eligibility labels into the eligibility classes

adelevie:
My attempt at implementing some of the feedback from https://github.com/18F/micropurchase/pull/558.

In this PR, `AuctionPresenter` is made to know nothing about strings/labels related to eligibilities. It's delegated out to the eligibility classes via the `Rules::BaseRules` class.

### DIFF
--- a/app/presenters/auction_presenter.rb
+++ b/app/presenters/auction_presenter.rb
@@ -137,12 +137,8 @@ class AuctionPresenter
     "#{root_url}/auctions/#{id}"
   end
 
-  def eligibility
-    if small_business?
-      'Small-business only'
-    else
-      'SAM.gov only'
-    end
+  def eligibility_label
+    auction_rules.eligibility.label
   end
 
   private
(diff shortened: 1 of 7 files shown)


app/presenters/auction_presenter.rb
```diff
@@ -137,12 +137,8 @@ def url
...
   end
 
-  def eligibility
-    if small_business?
-      'Small-business only'
-    else
-      'SAM.gov only'
-    end
+  def eligibility_label
+    auction_rules.eligibility.label
```
harrisj: We're getting Demeter a bit angry here, but let's work on that later

harrisj:
I approve of replacing these sort of conditionals with polymorphism.

So, we seem to be stuck with the current approach between a fair amount of delegation or violating the Law of Demeter. I don't want to hold up this PR, but it's something that would be good to resolve in the future. Merging.

Issue #572, opened 2016-05-16
Check Sam.gov on register to see if vendor is a small business


mtorres253:
As an admin user, I want to know which vendors are small businesses as defined by Sam.gov, so the system should check with Sam.gov when the vendor saves a DUNS number for the first time.

Scenario: System checks Sam.gov if vendor is a small business when vendor saves a DUNS number for the first time
Given that I am vendor
When I save a duns number in my profile for the first time
Then the system checks Sam.gov to see if I am a small business
And the system stores the response

Note: this might have to be part of the job that already exists to validate the vendor's DUNS number

PR #588, opened 2016-05-18
Check for small biz on register

adelevie:
aims to fix https://github.com/18F/micropurchase/issues/572

### refactoring / house cleaning:

- Removed the unused `Authenticator` class
- Because the story requires something to happen "on register" I thought it would be good to disambiguate `User` registration from creation (previously tied together by `User.find_or_create` in `LoginUser`. To accomplish this, I created a `FindOrRegister` class, which lightly wraps `User.create` and the newly-created `RegisterUser` class. This way we can easily add in any new "on register" events and test for them.

### changes to SamAccountReckoner

My original plan was to trigger a new job somewhere in the newly created `RegisterUser` class. After some time on this and chatting with @mtorres253, we changed the criteria to require the small business check to occur when a user updates their DUNS number. `SamAccountReckoner` seemed like the most logical place to add this functionality: it's already properly handles DUNS updates and has access to an instance of `Samwise::Client`. So most of this new feature occurs in `SamAccountReckoner#set!` and some of its new private methods.

### changes to samwise

This PR caused me to update the Samwise gem (https://github.com/18F/samwise/pull/29) so that in a single HTTP call we could determine if the vendor was in SAM and if the vendor is a small business. This PR includes an update in the Gemfile to use the newly-release 0.3.0 version of Samwise.


app/services/find_or_register_user.rb
```diff
@@ -0,0 +1,27 @@
+class FindOrRegisterUser
+  attr_reader :github_id
+
+  def initialize(params)
+    @github_id = params[:github_id]
+  end
+
+  def perform
+    @_user = found_user = find_user
```
jessieay: what about just using a local var here instead of instance? seems like we don't need it outside of this method?

adelevie: 👍

app/services/register_user.rb
```diff
@@ -0,0 +1,11 @@
+class RegisterUser
+  attr_reader :github_id
+
+  def initialize(params)
+    @github_id = params[:github_id]
+  end
+
+  def perform
+    @_user = User.create(github_id: github_id)
```
jessieay: I usually use `@_var` when doing memoization (`||/`) - should we do that here?

app/services/find_or_register_user.rb
```diff
@@ -0,0 +1,27 @@
...
+    @_user
+  end
+
+  private
+
+  def find_user
+    User.where(github_id: github_id).first
+  end
+
+  def register_user
```
jessieay: thoughts on naming this something like `CreateUser` since that is what we are doing in that class?

app/services/find_or_register_user.rb
```diff
@@ -0,0 +1,27 @@
+class FindOrRegisterUser
```
jessieay: `FindOrCreateUser` ?

app/services/register_user.rb
adelevie: I guess it couldn't hurt to memoize. It seems like it would be Bad to call `SomeServiceClass#perform` more than once, so I guess memoizing protects us from that.

app/models/login_user.rb
```diff
@@ -7,7 +7,7 @@ def perform
   private
 
   def user
-    @_user ||= User.find_or_create_by(github_id: auth_hash[:uid])
+    @_user ||= FindOrRegisterUser.new(github_id: auth_hash[:uid]).perform
```
jessieay: in this class we are also calling `from_oauth_hash(auth_hash)`, which sets attributes and creates the user. maybe move that logic into this new class?

app/services/find_or_register_user.rb
```diff
@@ -0,0 +1,27 @@
...
+      @_user = register_user
+    end
+
+    @_user
+  end
+
+  private
+
+  def find_user
+    User.where(github_id: github_id).first
```
jessieay: there will always be a user because `from_oauth_hash(auth_hash)` creates a user from the oauth hash we get from github. https://github.com/18F/micropurchase/blob/e3e02f33c326142f2689e52282fb5c9f0e1f9a2b/app/models/user.rb#L14

move oauth logic into here?

app/services/find_or_register_user.rb
adelevie: I initially had this called `CreateUser` when I threw this together. I changed it to `RegisterUser` since Registration seems more in line with the user-facing behavior. `Create` to me is a specific operation on the database (it inserts a row in the users table). `Register` implies creation plus the other things we need to happen when a new user joins the platform (check SAM, maybe send a welcome email, etc). Still sort of agnostic here, but those were my reasons.

jessieay:
I think extracting this is a great idea! Nice find on unused class, too. 

Left a few comments on naming / moving around some of the existing oauth logic

app/services/register_user.rb
jessieay: k cool, so let's memoize this (change to `@_user ||= User.create(github_id: github_id)`)

app/services/find_or_register_user.rb
jessieay: `RegisterUser` works for me -- that encapsulates the idea of finding or creating. Probably a smell to have the word "or" in a class name anyway since that indicates breakage of Single Responsibility Principle

app/models/login_user.rb
harrisj: 👍 on that suggestion

app/services/register_user.rb
harrisj: Any opportunity to use the word "idempotent" is a good one

spec/services/find_or_register_user.rb
```diff
@@ -0,0 +1,44 @@
+require 'rails_helper'
```
harrisj: So, did you rename FindOrRegisterUser to RegisterUser? Seems like these two files needs to be deleted

adelevie:
rebased

app/models/sam_account_reckoner.rb
```diff
@@ -31,4 +45,16 @@ def sam_status
   def client
     @client ||= Samwise::Client.new(api_key: DataDotGovCredentials.api_key)
   end
+
+  def vendor_summary
+    @vendor_summary ||= client.get_vendor_summary(duns: user.duns_number)
+  end
+
+  def duns_is_in_sam?
+    vendor_summary[:in_sam]
```
jessieay: assume this json returns a string val -- maybe we should check for equality here just in case?

eg: if it were `vendor_summary[:in_sam] == 'false' the current logic would return `true` because a string is "truthy"

app/models/sam_account_reckoner.rb
```diff
@@ -31,4 +45,16 @@ def sam_status
...
+  end
+
+  def duns_is_small_business?
+    vendor_summary[:small_business]
```
jessieay: same as above

app/services/find_or_register_user.rb
```diff
@@ -0,0 +1,27 @@
+class FindOrRegisterUser
+  attr_reader :github_id
+
+  def initialize(params)
+    @github_id = params[:github_id]
+  end
+
+  def perform
+    user = found_user = find_user
```
jessieay: woah what is going on here?

app/models/login_user.rb
jessieay: we are still doing a `create` at https://github.com/18F/micropurchase/blob/0f1958ca89b8778516b6b105f904cd9ddf5b6f77/app/models/user.rb#L14

let's move that oauth logic into `FindOrRegisterUser`

app/services/find_or_register_user.rb
```diff
@@ -0,0 +1,27 @@
...
+
+    if found_user.nil?
+      user = register_user
```
jessieay: we would never hit this case right now because we create the user at https://github.com/18F/micropurchase/blob/0f1958ca89b8778516b6b105f904cd9ddf5b6f77/app/models/user.rb#L14 -- let's bring that logic into here rather than creating and then registering

spec/support/fake_sam_api.rb
```diff
@@ -3,11 +3,20 @@
...
   VALID_DUNS = '0123456780000'
   INVALID_DUNS = '0876543210000'
+  SMALL_BUSINESS_DUNS = '0123456780001'
+  BIG_BUSINESS_DUNS = '0123456780002'
 
   get '/sam/v4/registrations/:duns' do
-    if params[:duns] == VALID_DUNS
-      json_response 200, 'duns_info.json'
-    elsif params[:duns] == INVALID_DUNS
+    case params[:duns]
```
jessieay: look at all of our hard work paying off! ✨

jessieay:
CI failure :(

Key piece of feedback here is that we are still doing double duty on creating a user -- we create the user from oauth and then hit sam API and update again. We should either create once from oauth and then update from samwise OR do both at the same time. 

Would guess that hitting sam api async would be wise. If that is the plan, then we can write this code assuming there will *always* be a user for the github id because user records are created when someone logs in with github

app/models/sam_account_reckoner.rb
adelevie: Samwise::Client#get_vendor_summary returns a plain Ruby hash (https://github.com/18F/samwise/blob/master/lib/samwise/client.rb#L17-L24) so I think this is fine.

jessieay: thoughts on checking that it is `== true` then, just to be safe?

adelevie:
> Key piece of feedback here is that we are still doing double duty on creating a user -- we create the user from oauth and then hit sam API and update again. We should either create once from oauth and then update from samwise OR do both at the same time.

Still sort of wrapping my head around this ... I'll think harder and look closer at that part of the codebase.

> Would guess that hitting sam api async would be wise. If that is the plan, then we can write this code assuming there will always be a user for the github id because user records are created when someone logs in with github

I _think_ the SAM API is being hit async through a Delayed::Job. `UsersController#update` calls `UpdateUser#save` which calls `SamAccountReckoner#delay` (and then #set).

Issue #549, opened 2016-05-13
Validation on auction publish


mtorres253:
As an admin user, I want to be prevented from creating auctions that don't have all required information, so the system should tell me if required information is missing so that I can provide it

Scenario Outline: User tries to PUBLISH an auction but leaves a mandatory field blank
  Given I am an authenticated admin user
  And I am on the create auction page
  And I have selected "Publish"
  And I leave the [field] blank 
  When I click "Submit"
  Then I should see [message] 
  And the auction should not be published

| field | message  |
|---|---|
| Title | A title is required |
| Summary | An auction summary is required |
| Description | An auction description is required |
| Start Date  | The start date and time for the auction is required |
| End Date  | The end date and time for the auction is required  |
| Delivery Deadline  | The delivery date and time for the auction is required  |
| Start Price  | A starting price for the auction is required  |

    
Note: just start with one flash message containing all the errors and go from there.

mtorres253:
K updated @jessieay

jessieay:
PR for validations: https://github.com/18F/micropurchase/pull/589

@mtorres253 should I add helper text as part of this story or do you want to define that separately?

mtorres253:
I'll define it separately. Thanks @jessieay

jessieay:
oh shoot @mtorres253 I was reading this as "on auction create" not publish. Do you only want these validations to happen when someone is setting an auction to "published" ?

PR #589, opened 2016-05-18
Add form validations to Admin auctions page

jessieay:
* Validate presence of various fields
* Also add DB constraints
* Show errors in one flash if wrong
* Render page to inputs not lost if form errors
* Add selectors for datetime form stuff

* Closes https://github.com/18F/micropurchase/issues/549

Future work:
* Add helper text to form to explain field meanings
* Add client-side validation
* Make date / timepickers fancier

screen shot 2016-05-18 at 10 14 06 am


Gemfile
```diff
@@ -33,7 +33,6 @@ group :test do
   gem 'webmock'
   gem 'json-schema'
   gem 'db-query-matchers'
-  gem 'json-schema'
```
jessieay: duplicate

app/models/auction.rb
```diff
@@ -1,4 +1,8 @@
 class Auction < ActiveRecord::Base
+  MAX_START_PRICE = 3500
```
adelevie: shouldn't we instead use `StartPriceThresholds::MICROPURCHASE`?

jessieay: oops merge conflict resolution error - will remove

jessieay: removed at 369eaab

app/models/auction.rb
```diff
@@ -10,8 +12,13 @@ class Auction < ActiveRecord::Base
   # Disable STI
   self.inheritance_column = :__disabled
 
+  validates :delivery_deadline, presence: true
```
harrisj: Are we validating here or only on publish?

app/models/auction_parser.rb
```diff
@@ -1,86 +1,58 @@
...
-
-  def general_attributes
     {
-      type: type,
-      title: title,
-      description: description,
-      summary: summary,
-      github_repo: github_repo,
-      issue_url: issue_url,
+      awardee_paid_status: auction_params[:awardee_paid_status],
```
harrisj: I approve

config/locales/simple_form.en.yml
```diff
@@ -7,3 +7,18 @@ en:
       mark: '*'
     error_notification:
       default_message: "Please review the problems below:"
+    labels:
+      auction:
+        start_datetime: "Start date and time (in UTC)"
```
harrisj: :+1:

app/models/auction.rb
jessieay: the issue requested it be validated on create: https://github.com/18F/micropurchase/issues/549

do you think we should only validate on publish?

app/models/auction_parser.rb
jessieay: ✔️

jessieay:
Updated to only validate some fields on publish. Still want to validate title and a few others on create. :)

db/migrate/20160517224513_add_null_false_to_auction_fields.rb
```diff
@@ -0,0 +1,32 @@
+class AddNullFalseToAuctionFields < ActiveRecord::Migration
+  def up
+    change_column :auctions, :billable_to, :string, default: ''
```
jessieay: explanation of this: when fields are able to be null, I prefer to set the default to empty string at the DB level. That way, no `nil` errors when we render the obj in the view. :)

adelevie:
Where'd all the `let`s go?!! I miss them.

mtorres253:
closes #592 as well

PR #595, opened 2016-05-19
Update auction factory to reflect required values

jessieay:
* Factories records should all be valid
* Should also not inlclude *more* data than is required, bc then your
  tests do not reflect reality; they are a world where there are not nil
  values when in reality you might have nil values
* Also update JSON schema to reflect this reality


harrisj:
I'm cool with all this. Merging.

PR #598, opened 2016-05-20
Rename datetime fields to use `at` suffix

jessieay:
* Consistent with rails style
* Eg: `updated_at` and `created_at`
* This is pre-work for datepicker stuff


harrisj:
Looks good to me

PR #599, opened 2016-05-20
Remove unused data json file

jessieay:
* Why was it here? The world may never know


adelevie:
oooo .... it used to be here as a quick and dirty way to get @lauraGgit and/or @gemfarmer developing against the API until we got CORS enabled.

adelevie:
RIP unused data.json file

harrisj:
You never realize how much you'll miss something until it's gone.

Issue #560, opened 2016-05-16
Date picker for create auction with defaults


mtorres253:
```gherkin
Feature: As an admin user, I want to be able to easily specify a start date/time with the minimum amount of manual work, so I should see defaults for the start date/time, and I should be able to change the date and time independently

Context: 
Given I am logged in as an admin
When I am creating a new auction (https://micropurchase.18f.gov/admin/auctions/new)

Scenario: Admin sees date and time inputs separately
Then I should see a section labeled <label>
And the date input should default to <date default> (mm/dd/yyyy)

Scenario: Admin sees a separate input for the hour bidding starts
Then I should see a dropdown for hours
And the hours dropdown should have the following options (01 - 12)
And the hours dropdown should default to <hour default>

Scenario: Admin sees a separate input for the minute bidding starts
Then I should see a dropdown for minutes
And the hours dropdown should have the following options (00 - 60)
And the hours dropdown should default to <minute default>

Scenario: Admin sees a separate input for "AM/PM"
Then I should see a dropdown with options (AM, PM)
And PM should be selected by default

Scenario: Admin sees time in ET
Then I should see ET next to the time inputs (HH, MM, AM/PM)

```

| label | date default | hour default | minute default |
| ---- | ---- | ---- | ---- |
| Bidding Start Date | today | 01 | 00 |
| Bidding End Date | today | 01 | 00 |
| Delivery Deadline | today | 01 | 00 |

PR #601, opened 2016-05-23
Create datepicker for auction form (with defaults)

jessieay:
* Closes https://github.com/18F/micropurchase/issues/560
* Separate date and time inputs for datetie fields
* Use HTML input for date
* Use custom selectors for time
* Defaults to 7pm UTC
* Next step is to make these ET and convert them in parser

screen shot 2016-05-23 at 11 51 44 am

screen shot 2016-05-23 at 11 51 48 am

### DIFF
--- a/spec/models/auction_parser_spec.rb
+++ b/spec/models/auction_parser_spec.rb
@@ -29,17 +29,16 @@ RSpec.describe AuctionParser do
         user = create(:user)
         params = {
           auction:  {
-            'started_at(1i)' => '2016',
-            'started_at(2i)' => '04',
-            'started_at(3i)' => '15',
-            'started_at(4i)' => '13',
-            'started_at(5i)' => '00',
+            'started_at' => '2016-04-15',
+            'started_at(1i)' => '13',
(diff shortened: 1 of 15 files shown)


jessieay:
code climate errors are due to long controller methods caused by inlining error msgs -- vote we ignore because this will all be solved by separating out API controllers

spec/models/auction_parser_spec.rb
```diff
@@ -29,17 +29,16 @@
...
+            'started_at(1i)' => '13',
+            'started_at(2i)' => '00',
+            'started_at(3i)' => 'PM',
           }
         }
 
         attributes = AuctionParser.new(params, user).attributes
 
-        expect(attributes[:started_at]).to be_a(ActiveSupport::TimeWithZone)
+        expect(attributes[:started_at]).to be_a(DateTime)
```
adelevie: RIP `ActiveSupport::TimeWithZone`

adelevie:
🎷 🐲 💯

PR #603, opened 2016-05-24
Kicking off the great view model refactoring tour of 2016

jessieay:
* We have a strategy, it's called doing stuff.
* Horrible horrible stuff
* Moving to a world where we stick all of the relevant logic that we
  need into a view model
* Starting with Auctions#index
* Thanks for kane for pairing with me for 3 straight hours on this


jessieay:
cc @baccigalupi

features/step_definitions/debugging_steps.rb
```diff
@@ -0,0 +1,3 @@
+When(/^Let me see that$/) do
```
adelevie: :100:

harrisj:
Can we disable the Too Many Lines style checking for the immediate future?

app/controllers/auctions_controller.rb
```diff
@@ -1,13 +1,13 @@
 class AuctionsController < ApplicationController
   def index
     collection = AuctionQuery.new.public_index
-    @view_model = AuctionsIndexViewModel.new(current_user, collection)
+    @auctions = AuctionsIndexViewModel.new(auctions: collection, current_user: current_user)
```
harrisj: I'm tempted to name the parameter in the model just `user` and we pass it `current_user` from the controller

app/presenters/human_time.rb
```diff
@@ -0,0 +1,23 @@
+class HumanTime
+  include ActionView::Helpers::DateHelper
```
harrisj: Oh I like this

app/view_models/auction_list_item.rb
```diff
@@ -0,0 +1,193 @@
+class AuctionListItem
```
harrisj: Excellent

app/view_models/auctions_index_view_model.rb
```diff
@@ -1,29 +1,30 @@
-class AuctionsIndexViewModel < Struct.new(:current_user, :auctions_query)
-  def auctions
-    @auctions ||= auctions_query.map { |auction| AuctionViewModel.new(current_user, auction) }
+class AuctionsIndexViewModel
+  attr_reader :auctions, :current_user
+
+  def initialize(auctions:, current_user:)
+    @auctions = auctions.to_a
```
harrisj: Is this because sometimes ActiveRecord will return a single item?

app/view_models/auctions_index_view_model.rb
```diff
@@ -1,29 +1,30 @@
...
+    @current_user = current_user
   end
 
   def active_auction_count
-    auctions.count { |i| i.started_at < Time.now && Time.now < i.ended_at }
+    auctions.count { |i| i.started_at < Time.current && Time.current < i.ended_at }
```
harrisj: I suppose eventually this would be replaced with an `auction.active?` in the block clause if we ever moved away from just using time conditions.

app/views/auctions/_bids_and_form.html.erb
```diff
@@ -15,3 +15,7 @@
     </span>
   </div>
 </div>
+
+<div class="issue-bids-info-item">
+  <%= render partial: auction.bid_button_partial, locals: { auction: auction.auction } %>
```
harrisj: This `auction.auction` looks a bit weird. Is this to get to an less-wrapped version of the auction object? Can we give that a more descriptive name?

app/views/auctions/_winning_bid.html.erb
```diff
@@ -1,20 +1,4 @@
...
-  <div class="issue-bids-info-item">No bids</div>
-<% end %>
-
-<% unless AuctionStatus.new(auction).over? or defined? for_winners_page %>
-  <span class="issue-bids-info-item bid-deadline">
-    Bid start time: <%= auction.relative_start_time %>
-</span>
-<% end %>
+<%= render partial: auction.winning_bid_partial, locals: { auction: auction } %>
+<%= render partial: auction.start_time_partial, locals: { auction: auction } %>
```
harrisj: Very nice

features/step_definitions/debugging_steps.rb
harrisj: Ha, I definitely can relate to this

app/view_models/auction_list_item.rb
gemfarmer: Agreed. This is so much better split up.

app/view_models/auction_list_item.rb
```diff
@@ -0,0 +1,193 @@
...
+    else # not for small business
+      true
+    end
+  end
+
+  def bid_button_partial
+    if current_user_can_bid?
+      'auctions/bid_button'
+    else
+      'components/null'
```
gemfarmer: Is this a convention to render an empty partial? It seems a little odd to me. Adding the conditional logic `current_user_can_bid?` in the template would take up more space and lead to some redundancy, but seems worth it for easier readability.

gemfarmer: If we think that we will be rendering null presenters all over the place, however, I definitely 👍 this approach.

app/view_models/auctions_index_view_model.rb
jessieay: this one took us a bit to figure out -- it's because the collection based on from the controller (`AuctionQuery.new.public_index`)  is not actually a collection, it's a query. And it doesn't fire until you map over it or call `to_a`

app/view_models/auctions_index_view_model.rb
jessieay: yeah we could totally add a scope for that. In face, we should. Trying to make as few changes as possible here so didn't change the implementation of the method. Do you think we should do that now or wait?

app/views/auctions/_bids_and_form.html.erb
jessieay: yes it does look weird! I delegated `id` in the view model so we can do named routes. will use `auction.id` instead

app/views/auctions/_winning_bid.html.erb
jessieay: yes! I think this polymorphic partial stuff is the one thing that actually looks better after this PR

app/view_models/auction_list_item.rb
jessieay: yeah we are trying to remove all conditionals from the view. 

The benefit of doing that is that it allows us to see in the ViewModel what the branches are. Out of those branches we will see patterns and be able to extract some more classes. 

If the conditionals are hidden in the views, it is harder to see patterns and therefore harder to extract objects

We aren't to the object extraction phase yet, but we're gonna get there!

app/controllers/auctions_controller.rb
jessieay: yes we talked about that! One benefit of having it be `current_user` is that the use case is super duper clear, but it's many more chars and I do love typing less. Hmm...

app/serializers/auction_serializer.rb
```diff
@@ -42,6 +43,14 @@ def winning_bid
 
   private
 
+  def veiled_bids
```
jessieay: note: had to add this because we used to pass in a presenter auction to this serializer but now we are just passing `auction`

app/view_models/auction_collection.rb
```diff
@@ -0,0 +1,29 @@
+class AuctionCollection < Struct.new(:current_user, :auctions_query)
```
jessieay: this is a dupe of the previous `AuctionsIndexViewModel`

app/view_models/auction_list_item.rb
jessieay: super long, but all the methods we need for `auctions#index` are here! (now more presenter or `AuctionViewModel` being used 💃 )

app/views/auctions/_bids_and_form.html.erb
jessieay: updated at be968ef

features/step_definitions/debugging_steps.rb
jessieay: :)

jessieay:
@harrisj re: turning off code climate. I don't mind having it on now if we are ok with failures. I'd kind of rather see the failures and merge anyway that not see them, but def open to other opinions on the matter!

app/controllers/auctions_controller.rb
harrisj: I guess we could make the theoretical case that you could see what another user sees even if the current_user is different (like an admin preview), but that is a bit of a stretch. I think the reason I like renaming it though is to distinguish between a `user` variable and a call to the `current_user` method in a controller

app/view_models/auctions_index_view_model.rb
harrisj: Oh that is right. I would recommend though that you perhaps move the `to_a` into the `AuctionQuery#public_index` method to make it explicitly run there, because we aren't really designing the AuctionQuery objects to be chained, are we? And this delayed execution can lead to confusing errors being thrown from views or such.

app/view_models/auctions_index_view_model.rb
harrisj: I think it's your call (I know the need to limit changes), but yeah I would love if we could abstract any downstream users from knowing what makes an auction active at some point down the road, even if they are all only time-based conditions for the immediate future

app/views/auctions/_bid_button.html.erb
```diff
@@ -1 +1,3 @@
-<a href="<%= new_auction_bid_path(auction) %>" class='usa-button usa-button-outline-inverse button-bid'>Bid <icon class="fa fa-angle-double-right"></icon></a>
+<%= link_to 'Bid <icon class="fa fa-angle-double-right"></icon>'.html_safe,
```
harrisj: I generally am in favor of using link_to, but this seemed like an edge case, and you have to do some unconventional stuff to make it work. I'm fine with this, but is there a good rule on when to use which?

harrisj:
Looks good to me. I'm happy if you want to merge.

app/view_models/auctions_index_view_model.rb
adelevie: They were designed to be chained, but I don't think public_index needs to be chainable. :thumbsup: to moving to_a  into public_index.

app/controllers/auctions_controller.rb
jessieay: ok cool -- how about tabling for now but as I move to next refactor I will change to `user` ?

app/views/auctions/_bid_button.html.erb
jessieay: I pretty much always use `link_to` -- adding classes doesn't seem too crazy. nor does the icon.  just did a quick SO post and it seems like this is the recommended way: http://stackoverflow.com/a/22358718/1019369

what about this feels edge casey?

app/view_models/auctions_index_view_model.rb
jessieay: so moving `to_a` into `public_index` breaks the unit tests for the query

convo at https://18f.slack.com/archives/micropurchase-dev/p1464120732004794

vote for tabling rather than falling down this rabbit hole atm

app/view_models/auctions_index_view_model.rb
jessieay: gonna create a card for it so I don't forget!

jessieay: done! xhttps://github.com/18F/micropurchase/issues/605

jessieay:
@harrisj yes please!

app/view_models/auctions_index_view_model.rb
jessieay: BOOM fixed at 36c01ba

Reason `to_a` was needed was that `count` does not work properly with more than one SQL condition http://stackoverflow.com/a/18949440/1019369

solution was to use AR query rather than select on an array

app/view_models/auctions_index_view_model.rb
```diff
@@ -2,16 +2,19 @@ class AuctionsIndexViewModel
...
   def initialize(auctions:, current_user:)
-    @auctions = auctions.to_a
+    @auctions = auctions
     @current_user = current_user
   end
 
   def active_auction_count
-    auctions.count { |i| i.started_at < Time.current && Time.current < i.ended_at }
+    auctions
+      .where('started_at < ?', Time.current)
```
harrisj: We'll probably move this back to an `select` or `inject` when it no longer becomes an easy thing to count. And/or we should move this method to `Presenter::Auction` but I don't want to bikeshed this any more right now.

harrisj:
Merging

PR #604, opened 2016-05-24
A first cut at reorganizing Cucumber steps

harrisj:
The goal is to group certain types of steps together by functionality rather than the features/scenarios they are used in. So, all steps for navigating to a page (or checking you are on one) are in the same place and so on. This will hopefully make it easier to see when a step like _show the auction start time_ is implemented independently in several different steps (as it seems to be)

This is still a work in progress, and I plan to clean up a few redundant steps and see if I can split some larger files a bit more cleanly as well.


features/step_definitions/admin_user_form_steps.rb
```diff
@@ -0,0 +1,26 @@
...
+  check('user_contracting_officer')
+end
+
+When(/^I submit the changes to the user$/) do
+  click_on "Update User"
+end
+
+Then(/^I expect there to be a contracting officer in the list of admin users$/) do
+  within(:css, "table#table-admins tbody tr td:nth-child(5)") do
+    expect(page).to have_content("true")
```
jessieay: this seems more like an `admin_user` step bc it doesn't relate to the form, does it?

features/step_definitions/user_profile_form_steps.rb
```diff
@@ -0,0 +1,133 @@
...
+  when 'duns_number'
+    Faker::Company.duns_number
+  when 'email'
+    "random#{rand(10000)}@example.com"
+  else
+    fail "Unknown attribute '#{attribute}'"
+  end
+end
+
+When(/^there is a (.+) associated with my account$/) do |attr|
```
jessieay: this doesn't seem like a form step, per se. Maybe we should have a file for steps related to user attributes?

features/step_definitions/user_dashboard_steps.rb
```diff
@@ -0,0 +1,7 @@
+Then(/^I expect the page to show me the number of regular users$/) do
```
jessieay: maybe `admin_user_dashboard_steps` ?

features/step_definitions/user_profile_form_steps.rb
jessieay: oh it looks like we *do* have a file like that! Tis step and the one above seem like they should do there?

features/step_definitions/page_steps.rb
```diff
@@ -1,113 +0,0 @@
-When(/^I refresh the page$/) do
```
jessieay: yay so glad to see this one go 👋

features/step_definitions/page_content_steps.rb
```diff
@@ -0,0 +1,53 @@
+Then(/^I should see "(.+)"$/) do |text|
```
jessieay: I wonder if we should separate these out into files related to specific pages -- as of right now, I am not sure when I would look in here vs other step files. we are checking for page content all over the place

features/step_definitions/meta_tag_steps.rb
```diff
@@ -0,0 +1,29 @@
...
+  expect(page).to have_css("meta[property='og:title'][content='18F Micro-purchase - #{pr_auction.title}']", visible: false)
+end
+
+# FIXME
+Then(/^there should be meta tags for the index page for (\d+) open and (\d+) future auctions$/) do |open_count, future_count|
+  expect(page).to have_css("title", visible: false, text: "18F - Micro-purchase")
+  expect(page).to have_css("meta[property='og:title'][content='18F - Micro-purchase']", visible: false)
+  expect(page).to have_css("meta[name='description'][content='The Micro-purchase Marketplace is the place to bid on open-source issues from the 18F team.']", visible: false)
+  expect(page).to have_css("meta[property='og:description'][content='The Micro-purchase Marketplace is the place to bid on open-source issues from the 18F team.']", visible: false)
+  expect(page).to have_css("meta[name='twitter:label1'][value='Active Auctions']", visible: false)
```
jessieay: I've been wondering about these meta tags! Why do we have them?

features/step_definitions/meta_tag_steps.rb
```diff
@@ -0,0 +1,29 @@
+Then(/^there should be meta tags for the edit profile form$/) do
+  expect(page).to have_css("title", visible: false, text: "18F Micro-purchase - Edit user: #{@user.name}")
+end
+
+Then(/^there should be meta tags for the closed auction$/) do
+  pr_auction = AuctionPresenter.new(@auction)
+
+  expect(page).to have_css("title", visible: false, text: "18F Micro-purchase - #{pr_auction.title}")
```
jessieay: this `title` is just delegated to auction -- thoughts on using `@auction.title` here?

features/step_definitions/admin_form_steps.rb
```diff
@@ -1,22 +1,11 @@
-When(/^I visit the auctions admin page$/) do
```
jessieay: should this file me named `admin_auction_form_steps`? Not sure but seems like this is all related to the auction form. worried that `admin_form` is too vague, although we only have one form right now . Or do we?

jessieay:
Few naming comments. This is a DEFINITE improvement 🎉 👏

features/step_definitions/page_steps.rb
harrisj: I just moved it to another file. Sorry!

features/step_definitions/admin_user_form_steps.rb
harrisj: Good point

features/step_definitions/meta_tag_steps.rb
harrisj: Ask @stvnrlly, but I think it was a fun feature for Slack and such, but I'd be fine if they went away.

features/step_definitions/page_content_steps.rb
harrisj: Let me think about this one a little. These are mostly looking for text (supplied as parameters or such), but there might be a better way to organize them.

jessieay:
@harrisj I'm ready to merge this when you are!

features/step_definitions/meta_tag_steps.rb
stvnrlly: If you're just wondering about the tests for the meta tags, it's useful to test so that we can check for information leakage through the tags.

As for the tags themselves: they should provide extra content around the link on Slack as well as Twitter, Facebook, Google, etc., which can make them more engaging, so I'd recommend keeping them in for that. However! They don't seem to be doing anything right now, so let me see if I can make them work properly first if you're thinking about removing them.

jessieay: > it's useful to test so that we can check for information leakage through the tags

yeah, as long as we have the tags I think it makes sense to test them. I was just wondering *why* we have them in the first place. The use case sounds somewhat compelling, but if these meta tags aren't working and we aren't going to prioritize fixing them then we should remove them

PR #606, opened 2016-05-24
Auction#show view model

jessieay:
* ALL THE METHODS WE NEED
* IN ONE PLACE


app/presenters/pluralize.rb
```diff
@@ -0,0 +1,14 @@
+class Pluralize
```
harrisj: Are we reinventing ActiveSupport's `Inflector` functionality here?

app/views/auctions/multi_bid/_auction_status.html.erb
```diff
@@ -1,8 +0,0 @@
-<span class="bid-amount"><%= @view_model.highlighted_bid_amount_as_currency %></span>
```
harrisj: Yay on this

app/presenters/pluralize.rb
jessieay: I don't think so, we're re-using ActionView's `pluralize` which I thin kis exactly what we want (pass in number and word and get pluralized or non-plurialized result depending) http://apidock.com/rails/ActionView/Helpers/TextHelper/pluralize

harrisj:
Merging, but I would like to do something about the complexity of some of these methods in a future PR

Issue #605, opened 2016-05-24
[Tech debt] Add query / scopes for auctions


jessieay:
See https://github.com/18F/micropurchase/pull/603/files/5873c836e07afea62b8226aed5376e3337350d85#r64381899

PR #608, opened 2016-05-24
Move `count` methods to `AuctionQuery`

jessieay:
* Closes https://github.com/18F/micropurchase/issues/605
* Also cleans up another view model woo hoo


app/controllers/auctions_controller.rb
```diff
@@ -48,28 +45,16 @@ def previous_winners_archive
...
-    @view_model = AuctionCollection.new(current_user, collection)
-
-    @auctions_json = @view_model.auctions.each { |a| AuctionSerializer.new(a, root: false) }.as_json
-
-    respond_to do |format|
-      format.html
-    end
-  end
-
-  rescue_from 'ActiveRecord::RecordNotFound' do
```
jessieay: removed this bc I wasn't sure what its purpose was -- why would we want this rather than a "record not found" error?

harrisj: I'd have to ask @adelevie as well. There must be a reason? I wonder if this was something to do with not revealing information about unpublished auctions or something else.

app/models/auction_query.rb
```diff
@@ -24,6 +24,19 @@ def initialize(relation = Auction.all)
     end
   end
 
+  def active_auction_count
```
harrisj: Excellent. This is so much better in this class

app/models/auction_query.rb
```diff
@@ -24,6 +24,19 @@ def initialize(relation = Auction.all)
...
+    public_index
```
harrisj: So we are able to chain methods in the auction query object as scopes?

harrisj:
I am good to merge this, but it looks like it needs to be rebased now

jessieay:
@harrisj rebased! Thanks for your many reviews 🎩

app/controllers/auctions_controller.rb
jessieay: @adelevie thoughts n this? I myself would default to `ActiveRecord::RecordNotFound` rather than making it a routing error since that is a more obvious error from the dev perspective (saves time debugging) but both would be 404 on prod..

harrisj:
Merging now on my mark. 3... 2... 1... go

PR #609, opened 2016-05-25
View model for Bids#new

jessieay:
* this is getting fun

### DIFF
--- a/app/models/highlighted_bid.rb
+++ b/app/models/highlighted_bid.rb
@@ -1,0 +1,42 @@
+class HighlightedBid
+  attr_reader :auction, :user
+
+  def initialize(auction:, user:)
+    @auction = auction
+    @user = user
+  end
+
+  def find
+    if !auction.bids.any?
+      NullBid.new
+    elsif available_single_bid? && user_is_bidder?
+      lowest_user_bid
+    elsif available_single_bid?
+      NullBid.new
+    else
+      auction.lowest_bid
+    end
+  end
+
+  private
+
+  def available_single_bid?
+    available? && auction.type == "single_bid"
+  end
+
+  def available?
+    AuctionStatus.new(auction).available?
+  end
+
+  def user_is_bidder?
+    user_bids.any?
+  end
+
+  def lowest_user_bid
+    user_bids.first
+  end
+
+  def user_bids
+    auction.bids.where(bidder: user).order(amount: :asc)
+  end
+end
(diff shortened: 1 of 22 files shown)


jessieay:
Code climate error is related to class being too long. Will be changed in future, promise!

spec/models/highlighted_bid_spec.rb
```diff
@@ -0,0 +1,69 @@
...
+        user = create(:user)
+        user_bid = create(:bid, auction: auction, bidder: user)
+
+        bid = HighlightedBid.new(auction: auction, user: user).find
+
+        expect(bid).to eq user_bid
+      end
+    end
+
+    context 'available single bid auction where is not bigger' do
```
jessieay: typo: should be bidder

app/view_models/bids_new_view_model.rb
```diff
@@ -0,0 +1,167 @@
+class BidsNewViewModel
```
jessieay: how are we feeling about this naming convention? I kind of hate it. I originally moved in this direction because I wanted to know what was a presenter and what was a view model without look at where they were..but in this new world of view models everywhere maybe concise names would be better.  eg: `BidsNew` or `NewBid` instead of `BidsNewViewModel` -- thoughts  @adelevie @harrisj ?

harrisj: Naming things is so hard. Let's just give everything random UUIDs as names. Maybe we should just do `NewBidViewModel`?

app/models/highlighted_bid.rb
```diff
@@ -0,0 +1,42 @@
+class HighlightedBid
```
harrisj: Okay, good

app/view_models/auction_show_view_model.rb
```diff
@@ -172,7 +172,7 @@ def user_is_bidder?
 
   def html_description
     return '' if auction.description.blank?
-    markdown.render(auction.description).html_safe
+    MarkdownRender.new(auction.description).to_s.html_safe
```
harrisj: Good plan, but I wonder if we should make a method on the MarkdownRender class that combines the `to_s.html_safe`

app/view_models/bids_new_view_model.rb
jessieay: sure, I like that (both suggestions)

jessieay: renamed at a51cba1

app/view_models/auction_show_view_model.rb
jessieay: good call, updated at 03a7186

harrisj:
Let's do this. Merging.

PR #615, opened 2016-05-26
Admin dashboard view models

jessieay:


app/view_models/admin/action_list_item.rb
```diff
@@ -0,0 +1,35 @@
+class Admin::ActionListItem
```
jessieay: should this be a `ActionItemListItem` or is that too terrible?

features/step_definitions/auction_create_steps.rb
```diff
@@ -1,5 +1,7 @@
 Given(/^there is an? (.+) auction$/) do |label|
   @auction = case label
+             when 'unpublished'
```
harrisj: Nice! I am considering actually splitting this method into a bunch of separate steps (since it's just a big case method anyway, but that doesn't hold up this PR)

harrisj:
It looks like we are getting close to the end.

PR #602, opened 2016-05-23
[WIP] Expiring auctions UI

gemfarmer:
⚠️ [WIP] ⚠️ – currently blocked by the viewmodel refactor

For #536

PR #616, opened 2016-05-27
Redesign for Auctions#index view

jessieay:
* Rebased / updated version of https://github.com/18F/micropurchase/pull/602


jessieay:
@gemfarmer I think there are a good number of code climate css violations here. I cleaned up the view logic so once  CC is taken care of this should be close to a merge!

harrisj:
Yeah, this branch will have to be rebased too it looks like, but if you can also look at the violations...

gemfarmer:
@jessieay Does this replace #602? I'm confused

jessieay:
Yes, it replaces it. I wanted to keep your branch intact in case I
overwrote something during rebase. :)

gemfarmer:
@jessieay gotcha!  I'm going to rebase and solve these scss issues now.

gemfarmer:
@jessieay This should be good to merge. _fingers crossed_ the checks pass!

app/view_models/auction_list_item.rb
```diff
@@ -43,92 +43,130 @@ def eligibility_label
...
-    end
+  def project_label
+    auction.github_repo.sub('https://github.com/', '')
   end
 
   def winning_bid_partial
+    if over? && !auction.bids.any?
+      'auctions/no_bids'
+    elsif auction.type == 'single_bid'
+      winning_single_bid_partial
```
jessieay: by going to these other methods we are making the logic a bit more complicated -- can we put it all pack into one method? I know it' SUPER long, but it makes it easier to understand the branching

gemfarmer: Definitely. I only did this because codeclimate was saying it was "too complex".

jessieay: if code climate was yelling at you for having a long conditional that is OK with me -- the point of getting this conditional on one place is to see all of the branching and extract new objects.

gemfarmer: I reverted it :shrug: It seems like a delicate balance. Oh the narrow lines we walk!

jessieay:
Sorry I made you fail code climate! I think this is ready to merge once we finish our small discussion about ribbons in the slack design channel

gemfarmer:
@jessieay ok new code pushed!

Issue #574, opened 2016-05-16
Prevent payment to vendor who is not a small business for auctions with start price > $3500


mtorres253:
As an admin I want to be prevented from paying a vendor via CAP integration for auctions with starting price > $3500 if that vendor is not a small business (This case could potentially arise if the vendor's status changes during the auction) So the system should prevent that from happening

Scenario: System prevents payment to vendor who is not a small business for auctions > $3500
Given that I am logged in as an admin
And there is an auction with a start price > $3500
And the contract was awarded to a vendor
And the vendor has delivered the work
And I have accepted the work via the UI https://micropurchase-staging.18f.gov/admin/auctions/[id]/edit (select accepted from result dropdown)
But Sam.gov tells me that the vendor is no longer a small business
Then I see an error message "This vendor is not a small business and is not qualified to receive the award"
And the payment request is not sent to CAP
And the auction result is rejected

Note: We may be overloading the "Accepted" field - is accepting the deliverable/paying the vendor two separate things? Let's see what happens

mtorres253:
Hi! Is this still in progress? It's been a while and I don't see any updates, so maybe it's already done? @adelevie

PR #617, opened 2016-05-27
When auction is accepted system checks Sam.gov to see if vendor is small business

### LINKED ISSUE #573: When auction is accepted system checks Sam.gov to see if vendor is small business
As an admin, I want to double check that a vendor is authorized to receive a payment before the payment is actually made, so if there is an auction with a starting price over $3500, before the payment is made through the system, the system should verify (again) that the winner is a small business

Scenario: System checks if vendor is small business when auction is accepted
And there is an auction with start price over $3500
And a vendor has won the auction
And the vendor has delivered acceptable work
When I accept the auction on https://micropurchase-staging.18f.gov/admin/auctions/[id]/edit (select accepted from result dropdown)
Then the system checks Sam.gov (again) to see if that vendor is still a small business

adelevie:
Fixes #573

### DIFF
--- a/app/models/auction_threshold.rb
+++ b/app/models/auction_threshold.rb
@@ -1,0 +1,16 @@
+class AuctionThreshold
+  MICROPURCHASE = 3500
+  SAT = 150000
+
+  def initialize(auction)
+    @auction = auction
+  end
+
+  def small_business?
+    (MICROPURCHASE + 1..SAT).cover?(auction.start_price)
+  end
+
+  private
+
+  attr_reader :auction
+end
--- a/app/models/rules/base_rules.rb
+++ b/app/models/rules/base_rules.rb
@@ -15,8 +15,12 @@ class Rules::BaseRules
     "#{base_dir}/#{partial_prefix}/#{name}.html.erb"
   end
 
+  def user_is_eligible_to_bid?(user)
+    @eligibility.eligible?(user)
+  end
+
   def user_can_bid?(user)
-    auction_available? && user.present? && @eligibility.eligible?(user)
+    auction_available? && user.present? && user_is_eligible_to_bid?(user)
   end
 
   def partial_prefix
--- a/app/presenters/auction_presenter.rb
+++ b/app/presenters/auction_presenter.rb
@@ -67,7 +67,7 @@ class AuctionPresenter
 
   delegate(
     :small_business?,
-    to: :start_price_thresholds
+    to: :auction_threshold
   )
 
   def bids?
@@ -171,8 +171,8 @@ class AuctionPresenter
     AuctionStatus.new(auction)
   end
 
-  def start_price_thresholds
-    @start_price_thresholds ||= StartPriceThresholds.new(start_price)
+  def auction_threshold
+    @_auction_threshold ||= AuctionThreshold.new(auction)
   end
 
   def decorated_bid(bid)
--- a/app/services/update_auction.rb
+++ b/app/services/update_auction.rb
@@ -10,10 +10,43 @@ class UpdateAuction < Struct.new(:auction, :params, :user)
   end
 
   def should_create_cap_proposal?
-    attributes[:result] == 'accepted' && auction.cap_proposal_url.blank?
+    auction_accepted_and_cap_proposal_is_blank? &&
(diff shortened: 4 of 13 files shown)


app/services/update_auction.rb
```diff
@@ -10,11 +10,31 @@ def perform
...
   end
 
   private
 
+  def auction_accepted_and_cap_proposal_is_blank?
+    attributes[:result] == 'accepted' && auction.cap_proposal_url.blank?
+  end
+
+  def winning_bidder_is_eligible_to_be_paid?
+    auction_presenter = AuctionPresenter.new(auction)
```
jessieay: thoughts on using `auction` here rather than `AuctionPresenter` ? I ask because I am going to kill `AuctionPresenter` some time today... :)

app/services/update_auction.rb
```diff
@@ -10,11 +10,31 @@ def perform
...
+    if auction_presenter.small_business?
+      winning_bidder = auction_presenter.winning_bid.bidder
+
+      reckoner = SamAccountReckoner.new(winning_bidder)
+      reckoner.set!
```
jessieay: should we be calling this in a delayed fashion since it hits the API?

app/services/update_auction.rb
```diff
@@ -10,11 +10,31 @@ def perform
...
+      winning_bidder.reload
+
+      auction_presenter.user_is_eligible_to_bid?(winning_bidder)
```
jessieay: if this returns `false`, should we notify the user somehow? Flash message?

adelevie: I think it should! The UI notification is part of this story https://github.com/18F/micropurchase/issues/574

app/services/update_auction.rb
adelevie: I'm fine with that, but I need access to `winning_bid`/winning bidder. What would be a good way to get that without `AuctionPresenter`?

jessieay: `WinningBid.new(auction).find` should do the trick!

app/services/update_auction.rb
jessieay: ok cool!

app/services/update_auction.rb
adelevie: I guess we should. This would mean `UpdateAuction#should_create_cap_proposal` will be need to be async.

app/services/update_auction.rb
adelevie: whoah!

app/services/update_auction.rb
jessieay: if it needs to happen synchronously that is fine, too. your call

adelevie: *if* we went the async route, would it make sense to stick the  `winning_bidder_is_eligible_to_be_paid?` inside of `CreateCapProposal`/'Job' (and maybe rename it)?

jessieay: sure that works for me!

app/models/auction_threshold.rb
```diff
@@ -0,0 +1,16 @@
+class AuctionThreshold
+  MICROPURCHASE = 3500
+  SAT = 150000
+
+  def initialize(auction)
+    @auction = auction
+  end
+
+  def small_business?
```
jessieay: 😢  (of joy)

app/presenters/auction_presenter.rb
```diff
@@ -60,14 +60,15 @@ def initialize(auction)
     :partial_path,
     :show_bids?,
     :user_can_bid?,
+    :user_is_eligible_to_bid?,
```
jessieay: I think we can remove all changes to this class? (not used, right?)

app/models/rules/base_rules.rb
```diff
@@ -15,8 +15,12 @@ def partial_path(name, base_dir = 'auctions')
     "#{base_dir}/#{partial_prefix}/#{name}.html.erb"
   end
 
+  def user_is_eligible_to_bid?(user)
```
jessieay: nice method name!

app/presenters/auction_presenter.rb
adelevie: lemme delete and run the tests to quickly check!

PR #619, opened 2016-05-27
Do not require factory girl in all envs

jessieay:
* Deploy was breaking because factories use Faker, which is not included
  in production
* We don't really need FG in staging / prod anyway, so let's not load
  factories in this envs

* Backtrace:

```
NameError: uninitialized constant Faker
       /tmp/staged/app/spec/factories/auctions.rb:10:in `block (2
levels) in <top (required)>'
       /tmp/staged/app/vendor/bundle/ruby/2.3.0/gems/factory_girl-4.7.0/lib/factory_girl/syntax/default.rb:18:in
`instance_eval'
       /tmp/staged/app/vendor/bundle/ruby/2.3.0/gems/factory_girl-4.7.0/lib/factory_girl/syntax/default.rb:18:in
`factory'
       /tmp/staged/app/spec/factories/auctions.rb:2:in `block in <top
(required)>'
```


adelevie:
🌴

PR #620, opened 2016-05-27
Add default values to `Admin::EditAuctionViewModel`

jessieay:
* Started_at and ended_at are required, but `delivery_deadline` input
  (which is on create only) is not, leaving us with a blank
  `delivery_due_at` field on edit
* Was receiving errors on staging (`undefined method strftime for nil`)
* Also update `auctions#new` defaults to be ET defaults suggested in #560

### DIFF
--- a/app/view_models/admin/edit_auction_view_model.rb
+++ b/app/view_models/admin/edit_auction_view_model.rb
@@ -14,24 +14,44 @@ class Admin::EditAuctionViewModel
   end
 
   def hour_default(field)
-    dc_time(field).strftime("%l").strip
+    if field_value(field).present?
+      dc_time(field).strftime("%l").strip
+    else
+      "1"
+    end
   end
 
   def minute_default(field)
-    dc_time(field).strftime("%M").strip
+    if field_value(field).present?
+      dc_time(field).strftime("%M").strip
+    else
+      "00"
+    end
   end
 
   def meridiem_default(field)
-    dc_time(field).strftime("%p")
+    if field_value(field).present?
+      dc_time(field).strftime("%p")
+    else
+      "PM"
+    end
   end
 
   def date_default(field)
-    dc_time(field).to_date
+    if field_value(field).present?
+      dc_time(field).to_date
+    else
+      DcTimePresenter.convert(Date.today).to_date
+    end
   end
 
   private
 
   def dc_time(field)
     DcTimePresenter.convert(auction.send("#{field}_at"))
+  end
+
+  def field_value(field)
+    auction.send("#{field}_at")
   end
 end
--- a/app/view_models/admin/edit_auction_view_model.rb
+++ b/app/view_models/admin/edit_auction_view_model.rb
@@ -13,45 +13,33 @@ class Admin::EditAuctionViewModel
     false
   end
 
+  def date_default(field)
+    dc_time(field).to_date
+  end
+
   def hour_default(field)
-    if field_value(field).present?
-      dc_time(field).strftime("%l").strip
-    else
-      "1"
-    end
+    dc_time(field).strftime('%l').strip
   end
 
   def minute_default(field)
-    if field_value(field).present?
-      dc_time(field).strftime("%M").strip
-    else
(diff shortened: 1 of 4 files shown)


app/view_models/admin/edit_auction_view_model.rb
```diff
@@ -14,24 +14,44 @@ def new_record?
   end
 
   def hour_default(field)
-    dc_time(field).strftime("%l").strip
+    if field_value(field).present?
+      dc_time(field).strftime("%l").strip
+    else
+      "1"
```
harrisj: I almost wonder if it makes sense to just call the NewAuctionViewModel for these, but I suppose a little duplication is fine too

app/view_models/admin/edit_auction_view_model.rb
```diff
@@ -14,24 +14,44 @@ def new_record?
...
+      dc_time(field).to_date
+    else
+      DcTimePresenter.convert(Date.today).to_date
+    end
   end
 
   private
 
   def dc_time(field)
     DcTimePresenter.convert(auction.send("#{field}_at"))
```
harrisj: Can you have this call field_value too?

app/view_models/admin/new_auction_view_model.rb
```diff
@@ -20,6 +20,6 @@ def meridiem_default(_field)
   end
 
   def date_default(_field)
-    Date.today
+    DcTimePresenter.convert(Date.today).to_date
```
harrisj: Good catch. This would be an issue late at night.

spec/view_models/admin/edit_auction_view_model_spec.rb
```diff
@@ -0,0 +1,93 @@
+require 'rails_helper'
+
+describe Admin::EditAuctionViewModel do
```
harrisj: Oh man, did we really have no tests? This is why I want to figure out the issues with coverage I guess!

app/view_models/admin/edit_auction_view_model.rb
jessieay: yeah let me think about how we can make this less ugly

app/view_models/admin/edit_auction_view_model.rb
jessieay: yeah good idea that will clean this up a bit

app/view_models/admin/edit_auction_view_model.rb
harrisj: I wouldn't sweat it too much though.

Sent from my iPhone

On May 27, 2016, at 6:20 PM, Jessie A. Young <notifications@github.com>
wrote:

In app/view_models/admin/edit_auction_view_model.rb
<https://github.com/18F/micropurchase/pull/620#discussion_r64973576>:

> @@ -14,24 +14,44 @@ def new_record?
>    end
>
>    def hour_default(field)
> -    dc_time(field).strftime("%l").strip
> +    if field_value(field).present?
> +      dc_time(field).strftime("%l").strip
> +    else
> +      "1"

yeah let me think about how we can make this less ugly

—
You are receiving this because you commented.
Reply to this email directly, view it on GitHub
<https://github.com/18F/micropurchase/pull/620/files/a0f13b309952c5e916ff6c0fa72877d75b258af9#r64973576>,
or mute the thread
<https://github.com/notifications/unsubscribe/AAAH_DC0bof-J5z-cT8wA5FY_J41Ymhwks5qF25HgaJpZM4Io3WW>
.

spec/view_models/admin/edit_auction_view_model_spec.rb
jessieay: these are tested at the integration level but we didn't have anything that hit these edge case so thought I'd throw in some unit tests!

jessieay:
@harrisj updates made at de9aefe6034a16a82e524f692da1b46b446a99e6 -- let me know what you think!

harrisj:
Looks good. Let's merge

PR #623, opened 2016-05-27
Modify new relic setup

jessieay:
* Was not logging
* Will try ENV vars instead via "user provided services"
* ENV vars override server-side config https://docs.newrelic.com/docs/agents/ruby-agent/installation-configuration/ruby-agent-configuration


jessieay:
Gonna self-merge so I can see if this works. Please forgive me teammates!

Issue #593, opened 2016-05-19
Auction form field placeholder text


mtorres253:
As an admin user, I want to make sure that I'm putting the right info in the right inputs when I create or edit an auction. So the system should offer some hints.

Scenario Outline: User tries to CREATE an auction and sees form input hints
  Given I am an authenticated admin user
  When I am on the create auction page
  Then I should see a placeholder `<message>` for each `<field>`

Scenario Outline: User tries to EDIT an auction and sees form input hints
  Given I am an authenticated admin user
  When I am on the create auction page
  Then I should see a placeholder `<message>` for each `<field>`

| field | message  |
|---|---|
| Title | A short description of the work |
| Summary  | A brief summary of the auction and rules |
| Description  | Add any context that might help the bidder, as well as language used, acceptance criteria and a definition of done |
| Start Price  | Specify the bidding start price |
| GitHub Repo  | The GitHub repo where the winner's pull request is to be filed |
| GitHub Issue  | The GitHub issue corresponding to this auction |
| Published | Selecting publish will make the auction visible to the public |
| Delivery URL | The URL of the pull request fulfilling the auction requirements |

harrisj:
Do we want to do these as hints underneath the form input or as text within the boxes? If the later, make sure they can't be published with that text.

jessieay:
When @mtorres253 and I discussed this, we were looking at `placeholder` style from simpleform

![screen shot 2016-05-19 at 10 54 55 am](https://cloud.githubusercontent.com/assets/601515/15549482/9297222a-2261-11e6-8a39-5796477f540e.png)

jessieay:
Closed via https://github.com/18F/micropurchase/pull/624

PR #624, opened 2016-05-31
Adding some hint text to the admin auction form

harrisj:
This turned out to be a lot simpler than I thought, although I would appreciate some guidance on the text copy. I also feel like these should be hints below rather than placeholder values within


jessieay:
@harrisj does this close https://github.com/18F/micropurchase/issues/593 ?

harrisj:
Yes, it's meant to close #593

PR #629, opened 2016-05-31
Create factory for creating auction status presenter

jessieay:
* Rather than a method in each view model
* Move auction status presenters to `presenters` dir


adelevie:
LGTM. @harrisj, wanna take a pass at this?

adelevie:
⛵

PR #630, opened 2016-05-31
Prevent payment to vendor 

### LINKED ISSUE #574: Prevent payment to vendor who is not a small business for auctions with start price > $3500

adelevie:
Fixes #574

### DIFF
--- a/features/step_definitions/auction_create_steps.rb
+++ b/features/step_definitions/auction_create_steps.rb
@@ -1,32 +1,13 @@
 Given(/^there is an? (.+) auction$/) do |label|
-  @auction = case label
-             when 'unpublished'
-               FactoryGirl.create(:auction, :unpublished)
-             when 'future'
-               FactoryGirl.create(:auction, :future)
-             when 'closed'
-               FactoryGirl.create(:auction, :closed, :with_bidders)
-             when 'closed bidless'
-               FactoryGirl.create(:auction, :closed)
-             when 'expiring'
-               FactoryGirl.create(:auction, :expiring, :with_bidders)
-             when 'open bidless'
-               FactoryGirl.create(:auction)
-             when 'open'
-               FactoryGirl.create(:auction, :with_bidders)
-             when 'single-bid'
-               FactoryGirl.create(:auction, :running, :single_bid)
-             when 'closed single-bid'
-               FactoryGirl.create(:auction, :closed, :with_bidders, :single_bid)
-             when 'needs evaluation'
-               FactoryGirl.create(:auction, :with_bidders, :evaluation_needed)
-             when 'within simplified acquisition threshold'
-               FactoryGirl.create(:auction, :between_micropurchase_and_sat_threshold, :available)
-             when 'below the micropurchase threshold'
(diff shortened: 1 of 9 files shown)


features/step_definitions/auction_create_steps.rb
```diff
@@ -1,32 +1,13 @@
...
-             when 'needs evaluation'
-               FactoryGirl.create(:auction, :with_bidders, :evaluation_needed)
-             when 'within simplified acquisition threshold'
-               FactoryGirl.create(:auction, :between_micropurchase_and_sat_threshold, :available)
-             when 'below the micropurchase threshold'
-               FactoryGirl.create(:auction, :below_micropurchase_threshold, :available)
-             else
-               fail "Unrecognized auction type: #{label}"
-             end
+  @auction = AuctionFactoryFactory.new(label).create
```
harrisj: I was actually going to take this long Cucumber method and turn it into a bunch of individual rules since I don't like features that are just big case statements

adelevie:
@jessieay @harrisj is there a clean way for `UpdateAuction` to add errors to the `Action` AR instance?

We're in this weird spot where I want there to be flash errors even though the update succeeds.

jessieay:
@adelevie I think I need to know more about what you're trying to accomplish here, but based on what I know so far I would say:

I would not expect an `UpdateAuction` to succeed if I am trying to pay someone who is not eligible. So I would expect to see the error flash message on a failure rather than on success.

Related product question: what do we do if the winner of an auction is not eligible? How is that handled?

harrisj:
We handle it by checking eligibility before they bid, right? I don't think we have any other way.

Note that I refactored that Cucumber file that you modified here so you will have to rebase. Sorry!

jessieay:
@harrisj I think this is related to an edge case where someone is eligible when they bid but they become uneligible later (business changes category in Sam.gov or something). So we need to check when their bid is accepted again.

jessieay:
Looking at the Code climate errors:

2 are easy to fix

Other one is related to the long controller method, which does seem kind of confusing. 

Why not prevent update from succeeding if the validation fails? eg: cannot set bid as accepted if a bidder is not eligible

jessieay:
Closing in favor of https://github.com/18F/micropurchase/pull/689

PR #632, opened 2016-05-31
Use status presenters in view models

jessieay:
* When a view model is checking auction status, use status presenter
  instead


app/view_models/auction_show_view_model.rb
```diff
@@ -254,11 +234,7 @@ def for_small_business?
   end
 
   def current_user_can_bid?
-    if auction.type == 'single_bid' && user_bids.any?
-      false
-    elsif over?
-      false
-    elsif future?
+    if !available? || auction.type == 'single_bid' && user_is_bidder?
```
adelevie: this might not be within scope of this PR, but this kind of logic I think should be completely outsourced to a `Rules` class.

adelevie:
One issue about ensuring bid-ability logic is in a Rules class and not duplicated elsewhere, otherwise: 🌊

app/view_models/auction_show_view_model.rb
jessieay: that's a good idea -- I will do that separately

PR #634, opened 2016-05-31
Use rules to determine if user can bid

jessieay:
* In UI, we want to show bid button to guests
* Othewise, we can use auction rules to determine if user can bid

### DIFF
--- a/app/view_models/auction_show_view_model.rb
+++ b/app/view_models/auction_show_view_model.rb
@@ -43,7 +43,7 @@ class AuctionShowViewModel
   end
 
   def eligibility_label
-    if for_small_business?
+    if AuctionThreshold.new(auction).small_business?
       'Small-business only'
     else
       'SAM.gov only'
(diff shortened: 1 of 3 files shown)


app/view_models/auction_show_view_model.rb
```diff
@@ -222,23 +222,11 @@ def status_presenter
...
-      false
-    elsif for_small_business? && current_user.small_business?
-      true
-    elsif for_small_business? && !current_user.small_business?
-      false
-    else # not for small business
+    if current_user.is_a?(Guest)
       true
+    else
+      RulesFactory.new(auction).create.user_can_bid?(current_user)
```
adelevie: memoize and stick in a method called `rules`? minor nit, ignore this if you want.

app/view_models/auction_show_view_model.rb
```diff
@@ -222,23 +222,11 @@ def status_presenter
...
-    elsif current_user.is_a?(Guest)
-      true
-    elsif !current_user.sam_accepted?
-      false
-    elsif for_small_business? && current_user.small_business?
-      true
-    elsif for_small_business? && !current_user.small_business?
-      false
-    else # not for small business
+    if current_user.is_a?(Guest)
```
adelevie: I wonder if this logic should be swept into the Rules class? Either that or rename `current_user_can_bid?` to `show_bid_button?` maybe?

app/view_models/auction_show_view_model.rb
jessieay: since we only hit this once, I don't think we need to memoize. Why do you think we should memoize?

app/view_models/auction_show_view_model.rb
jessieay: great idea -- renamed at ab626a0

app/view_models/auction_show_view_model.rb
adelevie: we also use it in `#show_bids` (line 161). even if we didn't memoize, it might be nice to not repeat `RulesFactory.new(auction).create` more than once in the class.

jessieay: you are right! totally agree. updated at e08a3dc

adelevie:
nice! merging when 🍏

PR #640, opened 2016-05-31
Polymorphize auction bid flashes

jessieay:
* Quack quack

Minor:

* Remove unused js and related specs


app/presenters/bid_status_flash_factory.rb
```diff
@@ -0,0 +1,175 @@
...
+  def lowest_user_bid
+    user_bids.order(amount: :asc).first
+  end
+
+  def user_bids
+    auction.bids.where(bidder: user)
+  end
+end
+
+class OverWithBids
```
jessieay: trying out putting super small classes all in one file when they are only serving one small function -- what do people think? I like not needing to open many files to see what's going on but I also find it weird to have many classes per file

adelevie:
Ready? Or still WIP?

app/presenters/bid_status_flash_factory.rb
harrisj: I don't think Rails' autoloader likes that :(

app/view_models/auction_show_view_model.rb
```diff
@@ -31,11 +31,7 @@ def rules_link_text
   end
 
   def rules_path
-    if auction.type == 'single_bid'
-      '/auction/rules/single-bid'
-    else
-      '/auctions/rules/multi-bid'
-    end
+    "/auctions/rules/#{auction.type.dasherize}"
```
harrisj: I would like to eventually rename single-bid and multi-bid to some better names

jessieay: totally!

jessieay:
@adelevie ready!

app/presenters/bid_status_flash_factory.rb
jessieay: it is working here since we are only referencing the classes within this file -- what do you think?

harrisj: Oh okay then

adelevie: I'm ok with multiple class files ... Maybe stuck into their own folder. We could also maybe start sticking README.md files in these folders to document/explain the API of some of the files in that folder.

adelevie:
LGTM

PR #641, opened 2016-06-01
Split Cucumber steps a bit further

harrisj:
I have decided against having a large switch statement in steps, so I have unrolled a few of these into separate step definitions

### DIFF
--- a/features/vendor_views_small_biz_auctions.feature
+++ b/features/vendor_views_small_biz_auctions.feature
@@ -12,7 +12,7 @@ Feature: Vendor views auctions set aside for small businesses only
     And I should see a "Small-business only" status
       
   Scenario: A vendor sees a non small business only auctions on the auction view page
-    Given there is a below the micropurchase threshold auction
+    Given there is an auction below the micropurchase threshold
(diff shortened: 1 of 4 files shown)


harrisj:
It looks like there is an issue with timezones. Investigating.

jessieay:
@harrisj looks like the same issue I ran into yesterday evening. If you rebase, should be fixed

features/vendor_views_small_biz_auctions.feature
```diff
@@ -12,7 +12,7 @@ Feature: Vendor views auctions set aside for small businesses only
     And I should see a "Small-business only" status
       
   Scenario: A vendor sees a non small business only auctions on the auction view page
-    Given there is a below the micropurchase threshold auction
+    Given there is an auction below the micropurchase threshold
```
jessieay: u write good

jessieay:
like this change a lot! Mind rebasing so we are sure the failure is the same as the one that is fixed on `develop` ?

harrisj:
Thanks, @jessieay. Rebased!

PR #645, opened 2016-06-01
Rescue errors on bid#confirm

jessieay:
* We rescue errors on #create but not #confirm, so things would just
  blow up on confirm if a bid was invalid
* Now we rescue error


harrisj:
Looks good to me. Let's ride!

PR #650, opened 2016-06-02
Rename AuctionStatus:: classes to StatusPresenter::

harrisj:
Since these are created by the StatusPresenterFactory, I feel like this is less confusing, especially since there is a separate AuctionStatus class.


jessieay:
❤️

PR #651, opened 2016-06-02
Merge develop into master

jessieay:
* deploy to prod!


jessieay:
This Hakiri warning seems like a false positive

adelevie:
> This Hakiri warning seems like a false positive

I agree.

PR #652, opened 2016-06-02
Add PlaceBidValidator class

jessieay:
* Encapsulate logic in one place
* Remove duplicate error messages `user_can_bid?` check for auction
  availability and bidder sam status

### DIFF
--- a/app/controllers/bids_controller.rb
+++ b/app/controllers/bids_controller.rb
@@ -23,20 +23,20 @@ class BidsController < ApplicationController
   end
 
   def confirm
-    bid = PlaceBid.new(params: params, user: current_user, via: via)
+    bid = PlaceBid.new(params: params, bidder: current_user, via: via)
     auction = Auction.find(params[:auction_id])
 
     if bid.valid?
       readonly_bid = bid.dry_run
       @confirm_bid = ConfirmBidViewModel.new(auction: auction, bid: readonly_bid)
     else
-      flash[:error] = bid.errors
+      flash[:error] = bid.errors.full_messages.to_sentence
       redirect_to new_auction_bid_path(auction)
     end
   end
 
   def create
-    @bid = PlaceBid.new(params: params, user: current_user, via: via)
+    @bid = PlaceBid.new(params: params, bidder: current_user, via: via)
 
     if @bid.perform
       respond_to do |format|
@@ -49,10 +49,12 @@ class BidsController < ApplicationController
     else
       respond_to do |format|
         format.html do
-          flash[:error] = @bid.errors
+          flash[:error] = @bid.errors.full_messages.to_sentence
           redirect_to new_auction_bid_path(params[:auction_id])
         end
-        format.json { render json: { error: @bid.errors }, status: 403 }
+        format.json do
+          render json: { error: @bid.errors.full_messages.to_sentence }, status: 403
+        end
       end
     end
   end
--- a/app/controllers/bids_controller.rb
+++ b/app/controllers/bids_controller.rb
@@ -23,51 +23,48 @@ class BidsController < ApplicationController
   end
 
   def confirm
-    bid = PlaceBid.new(params: params, user: current_user, via: via).dry_run
-    @confirm_bid = ConfirmBidViewModel.new(auction: Auction.find(params[:auction_id]), bid: bid)
-  rescue UnauthorizedError => error
-    flash[:error] = error.message
-    redirect_to new_auction_bid_path(params[:auction_id])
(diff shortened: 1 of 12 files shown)


spec/requests/bids_spec.rb
```diff
@@ -364,23 +372,6 @@
           end.to_not change { auction.bids.count }
         end
       end
-
-      context 'and the bid amount contains cents' do
-        let(:bid_amount) { 1.99 }
```
jessieay: the reason I deleted this test is that rails automatically rounds down the bid amount to the lower integer if you pass a float -- in that case, the person would see the rounded down float on the confirmation page, but not an error. is that a problem?

app/services/place_bid.rb
```diff
@@ -1,92 +1,46 @@
 class PlaceBid
-  BID_LIMIT = 3500
```
jessieay: 🔪

jessieay:
code climate error is that weird false positive thingy

app/controllers/bids_controller.rb
```diff
@@ -23,31 +23,40 @@ def new
...
-    @bid = PlaceBid.new(params: params, user: current_user, via: via).perform
+  def create
+    @bid = PlaceBid.new(params: params, bidder: current_user, via: via)
 
-    respond_to do |format|
-      format.html do
-        flash[:bid] = "success"
-        redirect_to auction_path(@bid.auction)
+    if @bid.perform
+      respond_to do |format|
```
harrisj: I guess this will be a bit cleaner when we move APIs to a different controller

app/services/place_bid.rb
```diff
@@ -1,94 +1,37 @@
...
+  include ActiveModel::Validations
+
   BID_INCREMENT = 1
 
-  attr_reader :bid, :params, :user, :via
+  attr_reader :params, :bidder, :via
+
+  validates_with PlaceBidValidator
```
harrisj: Ooooooh

spec/controllers/bids_controller_spec.rb
```diff
@@ -155,7 +155,7 @@
           expect(flash[:bid]).to eq("success")
           expect(response).to redirect_to("/auctions/#{auction.id}")
 
-          bid = auction.bids.order('created_at DESC').first
+          bid = auction.bids.last
```
harrisj: I actually would like to make this a method or something so that if bids is reordered later it won't break

app/validators/place_bid_validator.rb
```diff
@@ -0,0 +1,33 @@
+class PlaceBidValidator < ActiveModel::Validator
+  def validate(bid)
+    unless user_can_bid?(bid)
+      add_error(bid, 'You are not allowed to bid on this auction')
```
harrisj: Should we put these messages in the en.yml?

spec/requests/bids_spec.rb
harrisj: I think that is fine.

harrisj:
This looks good, but I had a few questions. Not sure if there should be another PR for that though

harrisj:
So, Brakeman is complaining in the CodeClimate that the partial rendering is taking a path that is derived from the `params` input and is thus dangerous. I don't really know how they reached that conclusion, but there are sometimes false positives. I would ask you to suppress the Brakeman warning for this before I can approve if it really is a false positive.

jessieay:
@harrisj we saw the same error on master -- not sure why it is popping up in unrelated PRs. will suppress

spec/controllers/bids_controller_spec.rb
jessieay: now or later? default ordering is by creation date and I am generally against default scopes / ordering so don't imagine it is likely to change, but it definitely could!

app/validators/place_bid_validator.rb
jessieay: love that idea!

jessieay: moved to locales at eaf91ce

jessieay:
@harrisj code climate is saying this is a common false positive:
> These warnings are often false positives, however, because it can be difficult to manipulate Rails' assumptions about paths to perform malicious behavior. Reports of dynamic render paths should be checked carefully to see if they can actually be manipulated maliciously by the user.

app/controllers/bids_controller.rb
jessieay: yes!

jessieay:
@harrisj I think I addressed all of your feedback. Thanks for the review! And for reviewing again :)

harrisj:
LGTM. Merging...

PR #653, opened 2016-06-02
CreateAuction should have a #perform method

harrisj:
Making it follow the same format as other service classes


jessieay:
code climate failure, while tiny, seems legit!

jessieay:
love this change!

harrisj:
Argh, fixed the CodeClimate issue

PR #659, opened 2016-06-02
Add Jasmine:CI to default rake task

jessieay:
* So our tests are run on Travis -- woot!

* Make jasmine specs pass
* Include helper file in spec where it is used
* Re-write function that was parsing github URL


gemfarmer:
@jessieay there are some tests commented out [here](https://github.com/18F/micropurchase/blob/5bac11da60d78ca2539578e2f1edcd553789a8d7/spec/javascripts/previous_winners_spec.js#L43)

I had commented out to isolate the error, but it looks like they are still hidden :(

gemfarmer:
I can't look right now, but I'm most of those will pass or will be similar errors. Interesting that they passed as they were in the browser, but not in the command line

gemfarmer:
@jessieay One other thing. [Th line where the mock data require was happening](https://github.com/18F/micropurchase/blob/93f25b6219f7196316ee0394a7f6b5d7ccc91cb8/spec/javascripts/previous_winners_spec.js#L1) was preventing the tests on that page from running. Fortunately, the tests pass in the browser AND command line when it is removed, so 🎉

jessieay:
@gemfarmer ok cool will uncomment them!

gemfarmer:
@jessieay I did with b248088, so should be good to squash/merge imo

jessieay:
yay thanks for the updates @gemfarmer !

PR #660, opened 2016-06-03
Update technical debt document

harrisj:
Revised to match the current file layout. Could probably use some more work, but it's getting there.

### DIFF
--- a/docs/technical_debt.md
+++ b/docs/technical_debt.md
@@ -368,6 +368,46 @@ Similarly, we have a few places where we branch depending on the
 auction type. This will likely make a good candidate for similar
 refactoring in the future.
 
+## Composition is Better than Inheritance
+
+The way we handle different rules for different types of auctions is a
+good example of how we favor composition/delegation over large
+inherited types. Currently we have two different types of auctions --
+a basic descending-value auction where all bid amounts are visible
+during the process and a sealed-bid auction where each user can only
+bid once and the winner is revealed when the auction is lower. We
+might add more, but these auctions have slightly different behavior
+from each other in a few different ways:
+
+1. Can the user bid on this auction?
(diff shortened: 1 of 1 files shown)


docs/technical_debt.md
```diff
@@ -14,8 +11,9 @@ branching nests of conditionals. We're trying a few techniques on this
 project to forestall that that could be summarized with the following principles:
 
 1. Long model classes are bad
-2. NullObject is better than nil
-3. View logic should not be in the templates
+2. Composition/delegation is better than inheritance
```
jessieay: we do have some inheritance still, eg: `BaseRules`. I think inheritance is fine as long as it is "shallow and wide"

docs/technical_debt.md
```diff
@@ -14,8 +11,9 @@ branching nests of conditionals. We're trying a few techniques on this
...
+3. NullObject is better than nil
```
jessieay: ❤️

jessieay:
This is great! 👍

docs/technical_debt.md
harrisj: True. I feel like maybe I should elaborate with the Rules object a little

harrisj:
Hold on, let me fix all the URL references

jessieay:
@harrisj let me know when this is ready for re-review! (you mentioned something about wanting to add docs on validator)

harrisj:
Needs to add in BidValidator as part of the user places bid walkthrough

harrisj:
Is there a reason this was assigned to me? It's ready for review

jessieay:
@harrisj cool, will re-review! I assigned you because I thought it wasn't ready - my bad!

Issue #582, opened 2016-05-17
Homepage clean-up


mtorres253:
Clean-up for Homepage
- [ ] Clean up welcome messaging on logged out view. Reduce vertical height/spacing but keep messaging as is. Try to fit on two lines.
- [ ] Flash messaging goes below
- [ ] Remove welcome section message from logged in view homepage
- [x] Replace "Issues open for bid" with "Auctions"
- [ ] Get rid of top and bottom spacing around the flash messaging on the homepage

jessieay:
https://github.com/18F/micropurchase/pull/674

PR #674, opened 2016-06-03
Home page updates

jessieay:
* Do not show welcome message to logged in user (Only for Guests)
* Update styles of welcome msg to make it more compact
* Remove upper margin from flash message

Closes https://github.com/18F/micropurchase/issues/582

Now and then (logged out):

screen shot 2016-06-03 at 3 35 51 pm

Now and then (logged in):

screen shot 2016-06-03 at 4 53 30 pm

### DIFF
--- a/app/controllers/auctions_controller.rb
+++ b/app/controllers/auctions_controller.rb
@@ -7,7 +7,7 @@ class AuctionsController < ApplicationController
 
     respond_to do |format|
       format.html do
-        sam_status_message_for(flash)
+        @auctions.sam_status_message_for(flash)
       end
       format.json do
         render json: published_auctions, each_serializer: AuctionSerializer
@@ -35,9 +35,5 @@ class AuctionsController < ApplicationController
 
   def published_auctions
     @_published_auctions ||= AuctionQuery.new.public_index
-  end
-
-  def sam_status_message_for(flash)
-    current_user.decorate.sam_status_message_for_auctions_index(flash)
   end
 end
--- a/app/controllers/auctions_controller.rb
+++ b/app/controllers/auctions_controller.rb
@@ -1,13 +1,13 @@
 class AuctionsController < ApplicationController
   def index
-    @auctions = AuctionsIndexViewModel.new(
+    @auction_collection = AuctionsIndexViewModel.new(
       auctions: published_auctions,
       current_user: current_user
     )
 
     respond_to do |format|
       format.html do
-        @auctions.sam_status_message_for(flash)
+        @auction_collection.sam_status_message_for(flash)
       end
       format.json do
         render json: published_auctions, each_serializer: AuctionSerializer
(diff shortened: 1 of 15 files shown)


app/presenters/guest_presenter.rb
```diff
@@ -18,4 +18,8 @@ def win_header_partial
   def nav_drawer_submenu_partial
     "components/guest_nav_drawer_submenu"
   end
+
+  def welcome_message_partial
```
harrisj: Not really for this PR, but I wonder if we should think about some other organizational layer that collects these partials in a single directory. I suppose this is what the ViewModels are supposed to do. Should we move GuestPresenter to a GuestViewModel in a future PR?

app/views/auctions/index.html.erb
```diff
@@ -5,15 +5,7 @@
...
-<section class="slab-intro usa-width-full">
-  <div class="usa-grid">
-    <h1 class="intro-text">Welcome!</h1>
-    <div class="usa-font-lead">
-      <p>The Micro-purchase Marketplace is the place to bid on open-source issues from the 18F team. <a class="link-highlighted" href="https://18f.gsa.gov/2015/11/06/micro-purchase-lessons/">Learn more about this experiment <icon class="fa fa-angle-double-right"></icon></a></p>
-      <p>Be sure to <a class="link-highlighted" href="<%= login_path %>">register your company</a> before placing a bid. If you have any questions, please email us at <a class="link-highlighted" href="mailto:micropurchase@gsa.gov">micropurchase@gsa.gov</a>.</p>
-    </div>
-  </div>
-</section>
+<%= render partial: @auctions.welcome_message_partial %>
```
harrisj: Always appreciate this.

app/controllers/auctions_controller.rb
```diff
@@ -7,7 +7,7 @@ def index
 
     respond_to do |format|
       format.html do
-        sam_status_message_for(flash)
+        @auctions.sam_status_message_for(flash)
```
harrisj: I will admit this confused me at first because I expected from Rails conventions that `@auctions` would be an array or something and not an instance of `AuctionsIndexViewModel`. I think it's fine for this PR, but do you think we should rename to `@auction_collection`? Does it cause any confusion in the relevant view to do so?

jessieay: yeah I agree with your confusion -- not a great name. renamed at 7f90d47

app/presenters/guest_presenter.rb
jessieay: Since these methods are not being called in one particular view but are "view-related" components, I think that putting them in a presenter makes sense. Basically, right now we have a 1:1 mapping of view models and controller methods. Presenters are decoration around that so they're a bit different. We can def discuss more, interested in why you think this should be a view model!

jessieay:
@harrisj this is ready for re-review

harrisj:
Thanks! Merging!

PR #675, opened 2016-06-04
Fix commenting of FAQs on winners page

jessieay:
* This is how you comment in erb


adelevie:
Agh, my bad for merging this in to begin with. Thanks for the fix.

PR #676, opened 2016-06-06
Move lib-based assets to `vendor` dir

jessieay:
* Also remove extra ref to `c3.css`
* Attempting to fix issue on prod:
  `404 not found for https://micropurchase.18f.gov/assets/source/c3.css`


harrisj:
Should be better for CodeClimate too

PR #677, opened 2016-06-06
Prevent "Unable to authenticate" errors

jessieay:
* Seeing a 5+ of these from this morning
* Request referrers mostly appear to be bots

screen shot 2016-06-06 at 11 04 20 am


adelevie:
The code looks good! Your call, but should there be a test for this? I can see why it might be overkill here.

jessieay:
@adelevie great call on testing, updated our existing specs at c7fdd7c which were previously not testing the right thing anyway!

harrisj:
Looks good. Merging

PR #678, opened 2016-06-06
WIP: Added a method for downloading the users as a CSV

### LINKED ISSUE #658: .csv export of vendor data
As an admin user, I will sometimes want to get an export of vendor data so I can do things like send bulk emails to the vendors. So the system should allow me to do this

Context:
Given I am a logged in admin user

Scenario: I see a link on the users page to dowload a .csv
When I am viewing the users page https://micropurchase.18f.gov/admin/users
Then I see a link labeled "Download CSV" next to the "Users" header

Scenario: I download a .csv containing users name and email
When I click on the "Download CSV" link
Then a CSV file should be downloaded to my computer
And the CSV column headers should be "First Name", "Last Name" and "Email Address"
And under each column header I should see the first name, last name and email address of each user

harrisj:
Fixes #658

### DIFF
--- a/app/controllers/admin/user_reports_controller.rb
+++ b/app/controllers/admin/user_reports_controller.rb
@@ -1,0 +1,16 @@
+class Admin::UserReportsController < ApplicationController
+  def index
+    @users = User.all
+
+    respond_to do |format|
+      format.csv do
+        response.headers['Content-Type'] = 'text/csv'
+        response.headers['Content-Disposition'] = "attachment"
+        send_data(
+          UserExport.new.export_csv,
+          filename: "micropurchase_users.csv"
+        )
+      end
+    end
+  end
+end
--- a/app/controllers/admin/user_reports_controller.rb
+++ b/app/controllers/admin/user_reports_controller.rb
@@ -1,4 +1,6 @@
 class Admin::UserReportsController < ApplicationController
+  before_filter :require_admin
+
   def index
     @users = User.all
 
--- a/spec/controllers/admin/user_reports_controller_spec.rb
+++ b/spec/controllers/admin/user_reports_controller_spec.rb
@@ -1,0 +1,17 @@
+require 'rails_helper'
+
+RSpec.describe Admin::UserReportsController do
+  describe '#index' do
+    context 'when called by an admin user' do
+      it 'should return a CSV' do
+        user = create(:admin_user)
+        10.times { create(:user) }
+
+        get :index, { format: 'csv' }, user_id: user.id
+
+        expect(response.code).to eq('200')
+        expect(response.header['Content-Type']).to include('text/csv')
+      end
+    end
+  end
+end
--- a/spec/controllers/admin/user_reports_controller_spec.rb
+++ b/spec/controllers/admin/user_reports_controller_spec.rb
@@ -15,13 +15,13 @@ RSpec.describe Admin::UserReportsController do
     end
 
     context 'when called by a non-admin user' do
-      it 'should return a 500 server error' do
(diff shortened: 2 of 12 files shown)


spec/controllers/admin/user_reports_controller_spec.rb
```diff
@@ -0,0 +1,17 @@
+require 'rails_helper'
+
+RSpec.describe Admin::UserReportsController do
+  describe '#index' do
+    context 'when called by an admin user' do
+      it 'should return a CSV' do
+        user = create(:admin_user)
+        10.times { create(:user) }
```
jessieay: do we need any users in the system in order for this csv export to work?

harrisj: It should work without I think

spec/factories/users.rb
```diff
@@ -6,6 +6,7 @@
     duns_number { Faker::Company.duns_number }
     name { Faker::Name.name }
     email
+    github_login 'ghlogin'
```
jessieay: maybe this should be a required field on `User` ?

app/controllers/admin/user_reports_controller.rb
```diff
@@ -0,0 +1,16 @@
+class Admin::UserReportsController < ApplicationController
+  def index
```
jessieay: do we want to ensure that only admins can access this endpoint?

spec/controllers/admin/user_reports_controller_spec.rb
jessieay: in that case, thoughts on not creating these records? would make tests faster ;)

jessieay:
Few small comments -- looking awesome!

jessieay:
looks like the story contains a requirement of a link to download this CSV -- are you planning on doing that separately?

harrisj:
Good point. I should add that

spec/controllers/admin/user_reports_controller_spec.rb
harrisj: I wanted to have at least 2 for kicks... but yeah

spec/factories/users.rb
harrisj: I actually need to check if we are fetching this anywhere or not

harrisj:
I added the link, but it could use some styling assistance.

spec/controllers/admin/user_reports_controller_spec.rb
jessieay: cool can we just create 2 additional users, then? fast test suite = more helpful test suite ;)

spec/controllers/admin/user_reports_controller_spec.rb
```diff
@@ -0,0 +1,28 @@
...
+
+        get :index, { format: 'csv' }, user_id: user.id
+
+        expect(response.code).to eq('200')
+        expect(response.header['Content-Type']).to include('text/csv')
+      end
+    end
+
+    context 'when called by a non-admin user' do
+      it 'should return a 302 redirection with an error' do
```
jessieay: niiiice

spec/factories/users.rb
jessieay: why did we add it here? If there is not a validation but we require it for something to work, we should add the validation

jessieay: fwiw, anyone signing in via github will have a github login so adding the validation shouldn't be a problem

harrisj: We wanted to grab it when names weren't being captured.

jessieay:
this looks perfecto -- merging!

PR #680, opened 2016-06-07
Fix winners page (I hope!)

jessieay:
* On staging, winners page works
* On prod, it is broken
* https://micropurchase.18f.gov/assets/source/c3.css does not exist
* but https://micropurchase-staging.18f.gov/assets/source/c3.css does
* We should not be directly referencing static assets -- hellooooo we
  are using rails and we <3 the asset pipeline
* Renaming `c3.css` to `c3.scss` makes it so we are putting it in asset
  pipeline
* I don't understand why this is the case, but confirmed behavior by
  compiling assets locally
* The other changes here are config cleanup


harrisj:
So, can you test this works okay by doing a manual cf push to staging?

jessieay:
@harrisj sure, I also tested that it works locally by using compiled assets locally

jessieay:
you still want me to test on staging before we merge?

jessieay:
pushed to staging and everything looks goood https://micropurchase-staging.18f.gov/auctions/winners

harrisj:
Rails' asset pipeline: "You have bested me, but I shall have my revenge!"

Issue #663, opened 2016-06-03
P-card payment method is specified for an auction


mtorres253:
As an admin user, I want to be able to prevent automatic CAP payment requests when the auction is being paid with another P-card, so I should be able to specify when automatic payments can be made

Scenario: Admin specifies whether an auction is paid through an 18F P-card
Given I am a logged in admin
When I am creating an auction
Then I should see a checkbox labeled "Use 18F P-card"
And that checkbox should be CHECKED by default
And I should be able to uncheck it
And I should be able to save my changes

jessieay:
moved back to "backlog" bc want to discuss naming before implementing 

https://18f.slack.com/archives/micropurchase-dev/p1465335446000118

jessieay:
in slack we settling on a `pcard` enum field 

![screen shot 2016-06-07 at 3 02 35 pm](https://cloud.githubusercontent.com/assets/601515/15876215/ebcfe4f8-2cc0-11e6-8d5b-dd22ec27e6aa.png)

jessieay:
@mtorres253 did this as as dropdown rather than a checkbox for reasons discussed in Slack

screenshot on PR https://github.com/18F/micropurchase/pull/681

PR #681, opened 2016-06-07
Add `purchase_card` field to Auctions

jessieay:
* Enum
* Defaults to "default" but can also be "other"
* Closes https://github.com/18F/micropurchase/issues/663

screen shot 2016-06-07 at 4 08 24 pm

### DIFF
--- a/app/models/auction.rb
+++ b/app/models/auction.rb
@@ -8,19 +8,21 @@ class Auction < ActiveRecord::Base
   enum type: { single_bid: 0, multi_bid: 1 }
   enum awardee_paid_status: { not_paid: 0, paid: 1 }
   enum published: { unpublished: 0, published: 1 }
+  enum purchase_card: { default: 0, other: 1 }
 
   # Disable STI
   self.inheritance_column = :__disabled
 
+  validate :start_price_equal_to_or_less_than_max_if_not_contracting_officer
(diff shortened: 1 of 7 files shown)


app/models/auction.rb
```diff
@@ -8,19 +8,21 @@ class Auction < ActiveRecord::Base
...
 
   # Disable STI
   self.inheritance_column = :__disabled
 
+  validate :start_price_equal_to_or_less_than_max_if_not_contracting_officer
+  validates :delivery_due_at, presence: true, if: :published?
+  validates :description, presence: true, if: :published?
   validates :ended_at, presence: true
-  validates :started_at, presence: true
+  validates :purchase_card, presence: true
```
harrisj: So, will someone have to select a value just to save the auction? Will this create a situation where someone might just pick one because they want to save it and it never gets reviewed later?

harrisj:
Looks good. Merging.

app/models/auction.rb
jessieay: the form has a drop down with two options -- defaults to "default". Since Michael originally wanted a boolean, I thought this made the most sense (because in a bool situation it would either be "18f pcard" or "not 18f pcard") but we can discuss more if you like!

jessieay: in short: yes, right now you must select a value for `purchase_card`

Issue #649, opened 2016-06-02
Charts aren't showing on the winner's page


mtorres253:
Go to https://micropurchase.18f.gov/auctions/winners

Expected behavior:
Charts appear on the page

Actual behavior:
They are no longer there

See screenshot
![screen shot 2016-06-02 at 11 23 42 am](https://cloud.githubusercontent.com/assets/100149/15756086/8204cc1a-28b4-11e6-95be-672779f3f959.png)

jessieay:
PR: https://github.com/18F/micropurchase/pull/680

PR #683, opened 2016-06-08
Asset debugging for production

jessieay:
* Reason it was working on staging and not production was that there was
  no config file for staging so assets were not being uglified like on prod
* Update env vars on staging so it uses same config file as prod
* For some reason, uglifying is causing javascript to break
* Until we get to the root of the issue, removing uglifier fixes problem
* To make sure assets are too insanely slow, use minified / CDN assets
  where possible

* Addresses https://github.com/18F/micropurchase/issues/649


jessieay:
pushed to staging manually to confirm that this works

app/assets/javascripts/application.js
```diff
@@ -1,12 +1,6 @@
...
- * = require utility.js
- * = require components.js
- * = require cookies.js
- * = require drawer.js
- * = require navDropdown.js
- */
-
-
+//= require jquery-accessibleMegaMenu.js
+//= require d3.v3.min
```
gemfarmer: This needs to be `require d3.v3.min.js`

spec/javascripts/support/jasmine.yml
```diff
@@ -13,6 +13,7 @@
 src_files:
   - assets/application.js
   - assets/winners-bundle.js
+  - assets/jquery.js
```
gemfarmer: jquery should probably go first on this list of assets, just in case it matters for jasmine's loading

vendor/assets/stylesheets/c3.min.scss
```diff
@@ -0,0 +1 @@
...-region{fill:#4682b4;fill-opacity:.1}.c3-brush .extent{fill-opacity:.1}.c3-legend-item{font-size:12px}.c3-legend-item-hidden{opacity:.15}.c3-legend-background{opacity:.75;fill:#fff;stroke:#d3d3d3;stroke-width:1}.c3-title{font:14px sans-serif}.c3-tooltip-container{z-index:10}.c3-tooltip{border-collapse:collapse;border-spacing:0;background-color:#fff;empty-cells:show;-webkit-box-shadow:7px 7px 12px -9px #777;-moz-box-shadow:7px 7px 12px -9px #777;box-shadow:7px 7px 12px -9px #777;opacity:.9}.c3-tooltip tr{border:1px solid #CCC}.c3-tooltip th{background-color:#aaa;font-size:14px;padding:2px 5px;text-align:left;color:#FFF}.c3-tooltip td{font-size:13px;padding:3px 6px;background-color:#fff;border-left:1px dotted #999}.c3-tooltip td>span{display:inline-block;width:10px;height:10px;margin-right:6px}.c3-tooltip td.value{text-align:right}.c3-area{stroke-width:0;opacity:.2}.c3-chart-arcs-title{dominant-baseline:middle;font-size:1.3em}.c3-chart-arcs .c3-chart-arcs-background{fill:#e0e0e0;stroke:none}.c3-chart-arcs .c3-chart-arcs-gauge-unit{fill:#000;font-size:16px}.c3-chart-arcs .c3-chart-arcs-gauge-max,.c3-chart-arcs .c3-chart-arcs-gauge-min{fill:#777}.c3-chart-arc .c3-gauge-value{fill:#000}
```
gemfarmer: this should be `c3.min.css` instead of `c3.min.scss`

app/assets/javascripts/winners-bundle.js
```diff
@@ -1,4 +1,5 @@
-//= require d3
-//= require winners/c3
-//= require winners/underscore
-//= require_tree ./winners/
```
gemfarmer: did `require_tree` not work? If it does, I think it would be cleaner

gemfarmer:
Sweet @jessieay!! I made a few comments, mostly to clean it up. It didn't look like it was working on staging, but I think that was mostly because of the d3 issue that I made a comment for. Hopefully it works!

app/assets/javascripts/application.js
jessieay: oh yes sorry staging is broken for other reasons (I was experimenting with some "conditional uglifying") this works as-is - will push to staging to show you!

vendor/assets/stylesheets/c3.min.scss
jessieay: we had some weird issues with it using `css` before --  see https://github.com/18F/micropurchase/pull/680

is there a problem with using `scss`?

app/assets/javascripts/application.js
gemfarmer: Ah ok. I was wondering if that was the case because you had said it was working. I figured you may have changed something after that comment

app/assets/javascripts/winners-bundle.js
jessieay: it did work, but I like having all of the files here explicitly. What do you think?

Gemfile
```diff
@@ -26,6 +26,8 @@ gem 'delayed_job_active_record'
 gem 'daemons'
 gem 'foreman'
 gem 'c2', github: '18f/c2-api-client-ruby'
+gem 'jquery-rails-cdn'
```
adelevie: Why do we want a CDN as opposed to loading JQuery locally?

manifest-staging.yml
```diff
@@ -14,5 +14,5 @@ applications:
     - data-dot-gov
   command: script/start
   env:
-    RAILS_ENV: staging
-    RACK_ENV: staging
+    RAILS_ENV: production
```
adelevie: 👏

adelevie:
LGTM ... I don't love our use of a CDN here, but it's not a huge deal.

🍧

Gemfile
jessieay: I think in general using a CDN is a lot of overhead for a site with our kind of traffic (aka - low) but in the case of jquery, since it's so easy, I think it is worth it. https://www.sitepoint.com/7-reasons-to-use-a-cdn/

> jQuery is ubiquitous on the web. There’s a high probability that someone visiting your pages has already visited a site using the Google CDN. Therefore, the file has already been cached by your browser and won’t need to be downloaded again.

jessieay:
@adelevie can remove CDN -- basically, since we are not uglifying our assets, I was trying to get them as speedy as possible. Minified jquery would be fine, maybe this was an overoptimization...

vendor/assets/stylesheets/c3.min.scss
gemfarmer: I don't think there would be any issues (that I'm aware of), it just isn't an accurate file extension. It is css, not scss

app/assets/javascripts/winners-bundle.js
gemfarmer: If both work, I would vote `require_tree` explicitly listing, them, but I don't have a strong opinion either way. Completely fine with this.

Gemfile
gemfarmer: Good point @jessieay . I would only vote local version jquery if the specific version was necessary, which I don't think is the case with MP.

gemfarmer:
CDN aside, what is the status on this? I see that it is broken on staging. I assume the latest here hasn't been deployed?

vendor/assets/stylesheets/c3.min.scss
jessieay: yeah it is weird. oh man I am so sick of this voo doo asset stuff!

jessieay:
@gemfarmer this is no longer on staging we have other stuff up there. We can merge when you're ready!

manifest-staging.yml
gemfarmer: 👏 👏

PR #684, opened 2016-06-09
First cut at pulling out APIs to different routes

### LINKED ISSUE #656: Refactor API into new controller
As discussed, we want to refactor the existing API so that we can eventually version it. This task is for the initial refactoring so that the API is in its own controller. For MVP, the same API with some capability to have it versioned

harrisj:
Fixes #656

This change only applies to public API methods. I am still letting the admin routes do the old way of things Note that this also has a Rails routing directive that takes old URL paths that end with .json and
redirects them to the new API.

### DIFF
--- a/app/controllers/api/v0/auctions_controller.rb
+++ b/app/controllers/api/v0/auctions_controller.rb
@@ -1,0 +1,9 @@
+class Api::V0::AuctionsController < ApplicationController
+  def index
+    render json: AuctionQuery.new.public_index, each_serializer: AuctionSerializer
+  end
+
+  def show
+    render json: AuctionQuery.new.public_find(params[:id]), serializer: AuctionSerializer
+  end
+end
--- a/app/controllers/api/v0/bids_controller.rb
+++ b/app/controllers/api/v0/bids_controller.rb
@@ -1,0 +1,27 @@
+class Api::V0::BidsController < ApplicationController
+  before_filter :require_authentication, except: [:index]
+  skip_before_action :verify_authenticity_token
+
+  def index
+    auction = AuctionQuery.new.bids_index(params[:auction_id])
+    @auction_bids = BidsIndexViewModel.new(auction: auction, current_user: current_user)
+  end
+
+  def create
+    @bid = PlaceBid.new(params: params, bidder: current_user, via: via)
+
+    if @bid.perform
+      render json: @bid.bid, serializer: BidSerializer
+    else
+      render json: { error: @bid.errors.full_messages.to_sentence }, status: 403
+    end
+  end
+
+  rescue_from 'ActiveRecord::RecordNotFound' do
+    respond_to do |format|
+      format.html do
+        fail ActionController::RoutingError, 'Not Found'
+      end
+    end
+  end
+end
--- a/app/controllers/api/v0/bids_controller.rb
+++ b/app/controllers/api/v0/bids_controller.rb
@@ -16,12 +16,4 @@ class Api::V0::BidsController < ApplicationController
       render json: { error: @bid.errors.full_messages.to_sentence }, status: 403
     end
   end
-
-  rescue_from 'ActiveRecord::RecordNotFound' do
-    respond_to do |format|
-      format.html do
-        fail ActionController::RoutingError, 'Not Found'
-      end
-    end
-  end
 end
--- a/app/controllers/auctions_controller.rb
+++ b/app/controllers/auctions_controller.rb
@@ -1,39 +1,19 @@
 class AuctionsController < ApplicationController
   def index
     @auction_collection = AuctionsIndexViewModel.new(
-      auctions: published_auctions,
+      auctions: AuctionQuery.new.public_index,
       current_user: current_user
     )
 
-    respond_to do |format|
-      format.html do
-        @auction_collection.sam_status_message_for(flash)
-      end
-      format.json do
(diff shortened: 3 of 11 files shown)


app/controllers/api/v0/bids_controller.rb
```diff
@@ -0,0 +1,27 @@
...
+    @bid = PlaceBid.new(params: params, bidder: current_user, via: via)
+
+    if @bid.perform
+      render json: @bid.bid, serializer: BidSerializer
+    else
+      render json: { error: @bid.errors.full_messages.to_sentence }, status: 403
+    end
+  end
+
+  rescue_from 'ActiveRecord::RecordNotFound' do
```
jessieay: do we need this?

app/controllers/auctions_controller.rb
```diff
@@ -1,39 +1,19 @@
 class AuctionsController < ApplicationController
   def index
     @auction_collection = AuctionsIndexViewModel.new(
-      auctions: published_auctions,
+      auctions: AuctionQuery.new.public_index,
       current_user: current_user
     )
 
-    respond_to do |format|
```
jessieay: 😂  (joy)

config/routes.rb
```diff
@@ -1,29 +1,10 @@
...
-  end
-
-  resources :bids, only: [:index]
-
-  get '/my-bids', to: 'bids#my_bids'
-
-  resources :users, only: [:update]
-  get 'users/edit' => 'users#edit'
-
+  # Admin needs to go above redirect
```
jessieay: not sure I understand this comment

config/routes.rb
```diff
@@ -39,4 +20,41 @@
...
+        resources :bids, only: [:new, :create, :index] do
+          collection do
+            post :confirm
+          end
+        end
+      end
+    end
+  end
+
+  # Temporarily send JSON requests to web to API
```
jessieay: when will we no longer do this? maybe make note in the comment?

spec/controllers/api/bids_controller_spec.rb
```diff
@@ -0,0 +1,5 @@
+require 'rails_helper'
+
+RSpec.describe Api::BidsController, type: :controller do
```
jessieay: remove empty spec file?

spec/requests/redirect_spec.rb
```diff
@@ -0,0 +1,18 @@
+require "rails_helper"
+
+# This file can be removed when the routing directive that sends JSON
+# requests to the public controllers is alos removed
```
jessieay: 👶 typo

app/controllers/api/v0/auctions_controller.rb
```diff
@@ -0,0 +1,9 @@
+class Api::V0::AuctionsController < ApplicationController
```
jessieay: thoughts on creating an `ApiController` that all api-related controllers inherit from? That we could slowly move shared api-related API logic into that.

jessieay:
amazing! 👏

app/controllers/api/v0/bids_controller.rb
harrisj: You're right, I guess we don't!

config/routes.rb
harrisj: This is left over from my own notes. Because I am currently keeping the RESTful stuff for admin controllers (because I want to wait until we can redo any internal scripts that use them as well), I can't rewrite requests to `/admin/auctions` to `/api/v0/admin/auctions` yet

spec/controllers/api/bids_controller_spec.rb
harrisj: ARGH. My bad

spec/requests/bids_spec.rb
```diff
@@ -44,18 +44,18 @@
       let(:bid_amount) { current_auction_price - 10 }
 
       it 'returns a json error' do
-        post auction_bids_path(auction), params, headers
+        post api_v0_auction_bids_path(auction, format: :json), params, headers
```
harrisj: Are we okay with these names?

app/controllers/api/v0/bids_controller.rb
harrisj: Actually, I am unsure. If I understand this correctly, this will take a 404 on an auction lookup and return a 500 instead?

app/controllers/api/v0/auctions_controller.rb
harrisj: Oh, I have an idea of what we will do there in a future PR

app/controllers/api/v0/bids_controller.rb
jessieay: I am not sure exactly, but I think we should generally move away from all of the exception rescuing we're doing here. Makes it hard to debug things / find errors

config/routes.rb
harrisj: When *will* we no longer do this? What do you all think?

jessieay: idk maybe when we decide to deprecate "alpha API"? (when versioning is ready / up and running)

spec/requests/bids_spec.rb
jessieay: works for me!

jessieay:
it's a beauty - merging!

PR #685, opened 2016-06-09
Use new VERSIONED API endpoint for winners page

jessieay:
* Old endpoints redirect, but we should use the versioned endpoint


harrisj:
LGTM

PR #686, opened 2016-06-09
Use USWDS gem

jessieay:
* Would also be OK with using *just* hard copies, but let's not use both
* Use version of gem that is compatible w Rails
* https://github.com/harrisj/us_web_design_standards_gem/commits/rails-assets-fixes
* https://github.com/18F/us_web_design_standards_gem/pull/7


harrisj:
Hold for me to comment on this before it is merged.

harrisj:
So, this fork of USWDS was really old and I think that we should potentially be moving to a new gem. This would be a good conversation to have with @rogeruiz on Slack next week?

jessieay:
@harrisj DM'd with Roger this morning. He is going to work to release a gem with your changes incorporated. Will close this until that is released

PR #687, opened 2016-06-10
Move admin API into api controllers

harrisj:
It doesn't look like we are testing the logic to create or update auctions via the API, so I removed it. This reorganization also lets us clean up the authenticator switching logic in the application controller by the addition of an `ApiController` base class that uses the ApiAuthenticator instead of WebAuthenticator


app/controllers/api_controller.rb
```diff
@@ -0,0 +1,9 @@
+class ApiController < ApplicationController
```
jessieay: happy dance 💃

app/controllers/admin/users_controller.rb
```diff
@@ -5,13 +5,6 @@ def index
     all_users = User.all.map {|user| UserPresenter.new(user) }
     @admins, @users = all_users.partition(&:admin?)
     @admin_report = AdminReport.new(users: all_users)
```
jessieay: is this instance var being used in the web views?

app/controllers/api/v0/admin/auctions_controller.rb
```diff
@@ -0,0 +1,12 @@
+class Api::V0::Admin::AuctionsController < ApiController
+  before_filter :require_admin
```
jessieay: this might be better for another PR, but thoughts on having an `AdminController` that all admin controllers inherit from that includes this very small but crucial `require_admin` line? On the one hand it is fine for it to be here, but on the other hand I've love to do whatever we can to ensure we don't forget to include it in new admin controllers in the future

app/controllers/api/v0/admin/users_controller.rb
```diff
@@ -0,0 +1,8 @@
+class Api::V0::Admin::UsersController < ApiController
+  before_filter :require_admin
+
+  def index
+    users = User.all.map { |user| UserPresenter.new(user) }
+    render json: AdminReport.new(users: users), serializer: ::Admin::AdminReportSerializer
```
jessieay: interesting  -- why `::Admin::AdminReportSerializer` instead of  `Admin::AdminReportSerializer` ?

jessieay:
This is looking awesome, just a few small comments / suggestions from me

app/controllers/admin/users_controller.rb
harrisj: I tried removing it but yeah, it actually is

app/controllers/api/v0/admin/auctions_controller.rb
harrisj: I hear you, but we have some API controllers that aren't admin and some admin controllers that aren't API

app/controllers/api/v0/admin/users_controller.rb
harrisj: For some reason, Rails got really confused and tried looking for Admin within the Api::V0::Admin::UsersController space

app/controllers/application_controller.rb
```diff
@@ -28,31 +28,15 @@ class ApplicationController < ActionController::Base
     handle_error(message)
   end
 
-  private
+  protected
```
harrisj: This is so I can override those methods in `ApiController`

spec/controllers/application_controller_spec.rb
```diff
@@ -1,23 +0,0 @@
-require 'rails_helper'
```
harrisj: Since I have removed the switch of authenticator, there is no purpose to this test anymore

spec/requests/bids_spec.rb
```diff
@@ -44,18 +44,18 @@
       let(:bid_amount) { current_auction_price - 10 }
 
       it 'returns a json error' do
-        post api_v0_auction_bids_path(auction, format: :json), params, headers
```
harrisj: All the `format: json` declarations were redundant

PR #689, opened 2016-06-10
Prevent ineligible vendor from getting paid.

### LINKED ISSUE #574: Prevent payment to vendor who is not a small business for auctions with start price > $3500

jessieay:
* Closes #574
* Added spec for Admin::AuctionsController#update
* Removed `let` from Admin::AuctionsController specs

Opening as new version of https://github.com/18F/micropurchase/pull/630/

### DIFF
--- a/app/services/update_auction.rb
+++ b/app/services/update_auction.rb
@@ -1,12 +1,29 @@
-class UpdateAuction < Struct.new(:auction, :params, :user)
+class UpdateAuction
+  def initialize(auction, params)
+    @auction = auction
+    @params = params
+  end
+
   def perform
     auction.assign_attributes(attributes)
+    create_purchase_request
 
+    if vendor_ineligible?
+      auction.errors.add(:base, 'The vendor cannot be paid')
+      false
+    else
+      auction.save
+    end
+  end
+
+  private
+
+  attr_reader :auction, :params
+
+  def create_purchase_request
     if should_create_cap_proposal?
       CreateCapProposalJob.perform_later(auction.id)
     end
-
-    auction.save
   end
 
   def should_create_cap_proposal?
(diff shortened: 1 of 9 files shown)


spec/services/update_auction_spec.rb
```diff
@@ -88,47 +91,10 @@
...
 
-        updater = UpdateAuction.new(auction, params)
+        UpdateAuction.new(auction, params)
 
         expect(CreateCapProposalJob).to_not have_received(:perform_later).with(auction.id)
       end
     end
   end
-
-  describe '#should_create_cap_proposal?' do
```
jessieay: this is now a private method

adelevie: as it should be 💯

app/services/update_auction.rb
```diff
@@ -1,20 +1,40 @@
-class UpdateAuction < Struct.new(:auction, :params, :user)
+class UpdateAuction
+  def initialize(auction, params)
+    @auction = auction
+    @params = params
+  end
+
   def perform
     auction.assign_attributes(attributes)
+    create_purchase_request
```
harrisj: Do you want to have something up here where it returns if the cap_request_url has been set?

app/services/update_auction.rb
```diff
@@ -1,20 +1,40 @@
...
-    auction.save
   end
 
   def should_create_cap_proposal?
     auction_accepted_and_cap_proposal_is_blank? &&
       winning_bidder_is_eligible_to_be_paid?
   end
 
-  private
+  def vendor_ineligible?
```
adelevie: solid naming

adelevie:
LGTM. I'll merge when 📗

app/services/update_auction.rb
jessieay: I don't think so?

jessieay:
@adelevie 🍏

Issue #664, opened 2016-06-03
CAP request is made only for auctions that should be paid on the 18F P-card


mtorres253:
As an admin, I want auctions that are supposed to be paid through the 18F P-card to actually be paid properly when they're accepted and those that should not be paid on the 18F P-card are NOT automatically paid via CAP

Scenario: Auction where Use 18F P-card = true is paid via CAP
Given I am a logged in admin
And there is an auction that is paid via 18F P-card
And there is a winner
And the winner has delivered acceptable code
When I accept the auction (Click "Accept" from the dropdown)
Then a payment request should be made to CAP

Scenario: Auction where Use 18F P-card = false is NOT paid via CAP
Given I am a logged in admin
And there is an auction that is NOT paid via 18F P-card
And there is a winner
And the winner has delivered acceptable code
When I accept the auction (Click "Accept" from the dropdown)
Then a payment request should NOT be made to CAP
And I should see a message "This payment needs to be handled outside of the micropurchase platform"

This depends on #663

PR #691, opened 2016-06-10
Only create purchase request when using 18F pcard

jessieay:
* If non- "default" pcard, do not create CAP proposal
* Closes https://github.com/18F/micropurchase/issues/664

### DIFF
--- a/app/services/update_auction.rb
+++ b/app/services/update_auction.rb
@@ -6,12 +6,12 @@ class UpdateAuction
 
   def perform
     auction.assign_attributes(attributes)
-    create_purchase_request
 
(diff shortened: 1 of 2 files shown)


app/services/update_auction.rb
```diff
@@ -28,7 +28,7 @@ def create_purchase_request
 
   def should_create_cap_proposal?
     auction_accepted_and_cap_proposal_is_blank? &&
-      winning_bidder_is_eligible_to_be_paid?
```
jessieay: we already check this in `vendor_ineligible?`

PR #693, opened 2016-06-11
Fix the issue with the pending iso8601 test

harrisj:
The issue is that the JSON includes milliseconds sometimes, but iso8601 doesn't normally put them out


jessieay:
very cool!

PR #709, opened 2016-06-13
Set billable_to field based on Tock data

jessieay:
* Store Tock project data in our database
* Update once daily via clock task
* Change billable_to field to be dropdown rather than input


jessieay:
Something to think about while reviewing: should `billable_to` still be a string field, or should be create an actual relation between an `auction` and a `tock_project` ?

app/services/tock_importer.rb
```diff
@@ -0,0 +1,26 @@
+class TockImporter
+  TOCK_PROJECTS = "https://tock-app.18f.gov/api/projects.json".freeze
+
+  def run
```
jessieay: this should probably be `perform`

config/clock.rb
```diff
@@ -9,4 +9,9 @@ module Clockwork
       LosingBidderEmailSender.new(auction).delay.perform
     end
   end
+
+  every(1.day, "tock_projects.import", at: "02:00", tz: "UTC") do
```
adelevie: 🍧 awesome

app/models/tock_project.rb
```diff
@@ -0,0 +1,18 @@
+class TockProject < ActiveRecord::Base
```
adelevie: we should consider a more generic name here.

db/migrate/20160613181643_create_tock_projects.rb
```diff
@@ -0,0 +1,10 @@
+class CreateTockProjects < ActiveRecord::Migration
```
adelevie: rename would apply here, too.

app/models/tock_project.rb
jessieay: ideas? `billable_entity` ?

features/admin_creates_auction.feature
```diff
@@ -0,0 +1,29 @@
+Feature: Admin creates auctions in the admins panel
+  As an admin
+  I should be able to create auctions
+
+  Scenario: Adding an auction
+    Given I am an administrator
+    And there is a tock project to bill to
```
adelevie: I wonder if this should be swept into a more generic setup block, since it doesn't seem to relate to the user interface.

app/models/tock_project.rb
adelevie: making a quick first pass at this whole thing, then I'll circle back with names, but yes, something like `billable_entity`...

features/admin_creates_auction.feature
```diff
@@ -0,0 +1,29 @@
...
+    And I click on the "Create a new auction" link
+    And I edit the new auction form
+    And I click to create an auction
+    Then I should see the auction's title
+    And I should see the start time I set for the auction
+    And I should see the end time I set for the auction
+
+  Scenario: Creating an invalid auction
+    Given I am an administrator
+    And there is a tock project to bill to
```
adelevie: same as above

spec/rails_helper.rb
```diff
@@ -31,6 +31,7 @@
   config.before do
     mock_github
     WebMock.stub_request(:any, /api.data.gov/).to_rack(FakeSamApi)
+    WebMock.stub_request(:any, /tock-app.18f.gov/).to_rack(FakeTockApi)
```
adelevie: 💯

features/admin_creates_auction.feature
jessieay: I like the idea of only creating it when it is really needed for a test.

The reason for my hesitation is that putting it somewhere general could get us into the territory of creating seed data before each test like C2, whose test suite I ran those morning and OH BOY let me tell you is it slow. 

If you were to make more generic, where would you put it?

adelevie:
Ok, this great. All good, but we should resolve the name of the new model.

Maybe it's too much to completely avoid introducing 18F-specific bureaucracy into the app, but I think we should at least try not to.

Some names to replace `TockProject`:

- billable_entity
- accounting_entity
- billable_project
- billable_entity
- account
- client_account
- funder
- funding_account

jessieay:
@adelevie renamed to `ClientAccount` and pushed 🍒

adelevie:
small failing spec, but I'll merge when ♻️

jessieay:
@adelevie 🍏

adelevie:
⛵

Issue #673, opened 2016-06-03
Cap request contains line item data for billable projects


mtorres253:
As an admin, I want to make sure my cap request specifies that the project is billable if the system knows that is is, so if the system knows this, it should tell Cap

Scenario: Cap payment request includes billable or non-billable information
Given that an auction has been accepted
And the project is in Tock
When the request for payment is made to CAP
Then the justification attribute in the request should be the name of the line item (e.g. billable to the Cloud.gov tock item)

Note: Just make sure that we have some clear indication that the project is billable. Could be as simple as having (billable) in the text.

Issue #672, opened 2016-06-03
Billable to: field data from Tock


mtorres253:
As an admin, I want to store accurate information about who is paying for the auction (with whom we have the agreement), so the system should limit my options to just those that actually exist

Scenario: I select a Tock project to bill to when I create a new auction
Given I am a logged in admin
When I'm creating a new auction
Then I should see a select dropdown labeled "Billable To"
And that dropdown should contain all Tock projects here https://tock.18f.gov/api/projects.json
And I see whether the project is billable or not billable
And this field is required
And I should be able to save this data

https://tock.18f.gov/api/projects.json

Note: ensure that old data conforms to our new way of doing things

jessieay:
Opened an issue on Tock with some API questions here https://github.com/18F/tock/issues/363

mtorres253:
Project mapping:

Tock -> Tock(62)
Open Opportunities -> Custom Partner Solutions / OCSIT / Open Opportunities (206)
Micropurchase -> Acquisition Services / Marketplaces / Micropurchasing (237)
GP - API Program - GP - API Program (92)
Consulting / Marketplaces / Tools -> [blank]
Playbook in Action -> GP - Playbook in Action (83)

PR #712, opened 2016-06-13
Send Tock data to CAP

jessieay:
* In justification field, include name of Tock project
* Include whether project is billable or not
* Also adds validation on Billable_to field (required field)
* Small Rake task to update existing auctions on prod

* Finishes up https://github.com/18F/micropurchase/issues/672
* Closes https://github.com/18F/micropurchase/issues/673

### DIFF
--- a/db/chores/update_auctions_billable_data.rb
+++ b/db/chores/update_auctions_billable_data.rb
@@ -1,0 +1,21 @@
+class UpdateAuctionsBillableData
+  def perform
+    auction_tock_mappings.each do |mapping|
+      auctions = Auction.where(billable_to: mapping[:auction_billable_to])
+      account = ClientAccount.find_by(tock_id: mapping[:tock_id])
+      auctions.each { |auction| auction.update(billable_to: account.to_s) }
+    end
+  end
+
+  private
+
+  def auction_tock_mappings
+    [
+      { auction_billable_to: "Micropurchase", tock_id: 237 },
+      { auction_billable_to: "Open Opportunities", tock_id: 206 },
+      { auction_billable_to: "Tock", tock_id: 62 },
+      { auction_billable_to: "GP - API Program", tock_id: 92 },
+      { auction_billable_to: "Playbook in Action", tock_id: 83 }
+    ]
+  end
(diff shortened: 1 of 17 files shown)


db/chores/update_auctions_billable_data.rb
```diff
@@ -0,0 +1,21 @@
+class UpdateAuctionsBillableData
```
adelevie: Is this a thing that we have because we don't use relational Tock data?

db/chores/update_auctions_billable_data.rb
```diff
@@ -0,0 +1,21 @@
...
+    auction_tock_mappings.each do |mapping|
+      auctions = Auction.where(billable_to: mapping[:auction_billable_to])
+      account = ClientAccount.find_by(tock_id: mapping[:tock_id])
+      auctions.each { |auction| auction.update(billable_to: account.to_s) }
+    end
+  end
+
+  private
+
+  def auction_tock_mappings
```
adelevie: So this will need to be updated as we add projects?

adelevie:
Looks good. Had a couple questions.

db/chores/update_auctions_billable_data.rb
jessieay: if we made it relational we would still need to update. Here I am updating just for consistency. Eg: Rather than `"Micropurchase"` in the `billable_to` field we will now have `"Acquisition Services / Marketplaces / Micropurchasing (Billable)"` (what you would select in the form if you created the auction today)

adelevie: cool!

db/chores/update_auctions_billable_data.rb
jessieay: we shouldn't have to in the future because people will create auctions with the correct billable data. This is just for our existing auctions

adelevie:
all good, except for tests, merging when 🚥 is 🍀

jessieay:
yyaaaaaay 💚

adelevie:
🎳

PR #715, opened 2016-06-14
Deploy to prod 6/14/2016

jessieay:


jessieay:
this hakiri warning is a false positive.

PR #718, opened 2016-06-14
Remove unused json routes

jessieay:
* These are now in their own controllers
* Where they belong


harrisj:
Thanks for cleaning this up!

PR #719, opened 2016-06-14
Empty string is default duns number

jessieay:
* NR error on prod: `Samwise::Error::InvalidFormat: The format of the provided DUNS number is invalid`
* This is because some users have duns number as empty string, which was
  previously considered valid but should be considered same as having "no duns number"
* Add feature spec for this behavior


app/services/update_user.rb
```diff
@@ -33,7 +33,24 @@ def update_sam
     reckoner.delay.set!
   end
 
+  def parsed_user_params
+    user_params.each_with_object({ }) do |(key, value), hash|
+      converted_value = nil_not_empty_string(value)
+      hash[key] = converted_value
+    end
+  end
```
monfresh: Why not process an empty string DUNS number in the validator? https://github.com/18F/micropurchase/blob/develop/app/validators/duns_number_validator.rb

That's where I would think this logic belongs. It feels inefficient to me to call `nil_not_empty_string(value)` for every param when the only one you care about is the DUNS number.

jessieay: good idea!

jessieay: actually, on second thought, I think I want to keep this. I don't want different attributes to be set to empty string if someone changes the form to remove the value in a field.

monfresh: I see. So for those String attributes that are not required, you prefer them to be stored in the database as nil as opposed to an empty string. If that's the case, would this work?
```ruby
def normalized_user_params
  user_params.delete_if { |_, v| v == '' }
end
```

monfresh: Also, if you need more advanced normalization, such as making sure the input doesn't have extra spaces, I recommend this gem: https://github.com/holli/auto_strip_attributes

harrisj: Yeah, I think I've just deleted the blank params in the past when I've had this issue. But that would make for a weird UX issue if a user cleared their name. Instead of getting a validation error (for the name being set to `nil`), it would just not update the name because the param was removed.

monfresh: Good point.

jessieay: updated solution: set db default for fields that can be `nil` to `""` so when the form is updated and the value is `""` we do not record that as a change. Removes need to fancy params parsing and is also nice bc now we can render these values in the views without worrying about `nil` errors

features/step_definitions/user_profile_form_steps.rb
```diff
@@ -42,7 +42,7 @@
 
 When(/^there is no (.+) associated with my account$/) do |attribute|
   attribute = attribute.parameterize('_')
-  @user.update_attribute(attribute, nil)
+  @user.update(attribute.to_sym => '')
```
harrisj: Can we just use `@user.update_attribute(attribute, '')`. I don't think that method needs a sym coercion

spec/models/login_user_spec.rb
```diff
@@ -5,7 +5,7 @@
     context "user does not have github username, name, or email" do
       it "updates from auth hash" do
         github_id = '1234'
-        user = build(:user, github_login: nil, name: nil, email: nil, github_id: github_id)
+        user = build(:user, github_login: nil, name: "", email: "", github_id: github_id)
```
harrisj: Does our Rubocop style guide prefer '' over "" for strings that don't have interpolation? I don't remember anymore

features/step_definitions/user_profile_form_steps.rb
jessieay: I prefer `update` to `update_attribute` bc it runs validations. See http://www.davidverhasselt.com/set-attributes-in-activerecord

spec/models/login_user_spec.rb
jessieay: ah yeah I think it does! will update

jessieay: updated at a8edacc

harrisj:
Added a few nitpicks. One question, can we also add a Cucumber feature test to `vendor_edits_profile` for when the user sets their DUNS number to blank? Just want to test that they can do that okay and also that it resets the `sam_status` and `small_business` flags

features/step_definitions/user_profile_form_steps.rb
harrisj: Ah, got it. Good point.

jessieay:
@harrisj GREAT call on testing -- added test and updated behavior here https://github.com/18F/micropurchase/pull/719/commits/df51b1b9a277adcf318e341af027cc2f3dd05a3f

harrisj:
LGTM. Merging.

PR #721, opened 2016-06-15
Add an "Edit Auction" button for admins

### LINKED ISSUE #696: link to edit page (admin only)
As an admin user, I want to be able to edit an auction from the auction show page, so that I don't have to go back to the auctions index page first, so there should be an edit link on the show page itself

Scenario: Admin navigates to edit page from auction show page
Given I am an admin user
And there is an auction
When I click on the "Edit" link on the auction show page (https://micropurchase.18f.gov/admin/auctions/[id])
Then I should be taken to the auction edit page (https://micropurchase.18f.gov/admin/auctions/[id]/edit)

harrisj:
This should only appear on auctions if the current user is an admin.

Fixes #696


app/controllers/admin/auctions_controller.rb
```diff
@@ -35,14 +35,16 @@ def create
 
   def edit
     @auction = Admin::EditAuctionViewModel.new(Auction.find(params[:id]))
+    session[:return_to] = request.referer
```
jessieay: thoughts on extracting this into some kind of method and putting in `ApplicationController` where we already have some redirect behavior? https://github.com/18F/micropurchase/blob/develop/app/controllers/application_controller.rb#L42

app/presenters/user_presenter.rb
```diff
@@ -4,7 +4,11 @@ def github_json_url
   end
 
   def admin?
-    Admins.verify?(model.github_id)
+    authenticated? && Admins.verify?(model.github_id)
+  end
+
+  def authenticated?
```
jessieay: we shouldn't be using this presenter for a `Guest` - they get `guest_presenter`

app/view_models/auction_show_view_model.rb
```diff
@@ -140,6 +140,14 @@ def lowest_bidder_name
     auction.lowest_bid.bidder.name
   end
 
+  def admin_edit_auction_partial
```
jessieay: woooooooo

app/view_models/auction_show_view_model.rb
```diff
@@ -140,6 +140,14 @@ def lowest_bidder_name
...
+    if UserPresenter.new(current_user).admin?
```
jessieay: thoughts on having this be `current_user.decorate.admin_edit_auction_partial` and then having a `admin_edit_auction_partial` method in `UserPresenter` and `GuestPresenter` ?

features/admin_views_auction.feature
```diff
@@ -0,0 +1,24 @@
...
+    Given I am an administrator
+    And I sign in
+    And there is an open auction
+    And I visit the auction page
+    Then I should see an "Edit Auction" link
+
+    When I click on the "Edit Auction" link
+    Then I should be on the admin form for that auction
+
+    When I click on the "Update" button
```
jessieay: I think these last two steps are not necessary -we have separate tests for admin editing auction stuff

jessieay:
Just a few 👶  comments from me - lookin' good!

jessieay:
@harrisj merge conflicts

app/view_models/auction_show_view_model.rb
harrisj: Sounds good

features/admin_views_auction.feature
harrisj: This is testing the redirection back to the /auctions/{id} page

app/controllers/admin/auctions_controller.rb
harrisj: Good call

app/presenters/user_presenter.rb
harrisj: Oh, totally. My bad. Fixing all this now

app/controllers/application_controller.rb
```diff
@@ -44,4 +44,15 @@ def store_location
       session[:return_to] = request.original_fullpath
     end
   end
+
+  def store_referer
```
jessieay: thoughts on combining with https://github.com/18F/micropurchase/blob/6a3e1bcd937424b443acc98bac2acdc33bb5ff9d/app/controllers/application_controller.rb#L42

app/presenters/user_presenter.rb
```diff
@@ -35,6 +35,14 @@ def welcome_message_partial
     'components/null'
   end
 
+  def admin_edit_auction_partial
+    if admin?
```
jessieay: oooh we could even do an `AdminPresenter` !

features/admin_approves_delivery.feature
```diff
@@ -8,18 +8,18 @@ Feature: Admin approves delivery of a project
     Given I am an administrator
     And there is an auction that needs evaluation
     And I sign in
-    When I visit the admin edit page for that auction
+    When I visit the admin form for that auction
```
jessieay: much better!

features/admin_views_auction.feature
jessieay: ok so maybe this spec should be in a different file then? seems like are not quite testing `admin views auction` here

app/controllers/application_controller.rb
harrisj: Well, one is the path the user requested and the other is the location they came from...

app/presenters/user_presenter.rb
harrisj: I thought about that, but I decided against it for now

app/controllers/application_controller.rb
jessieay: ok cool - maybe in the future we can reconstruct the requests so that they can share logic since these seem highly related

PR #722, opened 2016-06-15
Only one API bids method to create new bids

harrisj:
There were some routes and methods in the API bids controller that should not be there


jessieay:
@harrisj just confirming: we return array of bids at auctions endpoint ,right?

Issue #669, opened 2016-06-03
System checks CAP to get payment date


mtorres253:
As an admin I want to see when someone was actually paid via CAP in relation to when the auction was delivered, so that I can measure the effectiveness of the payment method

Scenario: Admin sees payment date on auction view form for payment made via CAP
Given I am a logged in admin
And there is an auction that has been successfully accepted
And a payment request was successfully made via CAP
And the payment was made to the vendor
When I go to the show page for the auction (https://micropurchase.18f.gov/admin/auctions/24)
I should see the date the payment was made

As an vendor I want to see when someone was actually paid via CAP in relation to when the auction was delivered, so that I can get a sense of how quickly winners get paid in general.

Scenario: Vendor sees payment date on auction view form for payment made via CAP
Given I am a logged in vendor
And there is an auction that has been successfully accepted
And a payment request was successfully made via CAP
And the payment was made to the winning vendor
When I go to the show page for the auction (https://micropurchase.18f.gov/auctions/[24])
I should see the date the payment was made

As an unknown visitor, I want to see when someone was actually paid via CAP in relation to when the auction was delivered, so that I can get a sense of how quickly winners get paid in general. (This might motivate more vendors to sign up, and more customers to use the micropurchase platform)

Scenario: Unknown visitor sees payment date on auction view form for payment made via CAP
Given I am an unknown visitor to the micropurchase site
And there is an auction that has been successfully accepted
And a payment request was successfully made via CAP
And the payment was made to the winning vendor
When I go to the show page for the auction (https://micropurchase.18f.gov/auctions/[24])
I should see the date the payment was made

Note: If it's not there already, let's add a pull request to Cap to add this information. Add the check to the once a day update. Make sure this is monitored by NewRelic

jessieay:
https://github.com/18F/C2/blob/master/app/serializers/proposal_serializer.rb

so we will want to look at the “purchase” step https://github.com/18F/C2/blob/master/app/serializers/step_serializer.rb
and see if it has a status of `completed`
looks like there’s not timestamp for that — I can do a PR to add when that card comes up
there is a `completed_at` timestamp in the C2 API for steps

jessieay:
PR to C2 to add `completed_at` to API: https://github.com/18F/C2/pull/1343

PR #723, opened 2016-06-16
System checks CAP to get payment date

jessieay:
* Display payment date in admin auction#show and regular auction#show
* Check auctions pending payment once daily via cron job
* Closes https://github.com/18F/micropurchase/issues/669


jessieay:
Few notes on this one for reviewers:

* I want to share the C2 config used in the new service class and our existing `CreateCapProposal` class -- what do you think is the best way to do that?

* I removed the ability to set the paid timestamp via admin. Is that a problem? I did that because there weren't any stories around how this is set via admin. Formerly it was a dropdown, now it is a datetime so we could either do a datepicker or a `true` or `false` dropdown.

app/controllers/api/v0/admin/auctions_controller.rb
```diff
@@ -2,11 +2,11 @@ class Api::V0::Admin::AuctionsController < ApiController
   before_filter :require_admin
 
   def index
-    render json: Auction.all, each_serializer: Admin::AuctionSerializer
+    render json: Auction.all, each_serializer: ::Admin::AuctionSerializer
```
harrisj: lol Rails constant autoloading

app/serializers/admin/auction_serializer.rb
```diff
@@ -2,35 +2,31 @@ class Admin::AuctionSerializer < ActiveModel::Serializer
...
-  end
-
   def created_at
     object.created_at.iso8601
   end
 
-  def updated_at
-    object.updated_at.iso8601
+  def delivery_due_at
+    object.delivery_due_at.try(:iso8601)
```
harrisj: Not going to affect this PR, but these `try` clauses feel like a smell that would have to be resolved at some point

app/services/cap_payment_checker.rb
```diff
@@ -0,0 +1,43 @@
+class CapPaymentChecker
```
harrisj: Should we rename this to `CheckCapPayment` to match the verb names of other services?

app/view_models/admin/auction_show_view_model.rb
```diff
@@ -72,6 +76,6 @@ def auction_status
   end
 
   def formatted_time(time)
-    DcTimePresenter.convert(time).to_s(:long)
+    DcTimePresenter.convert_and_format(time)
```
harrisj: Nice

app/view_models/admin/edit_auction_view_model.rb
```diff
@@ -14,15 +14,27 @@ def delivery_due_partial
   end
 
   def date_default(field)
-    dc_time(field).to_date
+    if field == 'paid'
```
harrisj: Probably better than creating a NullTime class

db/migrate/20160615205001_add_paid_at_to_auctions.rb
```diff
@@ -0,0 +1,24 @@
+class AddPaidAtToAuctions < ActiveRecord::Migration
+  def up
+    add_column :auctions, :paid_at, :datetime
+
+    execute <<-SQL
+      UPDATE auctions
+      SET paid_at = CURRENT_TIMESTAMP
```
harrisj: I know you have to do this because `paid_at` is also used like a boolean flag for payment. It would be good if we could retroactively reconcile this to have the correct times later, because it would skew things if we try to do a data analysis of average payment delays. Can you add an issue to fix `paid_at` in production once this PR is applied?

features/vendor_views_closed_auction.feature
```diff
@@ -38,6 +38,11 @@ Feature: Vendor views closed auctions
     And I should see when the auction ended
     And there should be meta tags for the closed auction
 
+  Scenario: Winning vendor was paid
```
harrisj: This seems to be a test for a guest who isn't logged in. Maybe we should consider a `background` for all of these scenarios to guarantee the user is logged in?

spec/models/auction_query_spec.rb
```diff
@@ -114,7 +114,7 @@
     let!(:running_auction) { create(:auction, :running) }
     let!(:rejected_auction) { create(:auction, result: :rejected) }
     let!(:unpaid_auction) do
-      create(:auction, awardee_paid_status: :not_paid)
+      create(:auction, :not_paid)
```
harrisj: So much better

spec/requests/api/v0/admin/auctions_spec.rb
```diff
@@ -79,10 +79,10 @@
         get api_v0_admin_auctions_path, nil, headers
 
         json_auctions = JSON.parse(response.body)['auctions']
-        expect(json_auctions.map { |a| a['created_at'] }).to all(be_iso8601)
-        expect(json_auctions.map { |a| a['updated_at'] }).to all(be_iso8601)
+        # expect(json_auctions.map { |a| a['created_at'] }).to all(be_iso8601)
```
harrisj: Are these commented out because they are failing again?

harrisj:
I don't have an issue with the paid timestamp selector being removed from the form. I'm not sure what the best approach for sharing the C2 config should be yet, but let me have more coffee and think about it.

app/serializers/admin/auction_serializer.rb
jessieay: agreed!

app/services/cap_payment_checker.rb
jessieay: I like it!

db/migrate/20160615205001_add_paid_at_to_auctions.rb
jessieay: yeah good call I thought of that...do we have a way of knowing when past auctions were paid? Guessing that process would be manual? @adelevie

features/vendor_views_closed_auction.feature
jessieay: Yeah this feature is for anyone, guest or vendor -- will put in a different file to make that more clear

spec/requests/api/v0/admin/auctions_spec.rb
jessieay: oops! commented out during some debugging stuff will revert

db/migrate/20160615205001_add_paid_at_to_auctions.rb
harrisj: Yeah, I think we'd have to look at the C2 records for when the payment was processed

app/controllers/api/v0/admin/auctions_controller.rb
jessieay: 👾

app/services/cap_payment_checker.rb
jessieay: renamed at 1558d83

app/view_models/admin/edit_auction_view_model.rb
jessieay: ah yeah we actually don't need these now -- added when I was trying to put payment fields into form, which was getting complicated bc we wanted to default to nil. Will remove for now ;)

features/vendor_views_closed_auction.feature
jessieay: Moved spec at e08444b

jessieay:
ok @harrisj I think I've addressed all of your comments here

db/migrate/20160615205001_add_paid_at_to_auctions.rb
jessieay: Cool - updated the service class at f23d52e so we can run it via console for all existing auctions and get the payment dates via the CAP API

adelevie:
LGTM!

db/migrate/20160615205001_add_paid_at_to_auctions.rb
harrisj: Awwww snap! Nice

harrisj:
Let's merge this puppy

PR #725, opened 2016-06-17
Initial swagger documentation for public API

harrisj:
If you were curious how Swagger JSON looks and how testing against it might look, here you go. Note that it is currently failing because the swagger file says that we might get a 404 for `/auctions`, but I haven't been able to produce that yet (will it ever happen?)

This also isn't generatiing docs yet, but that would be a next step


harrisj:
This is mostly done. I will hold off on generating documentation for a future PR. I just want to verify I have good documentation for if you submit a bid and the auction is closed and remove a few more redundant tests elsewhere.

The one annoying thing about Swagger testing here with Apivore is that it validates the schema and the responses, but it doesn't give you an easy way to look at the values in the response. So I still have a few request tests that look at the values returned.

app/controllers/api/v0/admin/auctions_controller.rb
```diff
@@ -8,5 +8,7 @@ def index
   def show
     auction = Auction.find(params[:id])
     render json: auction, serializer: ::Admin::AuctionSerializer
+  rescue ActiveRecord::RecordNotFound
```
jessieay: maybe we can put all of these rescues into the `api_controller` like  http://stackoverflow.com/a/24451181/1019369 ?

public/api/swagger.json
```diff
@@ -0,0 +1,838 @@
+{
+    "swagger": "2.0",
```
jessieay: can we nest these with two spaces instead of 4? Finding the deep nesting hard to read..

public/api/swagger.json
```diff
@@ -0,0 +1,838 @@
...
+                                        "description": "Quis autem vel eum iure reprehenderit qui in ea voluptate velit esse quam nihil molestiae consequatur, vel illum qui dolorem eum fugiat quo voluptas nulla pariatur?\r\n\r\n## A Markdown Quote\r\n\r\n> Ut enim ad minima veniam, quis nostrum exercitationem ullam corporis suscipit laboriosam, nisi ut aliquid ex ea commodi consequatur? Quis autem vel eum iure reprehenderit qui in ea voluptate velit esse quam nihil molestiae consequatur, vel illum qui dolorem eum fugiat quo voluptas nulla pariatur?\r\n\r\n```\r\ndef foo(bar)\r\n  10.times do\r\n     puts bar\r\n   end\r\nend\r\n```",
+                                        "id": 1,
+                                        "created_at": "2015-12-21T16:40:01+00:00",
+                                        "updated_at": "2015-12-21T16:40:01+00:00",
+                                        "summary": "## Summary\r\n\r\nWe need some prose to be proof-read.",
+                                        "bids": [
+                                            {
+                                                "bidder_id": null,
```
jessieay: so we will return an empty record if there are no bids? I am a little confused by this data

public/api/swagger.json
```diff
@@ -0,0 +1,838 @@
...
+                },
+                "bids": {
+                    "type": "array",
+                    "items": {
+                        "$ref": "#/definitions/Bid"
+                    }
+                }
+            }
+        },
+        "Bid": {
```
jessieay: we should decide on what type of case we're gonna use in our API - I think we've been using snake case in past, although I know some people say JSON should be camel case.

spec/requests/swagger_spec.rb
```diff
@@ -0,0 +1,454 @@
+require 'rails_helper'
+
+RSpec.describe 'Swagger spec', type: :apivore, order: :defined do
```
jessieay: is it possible to break these out into separate spec files for each endpoint? Find it easier to read / understand them that way

spec/requests/swagger_spec.rb
```diff
@@ -0,0 +1,454 @@
...
+        expect(subject).to validate(:get,
+                                    '/admin/auctions/{id}',
+                                    403,
+                                    "id" => auction.id, "_headers" => headers(api_key))
+      end
+    end
+
+    context 'when the API key is valid' do
+      it 'returns a 200 HTTP response' do
+        user = FactoryGirl.create(:admin_user)
```
jessieay: seems like we're using `FactoryGirl.create` and `create` in here - fine with either, but can we stick with one or the other per spec file?

jessieay:
Cool you got this working! My high-level feedback on the Swagger json is that it would be much more useful broken out into separate files. Having a single file for all of our API endpoints is hard to read / update.

app/controllers/api/v0/admin/auctions_controller.rb
harrisj: I wasn't sure if I should put them here or be explicit. But I can do that

public/api/swagger.json
harrisj: I can do that. Let me just convert to tabs :trollface:

public/api/swagger.json
harrisj: This example is taken from Alan's documentation. I think this is a case of redaction in action

public/api/swagger.json
harrisj: Sadly, it's often determined by Swagger

spec/requests/swagger_spec.rb
harrisj: I tried that. Putting them all in here let's me do the trick at the bottom where it fails if there is an API endpoint I have described but I haven't checked. See here https://github.com/westfieldlabs/apivore for details

spec/requests/swagger_spec.rb
harrisj: I've been a fan of the explicit version, but I don't care really

harrisj:
I've thought about splitting into multiple files but I like that I can edit it all in [the Swagger interactive editor](http://editor.swagger.io/#/) when it's a single file. But you are right that it is long

jessieay:
@harrisj this ready for merge? Looks good to me! Perhaps we should add a note in docs on swagger usage?

PR #728, opened 2016-06-17
Small style cleanup for new bids toggle

jessieay:
* Prevent page from changing width by using `usa-grid` instead of
  `usa-grid-full`
* Remove table margin

Before:
screen shot 2016-06-17 at 11 08 19 am

After:
screen shot 2016-06-17 at 11 08 06 am


adelevie:
Not a CSS expert, but LGTM.

jessieay:
@adelevie 🍏 👼

PR #729, opened 2016-06-17
Update color of welcome slab

jessieay:
* Was using `color-gray` which we over-rode from USWDS
* Now we are using vars from USWDS so we needed to change or it was
  super dark


adelevie:
waiting for :leaves:

jessieay:
@adelevie 🍏 🐶

Issue #717, opened 2016-06-14
Winning bidders get automated email


mtorres253:
As an admin of micropurchase, I want to notify winners as soon as I can, so that I don't have to worry about manually notifying them, and so that they can be notified promptly. So, the system should send an automated email to the winner.

Scenario: Winning bidders get email notification that they've won the auction
Given there is an auction
And the auction has a winning bid from a qualified vendor
When the auction ends (or as soon as possible)
Then the winning vendor should receive an email from the system
And the email subject should be: “18f Micropurchase Auction Status: You have the winning bid!
And the from address should be: “admin@micropurchase.gov”
And the to address should be: [vendor email]
And the body should be:

Dear [Vendor Name]

Congratulations! You have the winning bid for [auction name]. This means that you have the opportunity to get paid [bid amount] if you deliver the requested work by [delivery timestamp] Eastern Time. If you have any questions about the issue, please add a comment to [github issue].  

Thanks for participating and we look forward to seeing what you can do!

Sincerely,

The 18f Micropurchase team

PR #731, opened 2016-06-17
Email the winner

adelevie:
fixes https://github.com/18F/micropurchase/issues/717


adelevie:
TODO:

- feature specs
- triggering the email sending when the auction ends

jessieay:
@adelevie re: triggering when auction ends -- I'd look into the clockwork task we have for losing bidder email (`config/clock.rb`) and re-use that if we can. I believe it looks at which auctions have closed in the past day and emails bidders

adelevie:
awesome, thank you!

config/clock.rb
```diff
@@ -10,6 +10,13 @@ module Clockwork
     end
   end
 
+  every(1.day, "winning_bidder_emails.send", at: "17:00", tz: "UTC") do
+    puts "Sending winning bidder emails"
+    AuctionsClosedWithinLastTwentyFourHoursFinder.new.perform do |auction|
```
adelevie: This seemed like a good idea at the time ...

spec/services/auctions_closed_within_last_twenty_four_hours_finder_spec.rb
```diff
@@ -0,0 +1,15 @@
+require 'rails_helper'
+
+describe AuctionsClosedWithinLastTwentyFourHoursFinder do
+  describe '#perform' do
+    it 'returns auctions with an end datetime that occured today' do
+      Timecop.freeze(Time.parse("19:00:00 UTC")) do
```
adelevie: this area, I think, would benefit most from review

jessieay:
@adelevie cool - looks like you set it to "17:00 UTC" which is 5pm UTC. what if an auction ends after 5pm UTC? In that case, they wouldn't get the email until the next day..

adelevie:
ah good call, @jessieay. and 5pm UTC is noon EDT.

What time settings do you think make sense here? 

cc @mtorres253: product question ^

app/services/auctions_closed_within_last_twenty_four_hours_finder.rb
```diff
@@ -0,0 +1,9 @@
+class AuctionsClosedWithinLastTwentyFourHoursFinder
```
jessieay: nice name!

jessieay: thoughts on sharing logic with `AuctionsClosedYesterdayFinder` ?

config/locales/en.yml
```diff
@@ -42,8 +42,17 @@ en:
         para_2: "Thank you for your participation and we look forward to working
           with you in the future."
         sign_off: "Sincerely,"
-        from: "The 18f Micropurchase team"
-        subject: "18f Micropurchase Auction Status"
+        from: "The 18F Micro-purchase team"
```
jessieay: yay !

features/vendor_views_closed_auction.feature
```diff
@@ -13,6 +13,8 @@ Feature: Vendor views closed auctions
     And I should see when the auction started
     And I should see when the auction ended
     And there should be meta tags for the closed auction
+    And an email notification is sent to the winning bidder
+    And I should receive an email notifying me that I won
```
jessieay: nice!

app/services/auctions_closed_within_last_twenty_four_hours_finder.rb
adelevie: I'm ok with that ... I was also wondering why we instead don't create and use methods on `AuctionQuery` for this.

features/vendor_views_closed_auction.feature
jessieay: another option would be to use Timecop to travel to the time that we send the emails...but not sure if clock would run in a test env

spec/services/auctions_closed_within_last_twenty_four_hours_finder_spec.rb
jessieay: One idea for improvement: whatever the time is that we send it should be a constant that we reference here

spec/services/auctions_closed_within_last_twenty_four_hours_finder_spec.rb
```diff
@@ -0,0 +1,15 @@
...
+        _old_auction = create(:auction, ended_at: 2.days.ago)
```
jessieay: maybe also create one that ended 25 hours ago?

spec/services/auctions_closed_within_last_twenty_four_hours_finder_spec.rb
```diff
@@ -0,0 +1,15 @@
...
+        newly_closed_auction = create(:auction, ended_at: 2.hours.ago)
+        _future_auction = create(:auction, ended_at: 2.days.from_now)
```
jessieay: maybe also create one that ends one hour from now?

spec/services/auctions_closed_within_last_twenty_four_hours_finder_spec.rb
adelevie: Great idea ... same thing for the losing bidder emails?

spec/services/winning_bidder_email_sender_spec.rb
```diff
@@ -0,0 +1,23 @@
+require 'rails_helper'
+
+describe WinningBidderEmailSender do
+  describe '#perform' do
+    context 'the winning bidder has an email address' do
```
jessieay: what if the winning bidder does not have an email address? (this shouldn't happen with github auth but if someone signed up before we asked for email and is somehow still logged in, it is *technically* possible)

app/services/auctions_closed_within_last_twenty_four_hours_finder.rb
jessieay: that's a great idea! I think I created that other service class before I was familiar with the codebase so I didn't know about our query class

features/vendor_views_closed_auction.feature
adelevie: I don't love that we need to use `And an email notification is sent to the winning bidder`. It's sort of like saying "And the sun rises the next day / And I should see a thing on my screen" ... the feature specs, I _think_ should assume the standard passage of time.

So maybe Timecop helps us avoid this?

spec/services/auctions_closed_within_last_twenty_four_hours_finder_spec.rb
adelevie: Easier to make these hours arguments for a method? And the arguments we end passing in would be constants (as you suggested above)?

features/vendor_views_closed_auction.feature
jessieay: hah yeah maybe we create a new feature spec for daily cron job-related stuff and in those tests we use time traveling steps? again, not sure how we can get clockwork running in cucumber but maybe not so hard...

spec/services/auctions_closed_within_last_twenty_four_hours_finder_spec.rb
jessieay: sure -- btw the reason I suggested doing a tighter timeframe is that we've dealt with some crazy stuff related to clockwork / timezones on Dolores. Basically, we had an issue where we were looking for the "current date" in a different timezone so things were getting sent at midnight ET.....if we'd had a test that included records *just* outside of the timeframe that we wanted, the bug would have been avoided

spec/services/winning_bidder_email_sender_spec.rb
adelevie: #productQuestion for @mtorres253 ^^^^

config/clock.rb
harrisj: Can we just makes this methods on AuctionQuery object?

jessieay: great minds https://github.com/18F/micropurchase/pull/731/files/e1d32ebe11e6de325e7344e94b84c977879c1cab#r67719930

config/locales/en.yml
```diff
@@ -42,8 +42,17 @@ en:
...
+        subject: "18F Micro-purchase Auction Status"
+      winning_bidder_notification:
+        para_1: "Congratulations! You have the winning bid for %{auction_title} (%{auction_url}).
+          This means that you have the opportunity to get paid %{bid_amount}
+          if you deliver the requested work by %{auction_delivery_deadline} Eastern Time.
+          If you have any questions about the issue, please add a comment to %{auction_issue_url}."
+        para_2: "Thanks for participating and we look forward to seeing what you can do!"
+        sign_off: "Sincerely,"
```
mtorres253: Can we move the existing para_2 -> para_3 and replace with the following?
`para_2: "Check out the policy page: {%{policy_page_url}} to read more about the process for submitting the code. We encourage delivery of code (even if it's incomplete) early and often so that we can work out any issues that arise before the deadline."

spec/services/winning_bidder_email_sender_spec.rb
mtorres253: I will write a story to clean that up. Seems like a corner case?

config/clock.rb
```diff
@@ -3,19 +3,26 @@
 require_relative "environment"
 
 module Clockwork
-  every(1.day, "losing_bidder_emails.send", at: "03:00", tz: "UTC") do
+  every(1.day, "losing_bidder_emails.send", at: "17:05", tz: 'Eastern Time (US & Canada)') do
     puts "Sending losing bidder emails"
-    AuctionsClosedYesterdayFinder.new.perform do |auction|
+    AuctionQuery.new.closed_yesterday do |auction|
```
jessieay: maybe we should use `.closed_within_last_24_hours` here too?

jessieay:
One final comment about auctionquery method consistency, otherwise LGTM!

config/locales/en.yml
adelevie: ✔️

PR #732, opened 2016-06-17
Update `ROOT_URL` logic

jessieay:
* Previous logic did not find a correct url
* Simpler to just set env var rather than use class / env var parsing
* From New Relic:
```
NoMethodError: undefined method `[]' for nil:NilClass
/home/vcap/app/app/models/
vcap_application.rb:   3:in `application_uris'
          /home/vcap/app/app/presenters/
auction_url.rb:  18:in `root_url'
          /home/vcap/app/app/presenters/
auction_url.rb:   9:in `find'
/home/vcap/app/app/services/
construct_cap_attributes.rb:  72:in `url'
```


adelevie:
merging when 🌳

PR #737, opened 2016-06-20
Use newest version of USWDS

jessieay:
* 0.9.4
* Installed via NPM
* See https://github.com/18F/web-design-standards
* Updated for Rails asset pipeline
* See https://github.com/18F/us_web_design_standards_gem/pull/7
* Remove gem, was only being used for javascript and we aren't using
  uswds js yet


jessieay:
There were many many code climate errors for our CSS since I moved it around. I addressed all of the low hanging fruit but think we can address the rest as we move forward.

jessieay:
Just pushed this to staging and all looks good :)

adelevie:
most of this is Greek to me, but looks good!

Issue #694, opened 2016-06-13
New layout and styling


mtorres253:
This story is to re-do the layout/styling of the existing auction show page. This will be for both the admin and vendor view. For now, keep the existing right rail, just adapt to the new layout. 

- [ ] Change the breadcrumb language to "Auctions -> [Auction name]"
- [x] Bar beneath auction title with the "Auctions Bids" nav
- [x] Ribbon with the auction status
- [x] Font and styling changes

https://pages.18f.gov/micropurchase/auctions/show/

Note: @andrewmaier will collaborate with developer and clean up the styles.

jessieay:
@mtorres253 wondering if we really need the breadcrumbs here -- people can click on the 18F logo to go back to the homepage. What do you think>

PR #740, opened 2016-06-21
Update auction show

jessieay:
* Move status flash to righthand nav
* Add border
* Update js to be more elegant
* Part of #694

screen shot 2016-06-20 at 6 18 36 pm

screen shot 2016-06-20 at 6 16 39 pm


jessieay:
remaining code climate errors are fixable but related to existing CSS that I moved. Will fix / update as I work with it more

PR #744, opened 2016-06-21
Add auction winning bid detail and time underneath title

jessieay:
*Fix weird layout issue for when no bids
* Was changing size slightly

Update logic around bid shown underneath title

* If auction is over, show winning bidder name and amount
* If user is bidder, show their lowest bidding amount
* If there are bids, show highlighted bid amount
* If there are no bids, show nothing

screen shot 2016-06-21 at 3 02 29 pm

screen shot 2016-06-21 at 3 02 13 pm


adelevie:
LGTM

Issue #698, opened 2016-06-13
Reverse Auction bid box - Active Auction


mtorres253:
As a vendor, I want to easily bid on auctions that I'm eligible for, so If I am eligible to bid on an auction then I should see an easy way to do that from the show page

Context:
Given I'm a logged in vendor
And there is a reverse auction that I'm eligible to bid on
And the auction is active (within start/end dates)
And I've not yet bid on the auction

Scenario: Eligible vendor sees a bid form
When I go to the show page for that auction
Then I should see a form at the top of the right rail
And the form should contain an input for my bid labeled "Your Bid"
And the form should contain a button labeled "Place Bid"
And I should see the starting price for the auction "Starting price: XXXX" in the bid form

Scenario: Vendor has not yet bid and sees current lowest bid in the bid box
And there is a bid
And there is still room for me to bid below the current bid
When I go to the show page for that auction
Then I should see in the bid box the current lowest bid "The current lowest bid is XXXX"

Scenario: Vendor sees that they have the lowest (winning) bid
And I have bid
And my bid is the lowest bid
When I go to the show page for the auction
Then I should see that my bid is the lowest bid "You have the lowest bid!"
And I should not see the form

Note: Look at this for reference https://pages.18f.gov/micropurchase/auctions/show/

PR #746, opened 2016-06-21
Inline bidding experience

### LINKED ISSUE #701: Inline bid experience for eligible vendor
As a vendor who is eligible to bid on an auction, I want to easily be able to do this and see if my bid was accepted or why it was not accepted. 

Context:
I am an eligible vendor
And there is a live auction that I'm eligible for
And I have entered my bid in the "Your Bid" box
When I click on the "Place Bid" button

Scenario: Eligible vendor submits bid - no errors
And there are no errors
Then I should be taken to the existing confirmation page

Scenario: Vendor is redirected to show page for the auction after confirming the bid
When I click on the confirm button
Then my bid should be saved
And I should go back to the show page for that auction
And I should see a success message ("Congratulations, your bid of $XXXX was accepted")

Scenario: Eligible vendor submits bid - with errors
And there are errors
Then my bid should not be recorded
And I should stay on the show page for that auction
And I should see an inline message describing the error (use existing error messaging)

Note: Use existing confirmation page. Just redirect to the auction show page and show success message there

jessieay:
* Closes https://github.com/18F/micropurchase/issues/698
* Closes #701
* Still need to add "You have the lowest bid!" for winner

screen shot 2016-06-21 at 4 57 09 pm


app/views/bids/_form.html.erb
```diff
@@ -1,30 +0,0 @@
...
-    </div>
-
-    <div class="usa-width-one-half">
-      <h2 class="center"><label for="bid_amount">Your bid:</label></h2>
-      <%= f.text_field :amount %>
-    </div>
-  </div>
-
-  <p>
-    Note that all users who submit bids consent to the publication of their GitHub
```
jessieay: do we need to include this warning elsewhere now that we are doing inline bidding? #lawyersneeded cc @adelevie

harrisj: I assume this is part of the Terms of Service already, right?

app/view_models/auction_show_view_model.rb
```diff
@@ -66,9 +70,17 @@ def eligibility_label
     end
   end
 
+  def bid_form_prompt
```
harrisj: Is this something we should delegate to Rules objects? Or is that polymorphism overkill?

features/vendor_views_small_biz_auctions.feature
```diff
@@ -8,13 +8,13 @@ Feature: Vendor views auctions set aside for small businesses only
     And I am a user with a verified SAM account who is not a small business
     And I sign in
     When I visit the auction page
-    Then I should not see the bid button
+    Then I should not see the bid form
```
harrisj: :clap:

app/view_models/auction_show_view_model.rb
adelevie: > Is this something we should delegate to Rules objects?

I think so.

app/views/bids/_form.html.erb
adelevie: https://github.com/18F/micropurchase/blob/develop/app/views/logins/index.html.erb has it:

> None of the content in this GitHub repository is considered proprietary or procurement-sensitive information. Furthermore, all users who submit bids consent to the publication of their GitHub usernames, DUNS numbers, and bid amounts following the conclusion of the bidding process. Bidders' email addresses will not be disclosed and will only be used for micro-purchase-related communication on an as-needed basis.

Although, looks like we should get some clarification from @allalala about this line: "None of the content in this GitHub repository is considered proprietary". We might want to change that to say "website" or something else.

adelevie:
LGTM ... your call on sweeping `bid_form_prompt` into a rules object.

So cool:

screen shot 2016-06-21 at 9 27 21 pm

harrisj:
Can I say how gorgeous things are looking? Merging.

PR #754, opened 2016-06-23
Rename Auction Types

### LINKED ISSUE #751: Single bid -> Sealed Bid, Multi-bid -> Reverse auction
This story is to migrate terminology in the code from single bid -> sealed bid, and from multi-bid to reverse auction. We are doing this to make the nature of these auctions a bit clearer to our vendors and customers.

I'll file a separate pull request for the UI updates.

harrisj:
This commit renames multi_bid -> reverse, single_bid -> sealed_bid to make things a bit more clear to the end user. Fixes #751

**Note:** This PR does not fix the many places where classes query the value of `auction.type` to determine the appropriate behavior. This caused some weird bugs and should be fixed in a future PR, but this PR is big enough as is.


adelevie:
LGTM! 📎

PR #755, opened 2016-06-23
A few places where I missed renaming things

harrisj:
There a few places not affected by the tests where I missed renaming things


adelevie:
🔨  LGTM

PR #758, opened 2016-06-23
Fix auction rules routes

jessieay:
* Were pointing to wrote file path
* Also make user#edit path more rails-y
* Remove duplicate view template


harrisj:
Excellent. I like cleaning up unused views and partials. Nice

PR #762, opened 2016-06-23
Use admin nav in all admin views

jessieay:
* Use admin layout so we aren't tracking google analytics in admin
* Update flash messages styles to be more slim

Before:
screen shot 2016-06-23 at 1 58 45 pm

After:

screen shot 2016-06-23 at 1 58 02 pm


adelevie:
LGTM!

PR #768, opened 2016-06-24
Move highlighted_bid logic to Rules classes

harrisj:
I noticed this while I was working on the great renaming refactor. It seems like this logic is better located within the Rules objects, so we can avoid having to look at the model type in other classes.

### DIFF
--- a/app/models/highlighted_bid.rb
+++ b/app/models/highlighted_bid.rb
@@ -7,36 +7,12 @@ class HighlightedBid
   end
 
   def find
-    if !auction.bids.any?
-      NullBid.new
-    elsif available_sealed_bid? && user_is_bidder?
-      lowest_user_bid
-    elsif available_sealed_bid?
-      NullBid.new
-    else
-      auction.lowest_bid
-    end
+    auction_rules.highlighted_bid(user)
   end
 
   private
 
-  def available_sealed_bid?
-    available? && auction.type == "sealed_bid"
-  end
-
-  def available?
-    AuctionStatus.new(auction).available?
-  end
-
-  def user_is_bidder?
-    user_bids.any?
-  end
-
-  def lowest_user_bid
-    user_bids.first
-  end
-
-  def user_bids
-    auction.bids.where(bidder: user).order(amount: :asc)
+  def auction_rules
+    RulesFactory.new(auction).create
   end
 end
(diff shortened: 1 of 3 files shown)


adelevie:
Looks great!

harrisj:
So, what do you think we should do for something like this in `AuctionListitem`:

``` ruby
  def winning_bid_partial
    if over? && !auction.bids.any?
      'auctions/no_bids'
    elsif auction.type == 'reverse' && over?
      'auctions/over_winning_bid_details'
    elsif auction.type == 'sealed_bid' && over?
      'auctions/over_with_bids'
    elsif auction.type == 'reverse' && available?
      auction_available_bids_partial
    elsif auction.type == 'sealed_bid' && available?
      sealed_bid_bidder_partial
    else # future
      'components/null'
    end
  end
```

I would like to eliminate the dispatch on `auction.type`. It seems like we have a few options.

1. Make this a method that is defined in the Rules classes via polymorphism. I'm not sure though that I want the Rules classes to know anything about partials.
2. Create a parallel to the Rules classes inside `view_models`. So, we'd define `ViewModel::Rules::SingleBidAuction` and put the logic for this method in there. What I don't like about this is that we have two separate classes which need to be coupled together. So you'd have to do something like this `RulesFactory.new(auction).view_model` to get the correct view model (with its own parallel factory constructor to `RulesFactory#create`)
3. We could generalize the tests here into named functions like `show_winning_bid_details?` that we have in the Rules object. To be honest though, I'm not sure why there is a difference btween reverse and sealed auctions here when the auction is over, so naming these methods might be hard.

Thoughts?

adelevie:
LGTM

jessieay:
Hey @harrisj ! this is awesome. Re: question above -- maybe delegate to a presenter for each rule type that deals with this view-related logic? similar to how we deal with auction status stuff....

also, it may be worth it to go over all of the above branching with @mtorres253 . When I did all of those refactors a few weeks ago a made a bunch of huge conditionals like that one, which revealed that in some places our logic is probably more complex / nuanced than it needs to be. Mapping out a consistent set of conditionals for something like the method you described above might allow is to simplify without extracting other objs!

harrisj:
@jessie: I think this is an excellent course of action. Let's review all the conditional branching here before we start refactoring.

Issue #707, opened 2016-06-13
Winners page shows total number of small businesses


mtorres253:
As an admin user, I want to keep track of the total number of vendors who are small businesses, so that I can track of the proportion of small businesses that make up the total number of qualified vendors. As a KPI, I would like this to be a large percentage

Scenario: I see the total numer of Sam.gov qualified vendors on the winners page
Given I am viewing the winners page
Then I should see a section labeled "Small Businesses"
And in that section I should see the total number of vendors in the system who are marked as small businesses in Sam.gov

PR #769, opened 2016-06-24
Winners page shows total number of small businesses

jessieay:
* Use Ruby to calculate stats - js is more complicated in this case
* Closes https://github.com/18F/micropurchase/issues/707


adelevie:
Awesome!

PR #782, opened 2016-06-27
Use 18F-wide SCSS linter

jessieay:
* See https://pages.18f.gov/frontend/#css-linting

### DIFF
--- a/.scss-lint.yml
+++ b/.scss-lint.yml
@@ -1,253 +1,145 @@
-# Default application configuration that all configurations inherit from.
-
-scss_files: "**/*.scss"
-plugin_directories: ['.scss-linters']
-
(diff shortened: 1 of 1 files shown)


jessieay:
Let's ignore violations for now and fix as we go

harrisj:
So, you think we should avoid the 42 violations and fix them later. I'm fine with that

jessieay:
@harrisj yeah I feel like it would be pretty time intensive to fix all our existing CSS, especially deeply nested stuff. But we should def update as we go along and touch files!

harrisj:
I think Hound suggested against correcting all Rubocop issues because you could create new errors in the process and instead just fixing things as you go along. I agree too. Merging.

Issue #705, opened 2016-06-13
Winners page shows average time to delivery


mtorres253:
As a product team member, I want to keep track of the time to delivery across auctions, so that I can have some sense of the effectiveness of the platform with respect to how quickly code gets delivered

Scenario: I see the average delivery time for all auctions on the winners page
Given I am viewing the winners page
Then I should see in the "Insights" section the average "Time to Delivery" for set of auctions featured on that page
And the time to delivery should be in days (if it's less than 24 hours, then it counts as one day)

Note: Time to delivery is the time it takes from when the auction ends to when the auction is accepted

PR #783, opened 2016-06-27
Set `accepted_at` when an auction is accepted

jessieay:
* Show timestamp on admin/auctions#show
* This data will help us with our stats page
* Part of https://github.com/18F/micropurchase/issues/705

### DIFF
--- a/db/migrate/20160627225519_add_accepted_at_to_auctions.rb
+++ b/db/migrate/20160627225519_add_accepted_at_to_auctions.rb
@@ -4,7 +4,8 @@ class AddAcceptedAtToAuctions < ActiveRecord::Migration
 
     execute <<-SQL
       UPDATE auctions
-      SET accepted_at = delivery_due_at;
+      SET accepted_at = delivery_due_at
+      WHERE result = 1;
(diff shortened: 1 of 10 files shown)


db/migrate/20160627225519_add_accepted_at_to_auctions.rb
```diff
@@ -0,0 +1,14 @@
+class AddAcceptedAtToAuctions < ActiveRecord::Migration
+  def up
+    add_column :auctions, :accepted_at, :datetime
+
+    execute <<-SQL
+      UPDATE auctions
+      SET accepted_at = delivery_due_at;
```
harrisj: Does this set accepted_at even for auctions that were not accepted? Maybe add a `where accepted = true`

jessieay: good call!

jessieay: updated at dfab52b

harrisj:
Perfect. Let's merge!

PR #784, opened 2016-06-27
Refine calculations on winners page

jessieay:
* Some calculations are related to *all* published auctions, others are
  are related to only completed or successful auctions
* Add test to fix logic in unique bidders method

### DIFF
--- a/app/view_models/winners_view_model.rb
+++ b/app/view_models/winners_view_model.rb
@@ -8,38 +8,40 @@ class WinnersViewModel
   end
 
   def total_auctions
-    auction_count
+    published_auction_count
   end
 
   def unique_auction_winners
-    completed_auctions.map do |auction|
+    successful_auctions.map do |auction|
       WinningBid.new(auction).find.bidder
     end.uniq.count
   end
 
   def average_bids_per_auction
-    if auction_count > 0
-      completed_auctions.map(&:bids).flatten.count / auction_count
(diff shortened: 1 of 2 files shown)


app/view_models/winners_view_model.rb
```diff
@@ -8,38 +8,40 @@ def upcoming_count
...
+    if successful_auction_count > 0
+      successful_auctions.map(&:bids).flatten.count / successful_auction_count
     else
       'n/a'
     end
   end
 
   def vendors_with_bids_count
-    User.includes(:bids).where(bids: { bidder_id: nil }).count
+    User.includes(:bids).where.not(bids: { bidder_id: nil }).count
```
jessieay: @adelevie 👀

adelevie: Hrmm ... what about getting all bids, map to vendor, then uniq? Just thinking out loud.

jessieay: This way of finding records without a nil `has_many` seems to be pretty standard; seems like it would be faster than iterating over an array since it is AR (compiles to SQL) http://stackoverflow.com/questions/35508289/rails-only-find-records-where-has-many-associated-records-are-present

I added a test to confirm that the old query was incorrect. Do you think it is still unclear?

adelevie: all good!

Issue #770, opened 2016-06-27
Email sent to vendor if credit card URL is not on file when auction is accepted


mtorres253:
As a vendor I want to be apprised if my work was accepted and I cannot be paid, so the system should do that for me

Scenario: Vendor receives email asking them to add a credit card URL to their profile
Given I am a registered vendor
And there is an auction that I have won and delivered
And I do not have a payment URL specified in my profile
When the auction is accepted (admin sets the result to "accepted")
Then the system sends an email to the vendor
And no payment action is taken
And the email should have the following values
From: micropurchase@gsa.gov
To: [vendor email]
Subject: Please update your profile so Micro-purchase can pay you for your great work!
Dear [vendor name]

A payment of $[amount] is pending release regarding work you recently completed on [auction name]. In order to take receipt of payment, you will need to first add a credit card URL to your profile. Please go to  https://micropurchase.18f.gov/profile to update this information. 

Questions? Please contact us at microupurchase@gsa.gov

"Sincerely,"

"The 18f Micropurchase team"

Note: This story includes changing https://micropurchase.18f.gov/edit_user to https://micropurchase.18f.gov/profile

adelevie:
Doing the following refactoring as a pre-cursor:

(not truly a feature spec since this describes application behavior, but not user-facing behavior)

When an `auction.end_time` is created, a delayed job task is queued up, for that time, which invoked a generic job called something like `AuctionEnded.new(auction).perform`.

When an `auction.end_time` is updated, the corresponding job is updated to run at the new `.end_time`.

When an `auction.end_time` is deleted (or an entire auction is), the corresponding job is deleted.

adelevie:
Following this, we could develop a somewhat standard pattern, with a class called `AuctionAccepted`, which would wrap all of the tasks that need to happen when an auction is accepted.

mtorres253:
Changed from address to be consistent with other emails - admin@micropurchase.gov

adelevie:
Wait we don't own the micropurchase.gov domain.

jessieay:
@mtorres253 after user updates credit card URL, I assume *that* is when the purchase request should get updated? 

Right now we update purchase request in C2 as soon as accepted. In that update we include winning bidder details, including credit card url (see /18F/micropurchase/blob/develop/app/services/update_c2_attributes.rb) 

This story was written before when we created the purchase request after acceptance so it refers to not creating the purchase request after the credit card url is updated.

mtorres253:
[...]

PR #793, opened 2016-06-28
Renamed BuildAuction to CreateAuction

adelevie:
Created and implemented SaveAuction

Groundwork for the groundwork for https://github.com/18F/micropurchase/issues/770

### DIFF
--- a/app/controllers/admin/auctions_controller.rb
+++ b/app/controllers/admin/auctions_controller.rb
@@ -20,9 +20,9 @@ class Admin::AuctionsController < Admin::BaseController
   end
 
   def create
-    auction = CreateAuction.new(params, current_user).perform
+    auction = BuildAuction.new(params, current_user).perform
 
-    if auction.save
+    if SaveAuction.new(auction).perform
       flash[:success] = I18n.t('controllers.admin.auctions.create.success')
       redirect_to admin_auctions_path
     else
--- a/spec/services/build_auction_spec.rb
+++ b/spec/services/build_auction_spec.rb
@@ -1,0 +1,16 @@
+require 'rails_helper'
+
+describe BuildAuction do
+  describe '#perform' do
+    context 'building with all valid params' do
+      it 'builds the auction object' do
+        user = create(:user)
+        params = { auction: FactoryGirl.attributes_for(:auction) }
+
+        auction = BuildAuction.new(params, user).perform
+
+        expect(auction).to be_an(Auction)
+      end
+    end
+  end
+end
--- a/spec/services/save_auction_spec.rb
+++ b/spec/services/save_auction_spec.rb
@@ -1,0 +1,37 @@
+require 'rails_helper'
+
+describe SaveAuction do
+  describe '#perform' do
+    context 'given a valid auction' do
+      it 'saves the auction' do
+        auction = build(:auction)
+        SaveAuction.new(auction).perform
+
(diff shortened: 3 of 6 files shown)


spec/services/build_auction_spec.rb
```diff
@@ -0,0 +1,16 @@
+require 'rails_helper'
+
+describe BuildAuction do
+  describe '#perform' do
+    context 'building with all valid params' do
+      it 'builds the auction object' do
+        user = create(:user)
+        params = {auction: FactoryGirl.attributes_for(:auction)}
```
jessieay: I think the code climate settings we have ask for spaces between curly braces (doesn't analyze spec files but would be cool to have it "in style")

spec/services/save_auction_spec.rb
```diff
@@ -0,0 +1,37 @@
...
+
+        expect { SaveAuction.new(auction).perform }
+          .to change{auction.new_record?}.from(true).to(false)
+      end
+
+      it 'returns true' do
+        auction = build(:auction)
+        saved = SaveAuction.new(auction).perform
+
+        expect(saved).to be(true)
```
jessieay: thoughts on `expect(auction).to be_persisted` instead?

spec/services/save_auction_spec.rb
```diff
@@ -0,0 +1,37 @@
...
+      end
+    end
+
+    context 'given an invalid auction' do
+      it 'does not save the auction' do
+        auction = build(:auction, title: nil)
+
+        expect { SaveAuction.new(auction).perform }
+          .not_to change{auction.new_record?}
```
jessieay: thoughts one `not_to change { Auction.count } ` instead?

jessieay:
few 👶  comments from me - looking good!

spec/services/build_auction_spec.rb
harrisj: Yeah, that is a rule, but I'm not sure if we want that rule or not, but we can look at it later

adelevie:
fixed the issues, rebased

app/controllers/admin/auctions_controller.rb
```diff
@@ -20,9 +20,10 @@ def new
   end
 
   def create
-    auction = CreateAuction.new(params, current_user).perform
+    auction = BuildAuction.new(params, current_user).perform
+    saved = SaveAuction.new(auction).perform
 
-    if auction.save
+    if saved
```
jessieay: why not inline `SaveAuction.new(auction).perform` here rather than add var?

spec/services/save_auction_spec.rb
```diff
@@ -0,0 +1,44 @@
...
+
+    context 'given an invalid auction' do
+      it 'does not save the auction' do
+        auction = build(:auction, title: nil)
+
+        expect { SaveAuction.new(auction).perform }
+          .not_to change { Auction.count }
+      end
+
+      it 'is not persisted' do
```
jessieay: it seems like this test and the one above are testing the same thing -- keep one?

jessieay:
couple more baby comments -- also I think the commit message should be reversed (renamed to `Build` not `Create`)

adelevie:
@jessieay fixed, rebased, and edited the commit msg

jessieay:
lovely! merging

Issue #773, opened 2016-06-27
Admin sees all data on auction show page


mtorres253:
As an admin user, I want to see all information about an auction on the auction show page, so that I don't have to go to the edit page.

Scenario: Admin sees <data_point> on auction show page
Given I am logged in as an admin
And there is a given auction
When I go to the show page for that auction
Then in addition to what I already see
I should see a section labeled <data_label>
And in that a section I should see the value for <data_point> if there is one

| Data Point | Data Label |
| :---- | :---- |
| auction_type | Auction Type |
| auction_github_repo | Github repo URL |
| auction_issue_url | Github issue URL |
| auction_delivery_due_at | Delivery deadline date and time |
| auction_start_price | Starting price |
| auction_purchase_card | Purchase card |
| auction_cap_proposal_url | Cap proposal URL |
| auction_published | Published status |
| auction_delivery_url | Delivery URL |

Note: Use the same layout as the vendor view. Put the admin specific stuff below the vendor stuff in the right rail (gray box)

jessieay:
should we add an edit link here, too?

PR #796, opened 2016-06-28
Admin sees all data on admin auctions#show page

jessieay:
* Use same layout as regular admins#show
* Closes https://github.com/18F/micropurchase/issues/773

screen shot 2016-06-28 at 4 20 56 pm

### DIFF
--- a/.codeclimate.yml
+++ b/.codeclimate.yml
@@ -6,6 +6,7 @@ engines:
     exclude_fingerprints:
     - 508b75e787610c57dc9780baf6be5d74d253df8af35b5c861d6aacc911901a56
     - 8d8bc90f85c52f9aa9cb5be01fb072fc42aa260520c2d8c0adaf6fc8080546cf
+    - 38eaf3a4c38ed941e5397a84c9f95b91a7c3dee9f70c0993131dd3be89180f75
   eslint:
     enabled: true
   csslint:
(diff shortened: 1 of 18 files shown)


app/presenters/status_presenter/available.rb
```diff
@@ -11,10 +11,6 @@ def relative_time
     "Time remaining: #{HumanTime.new(time: auction.ended_at).distance_of_time_to_now}"
   end
 
-  def status_text
```
harrisj: Glad to get rid of these

.codeclimate.yml
```diff
@@ -6,6 +6,7 @@ engines:
     exclude_fingerprints:
     - 508b75e787610c57dc9780baf6be5d74d253df8af35b5c861d6aacc911901a56
     - 8d8bc90f85c52f9aa9cb5be01fb072fc42aa260520c2d8c0adaf6fc8080546cf
+    - 38eaf3a4c38ed941e5397a84c9f95b91a7c3dee9f70c0993131dd3be89180f75
```
harrisj: Is this a CodeClimate issue?

harrisj:
This looks good, but I was wondering if we should get rid of `preview` now?

adelevie:
> This looks good, but I was wondering if we should get rid of preview now?
I *think* so.

adelevie:
🐲

.codeclimate.yml
jessieay: yeah we were getting a brakeman error for linking to `issue_url` because it is "user input"

jessieay:
@harrisj @adelevie  the benefit of preview is being able to see non-published auctions as they will look to the vendors. Admin view added in this PR is *similar* to the vendor view, but has a few key differences. Not sure how much preview is being used -- @mtorres253 ?

PR #797, opened 2016-06-29
WIP: Send emails when auction ends

adelevie:
~~Only merge after https://github.com/18F/micropurchase/pull/793 has been merged.~~

Groundwork for #770.

<hr />

This PR significantly changes how we send emails when auctions end. Previously, we used a cron job that would check for recently expired auctions and send batches of emails once a day. This worked well enough, but it leaves open a risk that we might miss an auction depending on when it ended (e.g. it could end right after the cutoff time for the query used by the cron job).

This PR schedules a Delayed Job when an auction is created (and `auction.ended_at` is set). The job is scheduled to run at `auction.ended_at` and to run `AuctionEnded#perform`. So if you need anything to run when an auction ends, be it an email to vendors, an email to clients, or an internal alert email, just stick the code in that `#perform` method, and don't worry about queuing/managing a Delayed Job. This PR sets up that infrastructure.

### DIFF
--- a/app/models/auction_ended.rb
+++ b/app/models/auction_ended.rb
@@ -1,0 +1,14 @@
+class AuctionEnded
+  def initialize(auction)
+    @auction = auction
+  end
+
+  def perform
+    WinningBidderEmailSender.new(auction).perform
+    LosingBidderEmailSender.new(auction).perform
+  end
+
+  private
+
+  attr_reader :auction
+end
--- a/app/services/save_auction.rb
+++ b/app/services/save_auction.rb
@@ -7,7 +7,8 @@ class SaveAuction
     saved = auction.save
 
     if should_schedule_auction_ended_job?(saved)
-      AuctionEnded.new(auction)
+      AuctionEnded
+        .new(auction)
         .delay(run_at: auction.ended_at)
         .perform
     end
--- a/app/services/save_auction.rb
+++ b/app/services/save_auction.rb
@@ -9,7 +9,7 @@ class SaveAuction
     if should_schedule_auction_ended_job?(saved)
       AuctionEnded
         .new(auction)
-        .delay(run_at: auction.ended_at)
+        .delay(run_at: auction.ended_at + 5.minutes)
         .perform
     end
 
--- a/app/services/save_auction.rb
+++ b/app/services/save_auction.rb
@@ -4,10 +4,22 @@ class SaveAuction
   end
 
   def perform
-    auction.save
+    saved = auction.save
+
+    if should_schedule_auction_ended_job?(saved)
+      AuctionEnded.new(auction)
+        .delay(run_at: auction.ended_at)
+        .perform
+    end
+
+    saved
   end
 
   private
 
   attr_reader :auction
+
+  def should_schedule_auction_ended_job?(saved)
+    saved && !auction.ended_at.nil?
+  end
 end
--- a/app/services/save_auction.rb
+++ b/app/services/save_auction.rb
@@ -15,12 +15,9 @@ class SaveAuction
   attr_reader :auction
 
   def schedule_auction_ended_job(saved)
-    if should_schedule_auction_ended_job?(saved)
-      AuctionEnded.new(auction).delay(run_at: auction.ended_at).perform
+    if saved
+      job = AuctionEndedJob.new(auction.id)
+      Delayed::Job.enqueue(job, run_at: auction.ended_at)
     end
-  end
-
-  def should_schedule_auction_ended_job?(saved)
-    saved && !auction.ended_at.nil?
   end
 end
--- a/app/services/save_auction.rb
+++ b/app/services/save_auction.rb
@@ -5,13 +5,7 @@ class SaveAuction
 
   def perform
     saved = auction.save
-
-    if should_schedule_auction_ended_job?(saved)
-      AuctionEnded
-        .new(auction)
-        .delay(run_at: auction.ended_at + 5.minutes)
-        .perform
-    end
+    schedule_auction_ended_job(saved)
 
     saved
   end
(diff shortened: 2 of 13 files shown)


app/models/auction_ended.rb
```diff
@@ -0,0 +1,14 @@
+class AuctionEnded
```
jessieay: seems like this should be a service obj rather than regular model? (given `perform` method)

app/services/save_auction.rb
```diff
@@ -4,10 +4,23 @@ def initialize(auction)
...
   attr_reader :auction
+
+  def schedule_auction_ended_job(saved)
+    if should_schedule_auction_ended_job?(saved)
+      AuctionEnded.new(auction).delay(run_at: auction.ended_at).perform
+    end
+  end
+
+  def should_schedule_auction_ended_job?(saved)
+    saved && !auction.ended_at.nil?
```
jessieay: I believe `ended_at` is a required field so should never be nil

app/services/save_auction.rb
```diff
@@ -4,10 +4,23 @@ def initialize(auction)
...
+    schedule_auction_ended_job(saved)
+
+    saved
   end
 
   private
 
   attr_reader :auction
+
+  def schedule_auction_ended_job(saved)
```
jessieay: great method name! 👏

features/admin_creates_auction.feature
```diff
@@ -2,6 +2,7 @@ Feature: Admin creates auctions in the admins panel
   As an admin
   I should be able to create auctions
 
+  @background_jobs_off
```
jessieay: why are we turning these off?

jessieay:
this is looking great to me!

app/models/auction_ended.rb
adelevie: I like `AuctionEndedJob`. It's the generic home to any and all tasks that need to happen when the auction ends. Perhaps it could even call an `AuctionEndedEmailSender` object.

app/services/save_auction.rb
adelevie: Is it worth removing the `if` clause? Still might be a nice check to have in place.

features/admin_creates_auction.feature
adelevie: when we create an auction is schedules a job for far into the future--a future where the auction has ended and where there are winning and losing vendors. if we keep background jobs on, it gets invoked immediately in the feature spec, and looks for a winning/losing vendor that doesn't yet exist (since the feature only is concerned with auction creation, not ending)

app/services/save_auction.rb
jessieay: postgres is doing that check for us! If anyone tries to create an auction an `ended_at` timestamp, postgres  and our AR validation will not let us get to this point

features/admin_creates_auction.feature
jessieay: oy! well maybe we should also make sure the logic in the job doesn't blow up if an auction ends without a winner or losers? can imagine we may someday have an auction with zero bids...

app/services/save_auction.rb
adelevie: alright, I'll remove the referenced to `ended_at`.

features/admin_creates_auction.feature
adelevie: good point ... I think a simple `if` statement or two would solve this.

features/contracting_officer_creates_auction.feature
```diff
@@ -2,6 +2,7 @@ Feature: Contracting Officers
   As an admin and contracting officer
   I should be able to create auctions over $3500
 
+  @background_jobs_off
```
jessieay: still need this?

features/step_definitions/auction_create_steps.rb
```diff
@@ -10,6 +10,22 @@
   @auction = FactoryGirl.create(:auction, :closed, :with_bidders)
 end
 
+Given(/^I have won an auction$/) do
+  @auction = FactoryGirl.build(:auction, :closed, :with_bidders)
+  bids = @auction.bids.sort_by(&:created_at).reverse
+  b = bids.first
```
jessieay: maybe we should ensure that the bid is the winning bid rather than the first?

features/step_definitions/auction_create_steps.rb
```diff
@@ -10,6 +10,22 @@
...
+  b.update_attribute(:bidder_id, @user.id)
+  SaveAuction.new(@auction).perform
+end
+
+Given(/^I have lost an auction$/) do
+  @auction = FactoryGirl.build(:auction, :closed, :with_bidders)
+  bids = @auction.bids.sort_by(&:created_at).reverse
+  b = bids.last
```
jessieay: maybe we should ensure the bid is the losing bid rather than last?

jessieay:
this looks great!

do you think it makes sense to introduce a feature spec that does not send background jobs immediately but travels in time and looks for a sent email? Would be nice to cover this new functionality at the feature level if we can...

features/vendor_views_closed_auction.feature
```diff
@@ -4,29 +4,25 @@ Feature: Vendor views closed auctions
...
-    And I have placed the lowest bid
+    Given I am an authenticated vendor
+    And I have won an auction
     When I visit the auction page
     Then I should see the auction had a winning bid with name
     And I should see I am the winner
     And I should see when the auction started
     And I should see when the auction ended
     And there should be meta tags for the closed auction
-    And an email notification is sent to the winning bidder
```
adelevie: weird to me how the tests pass even without this step. Definition for it:

```ruby
Then(/^an email notification is sent to the winning bidder$/) do
  WinningBidderEmailSender.new(@auction).perform
end
```

jessieay: I think that happens because delayed jobs run immediately in the test env. so in order to not have that happen, we'd set them up to *not* run immediately in a test

adelevie: So turn off background jobs, then use time cop to travel forward, then manually work off the jobs?

jessieay: I don't think turn off, just don't have them run right away (which is default) -- not sure if this is possible

adelevie:
How do I re-open this? Halp.

jessieay:
I think you might need to create a new branch and cherry pick these via commit sHA :(

PR #804, opened 2016-06-29
Update "average delivery" stat

jessieay:
* Use `accepted_at` timestamp instead of `delivery_due_at`


adelevie:
@jessieay how would this affect previous auctions where the `accepted_at` data wasn't recorded?

jessieay:
@adelevie in the migration we set all of those to be the same as the `delivery_due_at`

adelevie:
nice!

Issue #666, opened 2016-06-03
System can store Customer information


mtorres253:
As an admin user, I want to have an automated way to keep track of which customer agency is sponsoring which auction. So, the system should give me a way to store this information.

Scenario: Admin sees a new customer agency form
Given I am a logged in admin
When I go to https://micropurchase.18f.gov/admin/clients/new
Then I should see a form labeled "Create a New Customer Agency"
And there should be a "Customer Agency Name" text field
And there should be a "Contact Name" text field
And there should be a "Contact Email" text field
And there should be a "Create Customer Agency" button

Context:
Given I am a logged in admin user
And I am on the "Create Customer Agency" form https://micropurchase.18f.gov/customer/new

Scenario: I save a client without specifying a name
And the Customer Name field is blank 
When I click the "Create Customer Agency" button
Then the client should not be created
And I should stay on the same page
And I should see an error "The Customer Agency Name is required"

Scenario: I save a client without specifying a contact name
And the Contact Name field is blank 
When I click the "Create Customer" button
Then the client should not be created
And I should stay on the same page
And I should see an error "A contact name is required"

Scenario: I try to create Customer that already exists
And the Customer Name field has a value
And there is existing client with the same name
When I click the "Create Customer" button
Then the client should not be created
And I should stay on the same page
And I should see an error "A Customer with that name already exists"

Scenario: I successfully create a new Customer
And the Customer Name field has a value
When I click the Create Customer button
Then the customer should be created
And I should go back to the customers index page https://micropurchase.18f.gov/admin/customers
And I should see a success message "You have successfully created [Customer Name] as a new customer"
And I should see the new customer listed there

Issue #668, opened 2016-06-03
Customer index page


mtorres253:
As an admin, I want to see a list of customers currently in the system, so there should be an index page listing them

Scenario: I see existing customer agencies on customers page
Given I am a logged in admin
And I'm on the admin homepage
And there are clients in the system
Then I should see a link to "Customer Agency"

When I click on that links I should go to https://micropurchase.18f.gov/admin/customers
And I should see a header "Customer Agency"
And I should see the clients listed alphabetically
And I should see the customer agency name, the contact name and the contact email

jessieay:
Closed via #809

PR #809, opened 2016-06-30
Add customer model

harrisj:
This adds the Customer model and an admin controller for creating and editing customers. Note that I edited some of the features in the story to be more consistent with other features. I also reorganized the features slightly.

Fixes: #666, #668


adelevie:
🌵

PR #814, opened 2016-07-01
Add bidding history to user show page in admin

### LINKED ISSUE #775: Bid history shows on user page
As an admin user, I want to easily see the bid history for a vendor, so that I can can have some visibility into how the vendor is engaging with the micro-purchase platform.

Scenario: I see a user's bid history on the user show page
Given I am logged in as an admin
And there is a vendor in the system
And there have been auctions
And the vendor has participated in the auctions
When I go to the show page for that vendor
Then I should see a section labeled "Bid History"
And in that section I should see a table
And there should be a column labeled <bid_history_column_name>
And in that column I should see <bid_history_value>

| Bid History Column Name | Bid History Value |
| :----- | :----- |
| Auction | URL to auction show page |
| Auction Start Date | Start date of auction |
| Auction End Date | End date of auction |
| Number of Bids | The number of bids that vendor made on the auction |
| Won Auction? | Boolean showing whether vendor won the auction |
| Successful Delivery? | A Boolean showing whether the vendor successfully delivered on that auction |

harrisj:
Fixes #775

Generally I have tried to amend existing objects/methods where possible to add this new functionality. There are a few things worth mentioning that should be singled out for review:

1. I needed a good wrapper for displaying user-specific information about an auction (such as the user's bid count for the auction). I created a new class `Admin::UserAuctionViewModel` in the view_models, but I'm not too thrilled at my name for it.
2. I added a specific step definition file `admin_views_user_steps` because I had a step definition that was specific to that one feature file and this is the convention I've proposed for Cucumber steps are only linked to a specific feature.
3. The AuctionQuery object had defined a `my_bids` query, but it wasn't used anywhere in the code that I could see. So I renamed it and used it here (as well as added a test). This would be useful if we bring back the my_bids endpoint too.

### DIFF
--- a/app/models/auction_query.rb
+++ b/app/models/auction_query.rb
@@ -98,7 +98,7 @@ class AuctionQuery
       .find(id)
   end
 
-  def user_has_bid(user_id)
+  def with_bid_from_user(user_id)
     @relation
       .joins(:bids)
       .where(bids: { bidder_id: user_id })
--- a/app/models/auction_query.rb
+++ b/app/models/auction_query.rb
@@ -98,7 +98,7 @@ class AuctionQuery
       .find(id)
   end
 
-  def my_bids(user_id)
+  def user_has_bid(user_id)
     @relation
       .joins(:bids)
       .where(bids: { bidder_id: user_id })
--- a/app/presenters/dc_time_presenter.rb
+++ b/app/presenters/dc_time_presenter.rb
@@ -13,8 +13,8 @@ class DcTimePresenter
     new(time).convert
   end
 
-  def self.convert_and_format(time, format = FORMAT)
-    new(time).convert_and_format(format)
+  def self.convert_and_format(time, format = FORMAT, timezone_label: true)
+    new(time).convert_and_format(format, timezone_label: timezone_label)
   end
 
   def self.time_zone
@@ -26,9 +26,11 @@ class DcTimePresenter
     time.in_time_zone(time_zone)
   end
 
-  def convert_and_format(format = FORMAT)
+  def convert_and_format(format = FORMAT, timezone_label: true)
     if time
-      convert.strftime(format) + " #{timezone_label}"
+      str = convert.strftime(format)
+      str += " #{timezone_string}" if timezone_label
+      str
     else
       NULL
     end
@@ -38,7 +40,7 @@ class DcTimePresenter
     self.class.time_zone
   end
 
-  def timezone_label
+  def timezone_string
     convert.zone
   end
 end
--- a/app/view_models/admin/user_auction_view_model.rb
+++ b/app/view_models/admin/user_auction_view_model.rb
@@ -18,11 +18,11 @@ class Admin::UserAuctionViewModel
   end
 
   def start_date
-    DcTimePresenter.convert_and_format(auction.started_at, DATE_FORMAT, timezone_label: false)
+    DcTimePresenter.convert_and_format(auction.started_at, DATE_FORMAT)
   end
 
   def end_date
-    DcTimePresenter.convert_and_format(auction.ended_at, DATE_FORMAT, timezone_label: false)
+    DcTimePresenter.convert_and_format(auction.ended_at, DATE_FORMAT)
   end
 
   def user_bid_count
--- a/app/view_models/admin/user_auction_view_model.rb
+++ b/app/view_models/admin/user_auction_view_model.rb
@@ -2,7 +2,7 @@ class Admin::UserAuctionViewModel
   attr_reader :auction, :user
 
   DATE_FORMAT = '%m/%d/%Y'.freeze
-  NA_RESPONSE_STRING = '-'
+  NA_RESPONSE_STRING = '-'.freeze
 
   def initialize(auction, user)
     @auction = auction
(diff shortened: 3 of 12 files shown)


app/models/auction_query.rb
```diff
@@ -98,7 +98,7 @@ def public_find(id)
       .find(id)
   end
 
-  def my_bids(user_id)
+  def user_has_bid(user_id)
```
jessieay: maybe `user_bids` would be a more apt name for this given that it returns a collection of bids for a user?

app/view_models/admin/user_show_view_model.rb
```diff
@@ -48,4 +48,12 @@ def contracting_officer
       "No"
     end
   end
+
+  def bids?
+    user_auctions.count > 0
+  end
+
+  def user_auctions
+    AuctionQuery.new.user_has_bid(user.id).map { |auction| Admin::UserAuctionViewModel.new(auction, user) }
```
jessieay: nice!

app/views/admin/users/show.html.erb
```diff
@@ -9,3 +9,31 @@
   <li><b>Small business</b>: <%= @view_model.small_business %></li>
   <li><b>Contracting officer</b>: <%= @view_model.contracting_officer %></li>
 </ul>
+
+<% if @view_model.bids? %>
```
jessieay: let's use a partial instead! (in view model)

jessieay: eg: https://github.com/18F/micropurchase/blob/develop/app/view_models/auctions_index_view_model.rb#L31

jessieay:
code climate error seems fixable!

app/view_models/admin/user_auction_view_model.rb
```diff
@@ -0,0 +1,77 @@
...
+  def title
+    auction.title
+  end
+
+  def id
+    auction.id
+  end
+
+  def start_date
+    DcTimePresenter.convert_and_format(auction.started_at, DATE_FORMAT, timezone_label: false)
```
jessieay: why not just show this in ET?

app/presenters/dc_time_presenter.rb
```diff
@@ -26,9 +26,11 @@ def convert
     time.in_time_zone(time_zone)
   end
 
-  def convert_and_format(format = FORMAT)
+  def convert_and_format(format = FORMAT, timezone_label: true)
```
jessieay: why would we not want to show timezone label? seems potentially confusing

app/view_models/admin/user_auction_view_model.rb
```diff
@@ -0,0 +1,77 @@
...
+    DcTimePresenter.convert_and_format(auction.ended_at, DATE_FORMAT, timezone_label: false)
+  end
+
+  def user_bid_count
+    user_bids.count
+  end
+
+  def user_won_label
+    if auction_over?
+      if user_won?
```
jessieay: looks like `user_won?` already checks for `auction_over?`

app/views/admin/users/show.html.erb
```diff
@@ -9,3 +9,31 @@
...
+
+<table class="usa-table-borderless" id="bidding-history">
+  <thead>
+    <tr>
+      <th scope="col">Auction</th>
+      <th scope="col">Start Date</th>
+      <th scope="col">End Date</th>
+      <th scope="col">User Bids</th>
+      <th scope="col">User Won?</th>
+      <th scope="col">Accepted?</th>
```
jessieay: Looks like these labels are slightly different from the ones in the original issue -- cool with me if you think these are clearer, just wanted to point that out in case it was an oversight

app/view_models/admin/user_auction_view_model.rb
```diff
@@ -0,0 +1,77 @@
...
+      else
+        'No'
+      end
+    else
+      NA_RESPONSE_STRING
+    end
+  end
+
+  def accepted_label
+    if user_won?
```
jessieay: this has more duplication but might be easier to read:

```
if user_won? && auction_accepted?
  'Yes'
elsif user_won?
  'No'
else
  NA_RESPONSE_STRING
end
```

app/models/auction_query.rb
harrisj: It actually returns a collection of auctions where the user has placed a bid, so I struggled with the name a bit

app/views/admin/users/show.html.erb
harrisj: Oh good point!

app/view_models/admin/user_auction_view_model.rb
harrisj: I am doing this so I will show it in ET (even though the started_at is stored in UTC)

app/presenters/dc_time_presenter.rb
harrisj: I figured since this is an admin page it would be a little annoying to see 6/1/2016 EST and such over and over again

app/views/admin/users/show.html.erb
harrisj: You are correct. I think this is clearer, but I'm willing to hear it out

app/views/admin/users/show.html.erb
```diff
@@ -9,3 +9,5 @@
   <li><b>Small business</b>: <%= @view_model.small_business %></li>
   <li><b>Contracting officer</b>: <%= @view_model.contracting_officer %></li>
 </ul>
+
+<%= render partial: @view_model.bids_partial %>
```
jessieay: nice! I think you'll need to pass locals here too, right?

jessieay: oh wait I see it uses `@view_model` - this is super minor but I usually try to avoid instance vars in view partials (only use vars that are passed) - the benefit of that being that you only need to go to where something is rendered to know where the ivar comes from.

app/models/auction_query.rb
jessieay: yeah that is tricky....what about something like `with_bid_for_user(user_id)` too wordy?

app/presenters/dc_time_presenter.rb
jessieay: I think time zone is pretty helpful in admin. otherwise not sure if it is UTC or ET

harrisj: Okat, I added it back

PR #815, opened 2016-07-01
Testing how to inherit from a remote Rubocop

harrisj:
**DO NOT MERGE** This is a test of how our local rubocop can pull from a remote one and testing to see if it works okay. Checking the following cases

- [x] rubocop command
- [ ] editor linters
  - [x] emacs
  - [ ] vi
  - [ ] atom
  - [ ] sublime-text 3
- [ ] CodeClimate (FAIL)


harrisj:
Okay, it doesn't seem to work with CodeClimate, so I will file a ticket there

harrisj:
Response from CodeClimate:
> Based on the error I'm seeing, it looks like the engine is trying to make a network request to fetch your `inherit_from` configuration. Unfortunately, we're currently not able to support that kind of configuration because it's pointing to a file that doesn't actually exist in your repository, it's not available to us on the server -- so it just throws an error. We do hope to eventually support importing base config files via Rubocop's `inherit_from` and `inherit_gem` config directive, but we haven't figured out a good way to do it yet that doesn't break aspects of the limits we place on engines for security reasons. As a workaround, you could always move the shared configuration within your repo and change the `inherit_from` declaration to be relative to the repo.

harrisj:
Since this doesn't yet work with CodeClimate, I am going to close this and not investigate this approach further until they implement this on their end.

PR #816, opened 2016-07-01
Bids on admin auction show page now link to user

harrisj:
I want the auction bid history in admin view to link to user pages, so this is a PR for that.

I am kinda tempted to drop the `veiled_` parity for the admin BidListItem objects because these will never be veiled and we shouldn't suggest they will be, but that's up for future discussion.


jessieay:
nice!

PR #818, opened 2016-07-05
Update capitalization on response for successful bid.

andrewmaier:
See: https://18f.slack.com/archives/micropurchase-dev/p1467398596000066
cc @mtorres253 @adelevie


jessieay:
LGTM !

Issue #771, opened 2016-06-27
Auction approval request to CAP


mtorres253:
As an admin user, I want to make sure that there is budget for a given auction before I make it public, so if there is a way for the system to alert the budget holder to the upcoming auction before it goes live, then the system should do that

Scenario: Admin sees "Request Auction Approval" button on create auction form
Given I am logged in as an admin
And I have successfully created an auction
And the auction uses the default payment method (CAP)
When I go to the edit form for that auction https://micropurchase.18f.gov/admin/auctions/[id]/edit
Then I should see a button labeled "Request Auction Approval"

Scenario: Notification to CAP that an auction will be published
Given I am logged in as an admin
And a draft auction exists
And the auction uses the default payment method (CAP)
When I click on the request approval button
Then a request is made to CAP
And the request should specify the start price of the auction
And I should see an indication that approval was requested

Scenario: Admin tries to publish an auction without having requested approval
Given I am logged in as an admin
And a draft auction exists
And the draft auction uses the default payment method (CAP)
And I have not requested approval for the auction
When I try to publish the auction
Then I should see an error "You must request approval for this auction before publishing"
And the auction should not be published

Note: let's talk about the MVP implementation of this feature

jessieay:
When creating an auction that uses the default P card you cannot set the auction to "published" until there is cap proposal url

the cap_proposal_url field is visible but not update-able

there is a link in the auction form called "create c2 proposal" that creates a proposal in C2

once the proposal is created, that link goes away

when the cap proposal is created, that field is filled in

when cap request is approved, the "publish" option is available

system checks for approval hourly

jessieay:
PR https://github.com/18F/micropurchase/pull/822

PR #822, opened 2016-07-05
Auction approval request to C2

jessieay:
* Closes https://github.com/18F/micropurchase/issues/771
* Admin can create C2 request via button
* Request approval is checked once per hour
* Only after request is approved can auction be published
* When auction is accepted, C2 proposal is updated with winner info

screen shot 2016-07-05 at 4 02 53 pm


app/services/construct_cap_attributes.rb
```diff
@@ -16,13 +16,11 @@ def perform
...
         product_name_and_description: product_name_and_description,
         justification: justification,
-        link_to_product: link_to_product,
         cost_per_unit: cost_per_unit,
         quantity: ConstructCapAttributes::QUANTITY,
         recurring: ConstructCapAttributes::RECURRING,
         date_requested: date_requested,
-        urgency: ConstructCapAttributes::URGENCY,
-        additional_info: additional_info
+        urgency: ConstructCapAttributes::URGENCY
```
harrisj: Is there a specific reason those fields were removed here? Not complaining, just wondering.

app/services/construct_cap_attributes.rb
```diff
@@ -41,33 +39,18 @@ def product_name_and_description
...
-    auction.delivery_url
-  end
-
   def justification
     "Tock line item: #{auction.billable_to}"
   end
 
   def cost_per_unit
-    winning_bid.amount
+    auction.start_price
```
harrisj: So we bill the auction start price, regardless of how much it actually sold for?

harrisj: Oh I see. We need to create a proposal and that is the maximum price possible...

app/services/update_auction.rb
```diff
@@ -56,27 +55,15 @@ def updating_ended_at?
   def perform_approved_auction_tasks
     if auction_accepted? && auction.accepted_at.nil?
       auction.accepted_at = Time.current
-    end
-
-    if should_create_cap_proposal?
-      CreateCapProposalJob.perform_later(auction.id)
+      UpdateCapProposalJob.perform_later(auction.id)
```
harrisj: Much nicer

app/services/update_cap_attributes.rb
```diff
@@ -0,0 +1,40 @@
...
+      }
+    }
+  end
+
+  private
+
+  attr_reader :auction
+
+  def cost_per_unit
+    winning_bid.amount
```
harrisj: Nice, I see

app/services/construct_cap_attributes.rb
harrisj: Okay, I see. You add these fields when we update the CAP proposal

harrisj:
This looks good, but it looks like it needs to be rebased. I also think we could get prettier buttons by moving them to a separate div and checking the font sizes (let's make them non-outlines too), but I don't know CSS

adelevie:
Looks great to me once rebase is done 🌲

app/services/construct_cap_attributes.rb
jessieay: yeah, this is part of the functionality change we needed. The highest price needs to be included for approval before the auction runs, but the proposal is updated with real bid amount for purchase step on C2. C2 tracks all changes (a comment is added to the proposal after each update) so the changes will be recorded on their side.

jessieay:
@harrisj rebased!

jessieay:
@adelevie rebased!

harrisj:
LGTM. Merging.

Issue #801, opened 2016-06-29
Total successful auctions show on winners page


mtorres253:
As a visitor to micropurchase, I'm interested in seeing the historical success rate of auctions run on the platform. So there should be a way for me to see the total number of successful auctions vs. the total number of auctions as well as the success rate

Scenario: Total number of successful auctions and success rate on winner's page
Given I'm visiting the micropurchase site
When I navigate to https://micropurchase.18f.gov/winners
Then I should see the total number of successful (accepted) auctions
And the total success rate of all auctions (accepted auctions/total number of auctions)

See this mockup
https://micropurchase.mybalsamiq.com/projects/redesignbidflow/prototype/Dashboard?key=7ce1f9e53f0776f57692c2ea4484d505c99922aa

PR #824, opened 2016-07-05
Show percent of successful auctions on insights page

jessieay:
* Closes https://github.com/18F/micropurchase/issues/801
screen shot 2016-07-05 at 4 44 11 pm


jessieay:
will clean this up tomorrow - moving calculations out of the view model so they are more easily tested!

harrisj:
Okay, I will hold up on review then

app/view_models/insights_view_model.rb
```diff
@@ -63,7 +139,9 @@ def accepted_auctions_count
     accepted_auctions.count
   end
 
-  private
+  def accepted_auction_percent
```
harrisj: So, this will dip when an auction is running or even when an auction is over but the delivery deadline has not been met. I wonder if we should change the denominator to be simply the total of accepted + rejected?

app/view_models/insights_view_model.rb
```diff
@@ -7,8 +7,84 @@ def upcoming_count
...
+      statistic: average_auction_length,
+      href: 'auction-length',
+      label: 'average auction length'
+    }
+  end
+
+  def average_winning_bid_stat
+    {
+      href: 'chart-winning-bid',
+      statistic: average_winning_bid,
```
harrisj: Sometimes I wonder if we should calculate the average decline of the winning bid from the starting price as a useful metric (since this will be skewed once we start with > 3500 starting amounts). Or maybe median winning bid or maybe percent decline. I dunno.

mtorres253: Yes, the more i think about it the more complicated/rich it becomes. We can
talk about this in grooming

On Tue, Jul 5, 2016 at 6:21 PM, Jacob Harris <notifications@github.com>
wrote:

> In app/view_models/insights_view_model.rb
> <https://github.com/18F/micropurchase/pull/824#discussion_r69663985>:
>
> > +      label: 'vendors who have placed bids'
> > +    }
> > +  end
> > +
> > +  def average_auction_length_stat
> > +    {
> > +      statistic: average_auction_length,
> > +      href: 'auction-length',
> > +      label: 'average auction length'
> > +    }
> > +  end
> > +
> > +  def average_winning_bid_stat
> > +    {
> > +      href: 'chart-winning-bid',
> > +      statistic: average_winning_bid,
>
> Sometimes I wonder if we should calculate the average decline of the
> winning bid from the starting price as a useful metric (since this will be
> skewed once we start with > 3500 starting amounts). Or maybe median winning
> bid or maybe percent decline. I dunno.
>
> —
> You are receiving this because you are subscribed to this thread.
> Reply to this email directly, view it on GitHub
> <https://github.com/18F/micropurchase/pull/824/files/a6aafe5ad12de9d1ecf3c10ec3a30ae5d3f831de#r69663985>,
> or mute the thread
> <https://github.com/notifications/unsubscribe/AAGHNWciwtkkdW10K_s04g-9zCEfd15Lks5qSwMigaJpZM4JFodZ>
> .
>



-- 
Michael Torres
Product Lead, Micropurchase Platform
18F
(415) 920-3205

app/view_models/insights_view_model.rb
jessieay: I like that!!

jessieay:
@harrisj ready!

app/models/auction_query.rb
```diff
@@ -47,9 +45,16 @@ def complete_and_successful
       .paid
   end
 
-  def rejected
+  def completed
```
harrisj: So this is completed_and_bid_upon I guess if we want to be pedantic, but that's a mouthful

app/models/average_auction_length.rb
```diff
@@ -0,0 +1,21 @@
+class AverageAuctionLength
```
harrisj: I wonder if we should put all these calculation models into a namespace like `Calculations` or `Statistics` to make it clear what they are for?

app/presenters/average.rb
```diff
@@ -0,0 +1,29 @@
+class Average
+  attr_reader :value, :count, :type
+
+  def initialize(value, count, type = nil)
+    @value = value
+    @count = count
+    @type = type
+  end
+
+  def to_s
```
harrisj: This is good

app/models/auction_query.rb
jessieay: I also felt like this name wasn't quite right....also not sure this is the exact query we should use for all the stats. the basic thing I was looking for in some was "over with bids" but on some cases just "over" would be fine....

app/models/average_auction_length.rb
jessieay: good idea...preference?

app/view_models/insights_view_model.rb
```diff
@@ -7,137 +7,110 @@ def upcoming_count
...
+  def unique_auction_winners
+    AuctionQuery.new.completed.map do |auction|
+      WinningBid.new(auction).find.bidder
+    end.uniq.count
   end
 
-  def published_auctions
-    @_published_auctions ||= AuctionQuery.new.published
+  def accepted_auction_percent
+    Percent.new(accepted_auctions_count, published_auction_count).to_s
```
harrisj: I still want this to be over `accepted_auctions_count + rejected_auctions_count` so it's just auctions we have ruled on

harrisj:
Looking good, just a few comments

app/view_models/insights_view_model.rb
jessieay: oh yes! sorry I forgot about that comment after it was hidden via doing these refactors...

jessieay: updated!

app/models/auction_query.rb
harrisj: I guess we are overthinking it. Let's call it completed since we already have a different notion of over....

app/view_models/insights_view_model.rb
```diff
@@ -82,140 +84,37 @@ def average_auction_length_stat
...
-    UserQuery.new.in_sam.count
-  end
-
-  def accepted_auctions_count
-    accepted_auctions.count
-  end
-
   def accepted_auction_percent
-    Percent.new(accepted_auctions_count, published_auction_count).to_s
+    Percent.new(accepted_auctions_count, accepted_and_rejected_auction_count).to_s
```
harrisj: Nice.

harrisj:
LGTM. Let's merge it.

PR #825, opened 2016-07-06
Add customer to auctions

### LINKED ISSUE #671: Associate an auction with a customer
As an admin, I want to be able to actually associate an auction with a client via the UI, so that I can see reporting by client

Scenario: I associate a customer with a new auction
Given I am a logged in admin
And I am creating an auction
And customers exist
When I view the form
Then I should see a select dropdown containing the names of all customers in the system
And I should be able to associate a customer with the auction I create

Scenario: Admin sees customer on show page of auction
Given I am a logged in admin
And there is an auction with a customer associated with it
When I view the show page for that auction
Then I should see the name of that customer

Scenario: Admin sees customer on auction preview
Given I am a logged in admin
And there is an auction with a customer associated with it
When I view the auction index page
Then I should see the name of that customer for each auction preview

Scenario: Vendor sees customer on show page of auction
Given I am a logged in vendor
And there is an auction with a customer associated with it
When I view the show page for that auction
Then I should see the name of that customer

Scenario: Vendor sees customer on auction preview
Given I am a logged in vendor
And there is an auction with a customer associated with it
When I view the auction index page
Then I should see the name of that customer for each auction preview

harrisj:
Fixes #671 

This PR adds the following functionality:
- Admin can now select a customer that an auction belongs to
- If a customer is selected, it will be displayed on Auction show pages for admins and vendors
- The customer name will also be returned in the API responses

This does not share the POC or email of the customer, since I assume that is privileged information


app/models/auction.rb
```diff
@@ -32,6 +33,11 @@ def lowest_bids
     bids.select { |b| b.amount == lowest_amount }.sort_by(&:created_at)
   end
 
+  def customer_name
+    return nil if customer.nil?
```
jessieay: thoughts on using null object pattern instead? `NullCustomer`

app/models/auction_parser.rb
```diff
@@ -31,7 +31,8 @@ def auction_params
       :start_price,
       :summary,
       :title,
-      :type
+      :type,
+      :customer_id
```
jessieay: alpha order this?

features/step_definitions/admin_auction_form_steps.rb
```diff
@@ -159,3 +159,24 @@
   expect(DcTimePresenter.convert_and_format(@auction.ended_at)).to eq(DcTimePresenter.convert_and_format(@end_time))
   expect(page).to have_text(DcTimePresenter.convert_and_format(@end_time))
 end
+
+Then(/^I should see a select box with all the customers in the system$/) do
+  #customers = Customer.sorted.all
```
jessieay: remove?

adelevie:
Other than @jessieay's points, LGTM ✅

app/models/auction.rb
harrisj: I guess we should, although the Customer and NullCustomer model will only have one attribute

jessieay: I was thinking something like this:

```
class Auction
  def customer name
    find_customer.name
  end

private

  def find_customer
    customer | NullCustomer.new
  end
end

###

class NullCustomer
  def name
    nil
  end
end
```

features/admin_associates_customer_with_auction.feature
```diff
@@ -0,0 +1,45 @@
...
+    Given I am an administrator
+    And I sign in
+    And there is an auction with an associated customer
+    When I visit the auction page
+    Then I should see the customer name on the page
+
+    When I visit the admin auction page for that auction
+    Then I should see the customer name on the page
+
+  Scenario: Vendor sees the customer
```
jessieay: seems like these vendor-specific tests shouldn't require signing in-- should we remove user creation / sign in steps to prevent confusion about whether only certain types of ppl can see this info?

features/admin_associates_customer_with_auction.feature
```diff
@@ -0,0 +1,45 @@
+Feature: Admin associates customer with auction
+  As an administrator
+  I want to be able to associate auctions with customers
+  So that users and admins can see who is creating certain projects
+
+  Scenario: Associating an auction with a customer
```
jessieay: maybe this should be in a spec related to creating or updated an auction?

features/admin_associates_customer_with_auction.feature
```diff
@@ -0,0 +1,45 @@
+Feature: Admin associates customer with auction
```
jessieay: if below scenario is moved to create / update feature spec file, perhaps the other specs can be moved to files related to seeing auction attributes? I think that is how we are mostly dealing with our other attrs....

jessieay:
@harrisj left a few more comments re: organization / setup in feature spec. LMK what you think! Code looks 💯

features/admin_associates_customer_with_auction.feature
harrisj: Yeah, they work for guests too. I suppose I should move them to `vendor_views_auction` and `guest_views_auction`

features/admin_associates_customer_with_auction.feature
harrisj: You mean a controller test?

jessieay: was thinking more like a feature spec file that is related to more generally "creating auctions" or "updating auctions" -- your call!

harrisj: True I suppose I could move them all there

harrisj:
Rebased and updated featured files based on your comments

jessieay:
shoot! failing now @harrisj

harrisj:
Fixed the failing test now

jessieay:
love it - merging

Issue #776, opened 2016-06-27
New design for user sign-up/registration page


mtorres253:
This story is to implement the redesign of the sign-up/register page. The prototype is below. Behaviors are the same, we're just tweaking the language and re-doing the layout.

https://pages.18f.gov/micropurchase/sign-in/

PR #827, opened 2016-07-06
Improve login page design

jessieay:
* this page may be removed soon, but I already put together new design
 so may as well commit?
* closes https://github.com/18F/micropurchase/issues/776

screen shot 2016-07-06 at 2 21 22 pm

### DIFF
--- a/app/views/components/_guest_nav_drawer.html.erb
+++ b/app/views/components/_guest_nav_drawer.html.erb
@@ -1,2 +1,3 @@
-<!-- <a href="/login" class="button&#45;login usa&#45;button usa&#45;button&#45;outline&#45;inverse">Login</a> -->
-<%= link_to 'Login', login_path, class: 'usa-button usa-button-outline-inverse' %>
+<% if request.path != '/login' %>
+  <%= link_to 'Login', login_path, class: 'usa-button usa-button-outline-inverse' %>
+<% end %>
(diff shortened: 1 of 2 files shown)


app/views/components/_guest_nav_drawer.html.erb
```diff
@@ -1,2 +1,3 @@
-<!-- <a href="/login" class="button&#45;login usa&#45;button usa&#45;button&#45;outline&#45;inverse">Login</a> -->
-<%= link_to 'Login', login_path, class: 'usa-button usa-button-outline-inverse' %>
+<% if request.path != '/login' %>
```
adelevie: very logic-y for a view? I guess if the thing will get tossed out, then no big deal?

adelevie:
one comment, otherwise LGTM

app/views/components/_guest_nav_drawer.html.erb
jessieay: yeah I wasn't sure how else to do it. will probably be thrown out but I am open to ideas on a better way!

adelevie: controller:
```ruby
@request_presenter = RequestPresenter.new(request)
```

view:
```erb
<% if @request_presenter.show_login_button? %>
```

I can't tell if I'm joking or not.

jessieay: haha that's still a conditional in the view!

adelevie:
```erb
<%= render partial: @request_presenter.login_button_partial %>
```

:trollface:

Issue #779, opened 2016-06-27
Customer gets email on auction acceptance


mtorres253:
As an admin user, I want to automatically notify any customer who is using their own p-card when an auction is accepted, so that they can pay the vendor, so the system should do this for me.

Scenario: Customer using their own p-card gets email when auction is accepted
Given I am a logged in admin
And an auction for that customer has been successfully delivered
And the winning vendor has a Credit Card URL
When the auction is accepted
Then the system generates an email
From address: admin@micropurchase.gov
To address: [contact email]
Subject: Your auction has been successfully delivered!
Body: Dear [contact name]
Your auction [auction name] has been successfully delivered! The winning vendor is [vendor name] and the final price of the auction is [xxxx]. 

You may examine the delivered code here: [delivery_url]

Questions? Contact micropurchase@gsa.gov.

Sincerely,

The 18f Micropurchase team

Note: Let's discuss details and break down

jessieay:
@mtorres253 do we own admin@micropurchase.gov ? If not, I say we send it from `micropurchase@gsa.gov` so replies are captured properly...

jessieay:
also, not sure " You will need to pay the vendor using your own P-card." makes sense -- eg: if the client is Tock, they are not going to use their own p-card. Maybe different text for email based on pcard used for auction?

mtorres253:
Yes, cause Tock will be using our P-card

jessieay:
another caveat -- customers won't always have a `name` -- do we want to require the name field or remove the greeting? would be weird if there is no name

jessieay:
PR https://github.com/18F/micropurchase/pull/830

mtorres253:
Taking out the reference to P-card. We can add this in a future story if indicated.

PR #830, opened 2016-07-07
Customer gets email on auction acceptance

jessieay:
* Closes https://github.com/18F/micropurchase/issues/779

* Note: we may add some more content to this later depending on Pcard (eg: non-18F pcard users are asked to pay and given credit card URL).

screen shot 2016-07-07 at 12 22 08 pm

### DIFF
--- a/app/models/customer.rb
+++ b/app/models/customer.rb
@@ -1,7 +1,6 @@
-# Represents a single agency customer
 class Customer < ActiveRecord::Base
   validates :agency_name, presence: true, uniqueness: true
-  validates :email, email: true, presence: false
+  validates :email, email: true, allow_nil: true
 
   has_many :auctions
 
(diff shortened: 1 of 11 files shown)


mtorres253:
@jessie Hi! added the bit about the delivery URL that we discussed

jessieay:
@mtorres253 updated! new email copy:

screen shot 2016-07-07 at 1 24 50 pm

app/models/customer.rb
```diff
@@ -1,7 +1,6 @@
-# Represents a single agency customer
 class Customer < ActiveRecord::Base
   validates :agency_name, presence: true, uniqueness: true
-  validates :email, email: true, presence: false
+  validates :email, email: true, allow_nil: true
```
harrisj: Ah thank you, I knew I was doing that wrong

harrisj:
Looks good to me! Merging

app/models/customer.rb
jessieay: I never would have known until I tried creating a customer without an email!

PR #831, opened 2016-07-07
Add search lookup for "billable to"

jessieay:
* Use selectize http://selectize.github.io/selectize.js/
* Ignore brakeman warning `Render path contains parameter value` for:
```
<%= render partial: auction.delivery_due_partial, locals: { f: f,
auction: auction } %>
```

screen shot 2016-07-07 at 1 04 25 pm


app/assets/stylesheets/application.scss
```diff
@@ -2,6 +2,7 @@
 @import 'uswds_overrides/all';
 @import 'c3.min';
 @import 'source/accessible-mega-menu';
+@import 'selectize.bootstrap3';
```
harrisj: Did this require you to define some bootstrap compatible classes in the stylesheet?

app/assets/stylesheets/core/_elements.scss
```diff
@@ -6,14 +6,3 @@ html {
 ul li:before {
   display: none;
 }
-
```
harrisj: Do we need these base elements to have their sizes defined? Will this affect layouts across the app?

app/controllers/admin/auctions_controller.rb
```diff
@@ -29,8 +29,8 @@ def create
       flash[:success] = I18n.t('controllers.admin.auctions.create.success')
       redirect_to admin_auctions_path
     else
-      flash[:error] = auction.errors.full_messages.to_sentence
-      @view_model = Admin::NewAuctionViewModel.new
+      flash.now[:error] = auction.errors.full_messages.to_sentence
+      @view_model = Admin::NewAuctionViewModel.new(auction)
```
harrisj: Is this so you can carry the billable selection forward in case of an error?

adelevie:
LGTM

app/assets/stylesheets/application.scss
jessieay: nope this works as-is. comes from https://github.com/selectize/selectize.js/tree/master/dist/css

app/controllers/admin/auctions_controller.rb
jessieay: yes! also all other fields. noticed when testing this manually that we were not preserving fields values when there was an error, which is annoying for admins

app/assets/stylesheets/core/_elements.scss
jessieay: this does affect layout across app. but I clicked around a bit and didn't see anything too weird. The design standards should be defining these types of default styles for us.

The reason I removed in this PR is that the selectize lib puts a `span` around highlighted elements while you are searching. Because we have special `span` styles, the highlighted bits looked super weird when we had these styles defined. Example:

Before:

screen shot 2016-07-08 at 9 13 26 am


After:

screen shot 2016-07-08 at 9 13 39 am

jessieay:
@harrisj think I responded to all of your feedback on this - LMK what you think!

PR #833, opened 2016-07-08
Email sent to vendor if credit card URL is not on file when auction i…

jessieay:
…s accepted

* Closes https://github.com/18F/micropurchase/issues/770


config/locales/en.yml
```diff
@@ -75,7 +75,15 @@ en:
         para_3: "Questions? Reply to this email."
         sign_off: "Sincerely,"
         from: "The 18F Micro-purchase team"
-
+      winning_bidder_missing_payment_method:
+        subject: "Please update your profile so Micro-purchase can pay you for your great work!"
+        para_1: "A payment of %{amount} is pending release regarding work you
+          recently completed on \"%{auction_title}\". In order to take receipt of payment, you
```
adelevie: minor thing: `In order to take receipt of payment` -> `In order for us to pay you`

We should probably link to _something_ or add a sentence or two about what this means. Thoughts @mtorres253? Could be done in a subsequent PR, too.

adelevie:
Looks great! Minor email copy stuff, could be improved in a subsequent PR.

adelevie:
I'll merge now, copy can be updated as-needed.

Issue #780, opened 2016-06-27
Payment action is taken when vendor adds credit card URL after auction is accepted


mtorres253:
As an admin, I want the system to initiate payment action for an auction when all information is available, so when the auction is accepted, and the winning vendor has a credit card URL on file, the payment action should be started automatically

Scenario: Vendor adds credit card URL after auction is accepted
Given I am a logged in vendor
And I have successfully delivered on an auction
And I don't have a credit card URL on file
And the auction has been accepted
When I add a credit card URL to my profile
Then the payment action for that auction should automatically occur (update previous CAP request or email to customer agency)

Generic concept of payment method may need to be added? This could result in some refactoring.

jessieay:
added a question on another issue that is more related to this story: https://github.com/18F/micropurchase/issues/770#issuecomment-231205480

PR #843, opened 2016-07-11
Auction automatically accepted when vendor adds payment information

jessieay:
* When an admin "accepts" an auction but the winning vendor does not
  have payment information in the system, the vendor is sent an email
* The auction result is set to "accepted" but the "accepted_at"
  timestamp is not sent
* When the vendor logs in and adds a valid Credit Card Form URL to their
  profile, the auction is automatically accepted
* Result: `accepted_at` timestamp is set and C2 proposal is updated with
  payment information
* Closes https://github.com/18F/micropurchase/issues/780

### DIFF
--- a/app/controllers/users_controller.rb
+++ b/app/controllers/users_controller.rb
@@ -1,14 +1,12 @@
 class UsersController < ApplicationController
-  def edit
-    require_authentication
+  before_action :require_authentication
 
+  def edit
     @user = current_user
     @user.decorate.sam_status_message_for(flash)
   end
 
   def update
-    require_authentication
-
     updater = UpdateUser.new(params, current_user)
     if updater.save
       redirect_to root_path
--- a/app/services/accept_auction.rb
+++ b/app/services/accept_auction.rb
@@ -1,17 +1,20 @@
 class AcceptAuction
-  attr_reader :auction
+  attr_reader :auction, :credit_card_form_url
 
-  def initialize(auction)
+  def initialize(auction:, credit_card_form_url: '')
     @auction = auction
+    @credit_card_form_url = credit_card_form_url
   end
 
   def perform
-    if winning_bidder && winning_bidder.credit_card_form_url.blank?
-      AuctionMailer.winning_bidder_missing_payment_method(auction: auction).deliver_later
+    if credit_card_form_url.blank?
+      AuctionMailer
+        .winning_bidder_missing_payment_method(auction: auction)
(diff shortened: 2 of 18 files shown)


app/controllers/users_controller.rb
```diff
@@ -1,14 +1,12 @@
 class UsersController < ApplicationController
-  def edit
-    require_authentication
+  before_action :require_authentication
```
harrisj: Much nicer

app/services/accept_auction.rb
```diff
@@ -1,17 +1,20 @@
...
+  def initialize(auction:, credit_card_form_url: '')
     @auction = auction
+    @credit_card_form_url = credit_card_form_url
   end
 
   def perform
-    if winning_bidder && winning_bidder.credit_card_form_url.blank?
-      AuctionMailer.winning_bidder_missing_payment_method(auction: auction).deliver_later
+    if credit_card_form_url.blank?
+      AuctionMailer
```
harrisj: I wonder if we should move this to a separate method like `send_customer_email`

app/services/update_user.rb
```diff
@@ -33,6 +28,34 @@ def update_sam
     reckoner.delay.set!
   end
 
+  def update_auctions
+    if added_payment_information?
+      AuctionQuery.new.accepted_with_bid_from_user(user.id).each do |auction|
+        if auction.accepted_at.nil? && winning_bidder_for(auction) == user
```
harrisj: Should we take this condition and make it a method? `user_payment_info_needed_for_auction` or such?

features/step_definitions/login_steps.rb
```diff
@@ -20,3 +20,9 @@
   step "I visit my profile page"
   click_on "Submit"
 end
+
+When(/^I am the winning bidder$/) do
```
harrisj: Nice

app/services/accept_auction.rb
jessieay: I like it!

jessieay: updated at 1c60ac3

app/services/update_user.rb
jessieay: what do you think of this: dc2f9ab

harrisj:
Nice. Just waiting on Travis to finish and will merge

jessieay:
@harrisj 💚

PR #849, opened 2016-07-11
Update ribbon css

jessieay:
* Got messed up after `span` global style was removed in https://github.com/18F/micropurchase/pull/831/files

Before:
screen shot 2016-07-11 at 1 11 56 pm

After:

screen shot 2016-07-11 at 1 11 27 pm


harrisj:
CSS is best

Issue #835, opened 2016-07-10
I can view auction skills in auction preview and show page


mtorres253:
As a visitor to the site, I want to be able to see any skills associated with a given auction, so the system should show me that.

Scenario: Visitor sees auction skills on auction preview
Given I am a visitor to the site
And there are existing auctions
And there are skills associated with those auctions
When I am looking at the auction previews on the homepage (micropurchase.18f.gov)
Then I should see a section labeled "Skills"
And I should see the languages associated with that auction

Scenario: Visitor sees auction skills on auction show page
Given I am a visitor to the site
And there is an existing auction
And a language or set of skills are associated with that auction
When I go to the show page for that auction
Then I should see a section labeled "Skills"
And I should see the skills associated with that auction in that section

![preview](https://cloud.githubusercontent.com/assets/100149/16738991/024ecd88-474d-11e6-824e-7a97e8622280.png)

![screen shot 2016-07-11 at 9 59 10 am](https://cloud.githubusercontent.com/assets/100149/16739285/3495c778-474e-11e6-87a1-95b470d6103c.png)

Note: Add skills as a comma separated list

PR #855, opened 2016-07-12
Show auction skills on index and show pages

jessieay:
* Closes https://github.com/18F/micropurchase/issues/835

### DIFF
--- a/app/models/skill.rb
+++ b/app/models/skill.rb
@@ -1,3 +1,4 @@
 class Skill < ActiveRecord::Base
   has_and_belongs_to_many :auctions
+  validates :name, presence: true, uniqueness: true
 end
--- a/app/view_models/auction_list_item.rb
+++ b/app/view_models/auction_list_item.rb
@@ -38,6 +38,10 @@ class AuctionListItem
     end
   end
 
+  def skills
+    auction.skills.map(&:name).join(', ')
+  end
+
   def github_repo
     auction.github_repo
   end
--- a/app/view_models/auction_list_item.rb
+++ b/app/view_models/auction_list_item.rb
@@ -39,7 +39,7 @@ class AuctionListItem
   end
 
   def skills
-    auction.skills.map(&:name).join(', ')
+    auction.sorted_skill_names.join(', ')
   end
 
   def github_repo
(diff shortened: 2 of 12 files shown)


jessieay:
@mtorres253 do we still want these to be under a "Languages" header per the mocks or should we use the more general term "skills" ?

app/models/skill.rb
```diff
@@ -1,3 +1,4 @@
 class Skill < ActiveRecord::Base
   has_and_belongs_to_many :auctions
+  validates :name, presence: true, uniqueness: true
```
harrisj: Oh yes, that is important

app/view_models/auction_list_item.rb
```diff
@@ -38,6 +38,10 @@ def eligibility_label
     end
   end
 
+  def skills
+    auction.skills.map(&:name).join(', ')
```
harrisj: Should these be sorted (we could put a default order on the association even if you'd prefer)?

app/views/auctions/_auction.html.erb
```diff
@@ -1,6 +1,8 @@
 <div class="js-auction-view">
   <div class="auction-body">
     <div class="auction-description">
+      <h2>Skills</h2>
```
harrisj: Do we want this header to be displayed if we have not assigned skills to an auction?

app/views/auctions/_list_item.html.erb
```diff
@@ -37,6 +37,13 @@
               <%= auction.eligibility_label %>
             </div>
           </li>
+
+          <li class="issue-details-list-item">
```
harrisj: Do we want this to be displayed if we have not assigned skills to an auction?

app/view_models/auction_list_item.rb
jessieay: sure, let's do alpha. wil update

jessieay: added method that sorts skills by name and returns names at 293de95

app/views/auctions/_auction.html.erb
jessieay: I think so...wasn't specified in story, but I imagine we will always have some skills for an auction.

harrisj:
Okay, looks good to me. Merging.

Issue #850, opened 2016-07-11
Skills information comes back via the API


mtorres253:
As a consumer of the API, I want to see the skills for each auction, so the API should return this information

Scenario: I see skills for a given auction in the API response
Given I am an authenticated consumer of the API
When I make a request for auction data
Then the response should contain the skills for that auction

Note: this is just a first step. If we need to do more we can.

PR #858, opened 2016-07-12
Add `skills` to API

jessieay:
* Array of strings for now
* Closes https://github.com/18F/micropurchase/issues/850


dogweather:
Cool. Do you have to edit the swagger config by hand?

harrisj:
Yes, we are not using something like Grape or such that generates the swagger file for us from annotations on the method. This might seem weird, but I personally prefer it since we are using the swagger file to test the API responses so it's checking that our API matches the contract we have described (as opposed to generating a Swagger file to match whatever the API happens to be returning). But that's just me.

harrisj:
Forgot to add LGTM

Issue #836, opened 2016-07-10
Skills chart on the insights page


mtorres253:
As a visitor to the site, I want to know overall which skills are represented in the past auctions, so i should easily be able to see that in a chart 

Scenario: I see a programming langauge chart on the micropurchase insights page
Given I am a visitor to the site
When I am looking at the insights page
Then i should see a chart labeled "Skills"
And in that chart I should see a list of all skills that have been associated with auctions
And for each language I should see the number of auctions
And the list should be in order of most popular skill to least

![screen shot 2016-07-12 at 1 10 56 pm](https://cloud.githubusercontent.com/assets/100149/16781895/268e85b8-4832-11e6-9bc6-18a53e4d4171.png)

jessieay:
@mtorres253 screenshot upload in description of this ticket failed - can you try again?

mtorres253:
@jessieay done

PR #863, opened 2016-07-13
Add skills chart on the insights page

jessieay:
* Closes https://github.com/18F/micropurchase/issues/836
* Replaces donut chart made from static data that we were using before

screen shot 2016-07-12 at 5 09 56 pm

### DIFF
--- a/app/assets/javascripts/insights/charts.js
+++ b/app/assets/javascripts/insights/charts.js
@@ -205,30 +205,11 @@
             pattern: ['#1C304A','#00CFFF','#046B99','#B3EFFF']
           }
         });
-      },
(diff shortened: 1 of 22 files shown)


harrisj:
Looks fine, but CodeClimate is super annoyed. Do you want to look those over and see if any of them make sense to fix?

jessieay:
@harrisj yes will attempt to fix now!

jessieay:
ok @harrisj cleaned it up -- not sure I can eliminate final two CC failures

harrisj:
Eh, we'll fix them all some day. Merging...

PR #866, opened 2016-07-13
Fix login issue for users without name in GitHub

jessieay:
* If a user does not have a name in github, `name` key has a value of
  `null`
* We require `name` at the postgres level, but default to `""`
* Check for `null` in login user so we don't have problems when ppl
  without a name in GH try to log in

### DIFF
--- a/spec/models/login_user_spec.rb
+++ b/spec/models/login_user_spec.rb
@@ -2,6 +2,28 @@ require 'rails_helper'
 
 describe LoginUser do
   describe '#perform' do
+    context 'user does not have name in github' do
+      it 'creates the user without a name' do
+        github_id = '1234'
+        auth_hash = {
+          uid: github_id,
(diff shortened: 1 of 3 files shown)


spec/models/login_user_spec.rb
```diff
@@ -92,22 +114,22 @@
...
-      uid: github_id_from_oauth,
-      info: {
-        nickname: 'github_username',
-        name: 'Kane',
-        email: 'email@example.com',
-        image: 'github-image.png'
-      }
-    )
-  end
+    def auth_hash
```
harrisj: Looks like some weird whitespace got in there?

harrisj: Never mind. Looks fine locally in rubocop

harrisj:
LGTM. Merging.

PR #867, opened 2016-07-13
Add database cleaner configs

jessieay:
* Before, our DB was not being cleared of records before each test run
* Noticed when working on last feature that my test DB was not clearing
  itself of users when I ran a single test over and over


harrisj:
What? I know we had this working in the past. When did this go away? Anyhow, good catch.

jessieay:
@harrisj It was working when you ran the whole suite but not when you ran a single test. So if you are working on a feature and running a single file over and over the test db filled up with records. I noticed this because it was causing a false negative while I was working on something separately. Now the test DB is emptied after ever spec.

Issue #856, opened 2016-07-12
I can create skills to be associated with auctions


mtorres253:
As an admin user, I want to be able to maintain a library of skills that I can use to help me describe auctions better to vendors. So there should be a way for me to do this

Scenario: Admin creates a skill
Given I am an admin user
When I navigate to the admin section
Then I should see a tab labeled "Skills"

When I click on that tab
Then I should go to a page with a header "Skill"
And I should see a list of existing skills
And I should see a button labeled "Add skill"

When I click on the "Add skill" button
Then I should go to a page labeled "Add a new skill"
And I should see an input labeled "Skill name"
And I should see a button labeled "Add new skill"

When I click on "Add new skill"
And there is no skill specified in the "Skill name" field
Then I should see an error "Skill name cannot be blank"
And I should stay on the page

When I click on "Add new skill"
And there are no errors
Then the skill should be saved
And I should be taken back to the skills index page

PR #869, opened 2016-07-13
Admin can add skills

jessieay:
* Closes https://github.com/18F/micropurchase/issues/856
* Also update customer stuff to make the two more consistent


config/routes.rb
```diff
@@ -24,6 +24,7 @@
     resources :action_items, only: [:index]
     resources :drafts, only: [:index]
     resources :proposals, only: [:create]
+    resources :skills, only: [:index, :new, :create, :edit, :update]
```
harrisj: Do we want to allow admins to delete skills? I guess they can edit if there is a typo

harrisj:
LGTM

Issue #837, opened 2016-07-10
Sign-up to the micropurchase platform


mtorres253:
As a vendor who is signing up, I want my sign-up process to be as painless as possible, and I want to know what to expect, as well as what is expected of me, if I do sign up

Scenario: Vendor sees header with sign-up link
Given I am a vendor who is logged out
When I come to the micropurchase platform
Then I should see in the header, a message that I need to use GitHub to login or sign up (see header here https://pages.18f.gov/micropurchase/profile/?)
And I should see links to do both

When I click on the sign-up link
Then I should be taken through the GitHub authentication process (on GitHub)

When I have authenticated with GitHub
Then I should be returned to the micro-purchase site
And I should be taken to a page where I can see information on what to expect, and what is expected 
of me.  (we should change the term from "Terms and Conditions" to "How it works") https://pages.18f.gov/micropurchase/sign-up/#
And I should see a button labeled "Continue"

When I click on that button
Then I should be signed-up and logged in
And I should be taken to my profile page (https://pages.18f.gov/micropurchase/profile/)
And I should see in the header that I am logged in as: [my name]
And I should see a link to log out
And I should see a link (labeled "My Profile") to my profile (it should be inactive, as I am on my profile page)
And I should see a link (labeled "My Bids") to the "My Bids" page (https://micropurchase.18f.gov/bids)

Note: This story does NOT include the layout/style changes in the profile page mock listed above. 

*Use this copy for the page title:*
Hi, GitHub.name, it looks like you’re new around here! We want to make sure you know how to participate and what to expect, so take a look, or continue to the Micro-purchase site.

Instead of "Terms and Conditions" just show the copy. (Don't need to have the scroll box)

Button should say "Continue" not "Register"

PR #874, opened 2016-07-14
Update sign up flow

jessieay:
* New users see special "sign up" page
* Existing users just need to click link in header
* Closes https://github.com/18F/micropurchase/issues/837

screen shot 2016-07-14 at 1 57 20 pm


jessieay:
question for @mtorres253  -- as we learned from one user, not all people have a `name` in GitHub. Which means that some people would get to this point and it would read "Hi,, it looks like you're new around here!" -- thoughts on removing one comma so it would read more nicely? ("Hi, it looks like you're new around here!" or "Hi Jessie Young, it looks like you're new around here!") -- thoughts?

harrisj:
LGTM

PR #876, opened 2016-07-14
Only enqueue emails for non-closed auctions

jessieay:
* Updating closed auctions should not enqueue emails


adelevie:
merging when 🍏

adelevie:
merging before travis is done because travis is being extra slow today and this feature is needed to prevent spamming users.

PR #888, opened 2016-07-18
Rejected auctions with no bids would crash action items

### LINKED ISSUE #865: Bug on action items page
I think the issue is that if a rejected auction does not have any bidders, there is a `undefined method for nil` error. Load https://micropurchase-staging.18f.gov/admin/action_items and tail logs to get more details

harrisj:
Fixes #865 

It would be good to get a more elegant way of checking if there is a bidder before I generate the link, so suggestions welcome.

### DIFF
--- a/app/views/admin/action_items/_rejected_auctions.html.erb
+++ b/app/views/admin/action_items/_rejected_auctions.html.erb
@@ -15,7 +15,7 @@
       <td><%= link_to(auction.title, admin_auction_path(auction.id)) %></td>
       <td><%= auction.delivery_due_at %></td>
       <td><%= auction.delivery_url %></td>
-      <td><%= link_to auction.winning_bidder_name, admin_user_path(auction.winning_bidder_id) %></td>
+      <td><% if auction.winning_bidder? %><%= link_to auction.winning_bidder_name, admin_user_path(auction.winning_bidder_id) %><% end %></td>
       <td><%= auction.rejected_at %></td>
       <td><%= link_to('edit', edit_admin_auction_path(auction.id)) %></td>
     </tr>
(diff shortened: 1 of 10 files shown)


harrisj:
Okay, Travis had an issue, but rerunning it worked

app/views/admin/action_items/_rejected_auctions.html.erb
```diff
@@ -15,7 +15,7 @@
       <td><%= link_to(auction.title, admin_auction_path(auction.id)) %></td>
       <td><%= auction.delivery_due_at %></td>
       <td><%= auction.delivery_url %></td>
-      <td><%= link_to auction.winning_bidder_name, admin_user_path(auction.winning_bidder_id) %></td>
+      <td><% if auction.winning_bidder? %><%= link_to auction.winning_bidder_name, admin_user_path(auction.winning_bidder_id) %><% end %></td>
```
jessieay: rather than checking for `if bidder?` in the view, what about creating a duck-type in `NullBid` that returns something sensible. eg: "No bidder"

jessieay: this may require using a different method name / presenters (eg: you call `winning_bidder.link` and `UserPresenter` returns a link and `GuestPresenter` returns a string)

harrisj: Yeah, I thought about that, but it means basically calling `html_safe` on something from the presenter and putting view model stuff deep in the NullBid object so I didn't think that would be better...

jessieay: why is calling `html_safe` on something from the presenter bad? This situation feels similar to the other situations where we branch depending on `Guest` or `User` so using the same solution sounds alright to me...

jessieay:
This LGTM! Merging

PR #894, opened 2016-07-18
Redirect guests when trying to access `sign_up` page

jessieay:
* This is the page you see when you first sign up
* It requires that the current user be a user, not a guest


harrisj:
LGTM

Issue #864, opened 2016-07-13
Enable micropurchase request


phirefly:
As an admin of micropurchase who is responsible for paying on the 18f P-card, I want to see all micropurchase payment requests when I go to C2 so that I can get all email notifications w/r/t those requests. So, when a request is made from the micro-purchase system to C2, it should be labeled as a micropurchase request.

Scenario: Micropurchase marks C2 request as a micropurchase request
Given that an auction exists
And it uses the default payment method (18F p-card)
When the initial C2 request is made to get approval for that auction
Then the request should be marked as a micropurchase request
And emails for those requests should be sent to all users with the role `gsa18f_micropurchase_purchaser`

Update micropurchase requests to indicate type `Micropurchase` instead of previously used `Software`. C2 now handles the Micropurchase type https://github.com/18F/C2/pull/1395/files, which will route purchaser emails to the user that has the role `gsa18f_micropurchase_purchaser`.

jessieay:
@mtorres253 should we prioritize this? Would be ~ 2 points

jessieay:
PR for this is at https://github.com/18F/micropurchase/pull/897 -- will test on staging once its merged!

PR #897, opened 2016-07-18
Set C2 proposal type to "Micropurchase"

jessieay:
* New request type in C2 that routes to micropurchase purchaser
* Closes https://github.com/18F/micropurchase/issues/864


harrisj:
LGTM!

Issue #860, opened 2016-07-12
Fix logic on action items page


jessieay:
Right now, we have "payment needed" and "payment pending" sections

payment needed:

- delivery due date expired
- delivered
- accepted
- c2 NOT submitted
- not paid

payment pending:

- delivery due date expired
- delivered
- accepted
- c2 submitted
- not paid

The problem with these scopes is that in our new world, c2 proposals are submitted before an auction goes live, so nothing should be "payment needed". This also doesnt take into account auctions that are not for the 18F pcard.....so we need to change the logic

mtorres253:
@jessieay Question: do we even need to refer to C2 at all? Seems like payment pending is that auction has been delivered and accepted, but not paid. Paid is when it's been paid. I believe for auctions using the 18F P-card, we are checking C2 to see if the auction has been paid. For other auctions, we'll need to have someone set a paid status.

jessieay:
@mtorres253 so you want to remove the "payment pending" section?

mtorres253:
@jessieay I think that payment pending is just accepted but not paid yet. So we can/should keep that as admins will need to know that an auction winner needs to be paid.

jessieay:
so then should we remove "payment needed:"  ? As I described in the description of this ticket, the two sections currently contain the same logic. If we want to change one or both, we should! Otherwise, one needs to go.

mtorres253:
@jessieay Actually, let's keep "Payment Needed" for now. Auctions that have been accepted will appear there until they've been paid. We should have a column labeled "C2 URL" and if there is one, we should show it there (so admins can check on the status). If the auction is not using the 18F p-card, then there can be a message N/A in the C2 URL column. I think that would be good for now.

jessieay:
@mtorres253 so you want to keep both the "payment needed" and "payment pending" sections even though they will contain the same auctions?

mtorres253:
@jessieay Nope. Get rid of the payment pending and just peek payment needed

jessieay:
@mtorres253 got it!

PR #899, opened 2016-07-19
Remove "payment pending" section in action items index

jessieay:
* Use to be for auctions with a c2 proposal but without paid_at
* Now all auctions for the default purchase card have a c2 proposal before the auction goes live so everything that has been
  delivered and accepted but is not paid is 'payment needed'
* Closes https://github.com/18F/micropurchase/issues/860#issuecomment-233496240


harrisj:
LGTM. Merging!

mtorres253:
@jessieay also auctions that do are not using default p-card that are accepted appear here as well (with N/A in the Cap URL) field.

jessieay:
@mtorres253 yes that makes sense because they haven't been paid

PR #903, opened 2016-07-19
Unpublished auctions should not be available

### LINKED ISSUE #898: Vendors cannot bid on unpublished auctions
As an admin user, I want to prevent vendors from bidding on an auction that isn't published, so if an auction is in an unpublished state, then the vendor should not be able to bid on it.

Scenario: Vendor tries to bid on an auction that is unpublished
Given an auction exists
And there are vendors
And the auction is unpublished
When a vendor tries to bid on that auction
Then the vendor should see an error "This auction is not currently published"
And the bid should not be accepted

Note: This is to deal with the case where an auction that is live is unpublished (for whatever reason) and someone tries to bid via API

harrisj:
Fixes #898

Rather than make an integration test for a negative feature, I just added an rspec test to make sure the status available returns false if the auction is not published

### DIFF
--- a/spec/models/auction_status_spec.rb
+++ b/spec/models/auction_status_spec.rb
@@ -1,155 +1,115 @@
-require "rails_helper"
+require 'rails_helper'
 
 describe AuctionStatus do
-  describe "#available?" do
-    context "start datetime in past, end datetime in future" do
-      it "is true" do
-        auction = FactoryGirl.build(:auction, started_at: yesterday, ended_at: tomorrow)
-
+  describe '#available?' do
+    context 'start datetime in past, end datetime in future' do
+      it 'is true' do
+        auction = FactoryGirl.build(:auction, :available)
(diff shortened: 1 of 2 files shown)


spec/models/auction_status_spec.rb
```diff
@@ -1,155 +1,115 @@
...
-        it "is true" do
-          auction = FactoryGirl.build(:auction, ended_at: yesterday)
-
-          status = AuctionStatus.new(auction)
-
-          expect(status).to be_over
-        end
+    context 'start datetime in past, end datetime in past' do
+      it 'is false' do
+        auction = FactoryGirl.build(:auction, :closed)
```
jessieay: thoughts on *not* using `FactoryGirl` prepend thingy? Seems like we are doing this inconsistently around our test suite.

jessieay:
great solution! 👍

Issue #902, opened 2016-07-19
Total auctions on insights page is incorrect


mtorres253:
Currently the insights page shows 18 total auctions. This number should represent all auctions that were accepted or rejected. So, the total auctions should = 21. Total successful auctions should be those auctions that have been accepted = 16

16/21 = 76% which is the correct success rate

PR #904, opened 2016-07-19
Update "total" count on insights page

jessieay:
* Should include accepted and rejected count, regardless of published
  status
* Adding this revealed some issues around accepting and rejecting
  auctions without bids
* Closes https://github.com/18F/micropurchase/issues/902

### DIFF
--- a/app/models/auction_query.rb
+++ b/app/models/auction_query.rb
@@ -28,7 +28,7 @@ class AuctionQuery
     relation
       .delivery_due_at_expired
       .delivered
-      .accepted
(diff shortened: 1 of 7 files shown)


app/models/auction_query.rb
```diff
@@ -28,7 +28,7 @@ def complete_and_successful
     relation
       .delivery_due_at_expired
       .delivered
-      .accepted
+      .delivery_accepted
```
harrisj: I like this new name

harrisj:
LGTM!

Issue #859, opened 2016-07-12
Paginate auctions on homepage


jessieay:
As an developer on micropurchase, I want to minimize page load times so that I can more efficiently do my work. So there should be pagination on the auctions index so that I don't have to wait for all auctions to load every time

Scenario: There are more than 20 auctions and I view the micropurchase home page
Given I am a visitor to the micropurchase site
And there are more than 20 auctions
When I go to the homepage and I view all auctions
Then the results should be paginated
And I should see no more than 20 per page

Right now, I have several dozen auctions in my local dev environment. This makes page load time a little slow and also makes it hard to scroll down to the footer. Should we paginate records? Maybe 20 per page?

Note: use the style on this page if possible https://pages.18f.gov/micropurchase/

PR #906, opened 2016-07-19
Add pagination to auctions#index

jessieay:
* Use kaminari
* 10 items per page
* Closes https://github.com/18F/micropurchase/issues/859

screen shot 2016-07-19 at 2 20 12 pm

screen shot 2016-07-19 at 2 19 57 pm


harrisj:
This looks nice! But why are we backpedaling on the AdminConstraint that was added recently? Do we need to rebase?

jessieay:
@harrisj good question! What I found was that when I added pagination it was doing funky stuff because it tried to paginate the "root url" but as an admin my root URL was the action items page. So clicking to go to the next page brought me to the action items page with some weird params for pagination that were ignored.

In addition, I felt it was annoying that visiting "/" as an admin and seeing different content than a regular user was getting annoying. I kept getting into a situation where I wanted to be on the vendor home page but was on the admin action items page instead. Feels more straightforward to have the same URLs show same content for all user types (unless they are not allowed to view the content, of course)

The intent of the "admin logs in and sees action items" story was to have that be the default page after an admin logs in. In the updated code in this PR, we accomplish that be redirecting them after login rather than by redefining the root url. Am I making sense?

harrisj:
Got it. Thank you for the clarification on this. I am good to merge now.

PR #908, opened 2016-07-20
Miles of style

jessieay:
* Ran `bundle exec rubocop` and fixed stuff
* Not 100% green but this gets us closer to being able to run rubocop
  locally and get helpful information rather than waiting for Code
  Climate

### DIFF
--- a/app/serializers/admin/user_serializer.rb
+++ b/app/serializers/admin/user_serializer.rb
@@ -12,10 +12,10 @@ class Admin::UserSerializer < ActiveModel::Serializer
   )
 
   def created_at
-    object.created_at.iso8601 rescue nil
+    object.created_at.iso8601
   end
 
   def updated_at
-    object.updated_at.iso8601 rescue nil
+    object.updated_at.iso8601
   end
 end
--- a/spec/services/create_c2_proposal_spec.rb
+++ b/spec/services/create_c2_proposal_spec.rb
@@ -48,8 +48,9 @@ describe CreateC2Proposal do
 
         expect do
           CreateC2Proposal.new(auction).perform
+          auction.reload
         end
-          .to change { auction.reload.c2_proposal_url }
+          .to change { auction.c2_proposal_url }
           .from('')
           .to("https://c2-dev.18f.gov/proposals/#{fake_c2_proposal_id}")
       end
--- a/spec/services/create_c2_proposal_spec.rb
+++ b/spec/services/create_c2_proposal_spec.rb
@@ -8,7 +8,9 @@ describe CreateC2Proposal do
 
         fake_c2_attributes = { fake_c2: 'fake' }
         attributes_double = double(perform: fake_c2_attributes)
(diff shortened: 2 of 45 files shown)


app/serializers/admin/user_serializer.rb
```diff
@@ -12,10 +12,10 @@ class Admin::UserSerializer < ActiveModel::Serializer
   )
 
   def created_at
-    object.created_at.iso8601 rescue nil
+    object.created_at.iso8601
```
harrisj: I know that rubocop wants us to remove `rescue nil` but in this case I think we need to make sure this doesn't leave us with uncaught exceptions being thrown

jessieay: `created_at` and `updated_at` are never nil

spec/services/create_c2_proposal_spec.rb
```diff
@@ -38,28 +42,31 @@
...
+          .and_return(response_double)
 
         expect do
           CreateC2Proposal.new(auction).perform
-          auction.reload
         end
-          .to change { auction.c2_proposal_url }
-            .from('')
-            .to("https://c2-dev.18f.gov/proposals/#{fake_c2_proposal_id}")
+          .to change { auction.reload.c2_proposal_url }
```
harrisj: Was this test not working before?

spec/support/helpers/factory_girl.rb
```diff
@@ -1,7 +1,7 @@
-module TimeHelpers
-  def self.quartile_minute(time)
+module FactoryGirl::Syntax::Methods
```
harrisj: Oh, much nicer

app/serializers/admin/user_serializer.rb
harrisj: Okay then!

harrisj:
This looks good to me, but I was wondering if we are planning to copy over the Rubocop file from the Ruby style guide discussion? Or are we still waiting for that to be finalized?

jessieay:
@harrisj hmmmm I am not sure! we could try? I have my preferred styles (many of which we do not have currently) but am happy to use any style guide as long as we use one!

spec/support/helpers/factory_girl.rb
jessieay: yes! found this method via https://github.com/thoughtbot/factory_girl/issues/564#issuecomment-157669866

spec/services/create_c2_proposal_spec.rb
jessieay: ok will move back

jessieay: moved at 2caac27

harrisj:
Okay, let's merge this, but it's possible that if we adopt the central style guide, some of this work will be undone :(

jessieay:
@harrisj if that happens, then I will be on my way to code nirvana aka all my code being deleted YES

harrisj:
You are awaited in Code Valhalla

PR #912, opened 2016-07-20
Add security false positives to `config/brakeman.ignore`

jessieay:
* This will ignore them for both code climate and hakiri

### DIFF
--- a/.codeclimate.yml
+++ b/.codeclimate.yml
@@ -3,9 +3,6 @@ engines:
     enabled: true
   brakeman:
     enabled: true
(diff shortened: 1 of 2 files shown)


harrisj:
Nice, but what was generating that detailed information about where the false positive occurred in the old file?

jessieay:
@harrisj brakeman can do that in an automated way. I myself don't find it super helpful to have all of the extra info because if you manually run brakeman you can see all of that info on the command line. Can bring it back, though!

harrisj:
Will leave for now

Issue #886, opened 2016-07-16
Auctions not using 18F default payment method can be marked as paid


mtorres253:
As an admin user, I want to to know when any auction has been paid, so if there are auctions not using the default payment method (18F p-card) there should be a way for me to manually mark the auction as paid after it has been accepted

Scenario: Accepted auction with payment method != default is marked as paid
Given I am an authenticated admin
And there is an auction
And that auction is not using the default payment method
And the auction is accepted
Then I should see a checkbox on the edit form for that auction (https://micropurchase.18f.gov/admin/auctions/[id]/edit) 
And the checkbox should be labeled "Vendor has been paid"

When I check that checkbox
And I save my changes
Then the auction should be paid
And the paid at timestamp should be the timestamp I made my changes

When I return to the edit form for that auction
Then I should see the "Vendor has been paid" checkbox as checked and disabled

When I look at the auction show page for that auction (https://micropurchase.18f.gov/admin/auctions/[id]/edit) 
Then I should see that timestamp in the "Paid at" field

When I go to the action items page (https://micropurchase.18f.gov/admin/action_items)
Then I should see that auction in the "Completed and Successful" section
And I should see true in the "Paid?" column

jessieay:
@mtorres253 under what conditions should the auction be able to be marked as paid? should it already be accepted? just closed?

mtorres253:
@jessieay the auction needs to be accepted in order for it to be paid. I am fine with showing a checkbox in that case

jessieay:
> And I should see true in the "Paid?" column

We show the timestamp currently, is it OK to stick with that?

mtorres253:
I'm seeing this
<img width="1185" alt="screen shot 2016-07-20 at 3 11 06 pm" src="https://cloud.githubusercontent.com/assets/100149/17004935/4fd77d2a-4e8c-11e6-9fdd-e9f24099f69b.png">

jessieay:
oh ok I was looking at the admin/auctions#show page!

PR #916, opened 2016-07-20
Auctions not using 18F default payment method can be marked as paid

jessieay:
For an auction that uses "other" purchase card, on edit:
  * If the auction is not accepted, the checkbox does not show
  * If the auction is accepted and not marked as paid, checkbox shows
  * If auction is paid, checkbox shows but is disabled

* Closes https://github.com/18F/micropurchase/issues/886

Checked:

screen shot 2016-07-20 at 3 38 09 pm

Disabled:

screen shot 2016-07-20 at 3 26 36 pm


harrisj:
Looks good!

Issue #915, opened 2016-07-20
No CAP URL input for auctions that do not use the 18F p-card


mtorres253:
As an admin user, I don't want to see the CAP URL input when I don't need a CAP URL for a particular auction, so that input should not appear when the auction is not using an 18F p-card

Scenario: No C2 inputs on edit form for auction using other payment method for auctions NOT using default payment method
Given I am an authenticated admin
And there is an auction that does NOT use default payment method
When I go to the edit page for that auction (https://micropurchase.18f.gov/admin/auctions/[id]/edit)
Then I do not see the C2 URL input

Scenario: No C2 URL and C2 Approved at fields on auction show page for auctions NOT using default payment method
Given I am an authenticated admin
And there is an auction that does NOT use default payment method
When I go to the show page for that auction (https://micropurchase.18f.gov/admin/auctions/[id])
Then I should not see any reference to "C2 URL" or "C2 Approved At" in the right rail

Scenario: C2 URL column shows N/A for auctions NOT using default payment method
Given I am an authenticated admin
And there are auctions that do NOT use default payment method
When I go to the action items page (https://micropurchase.18f.gov/admin/
Then I should see N/A in the "C2 URL" column for those auctions

PR #917, opened 2016-07-20
Do not show C2 fields on admin/auctions#show for auctions for other p…

jessieay:
…urchase cards

* Also show 'N/A' on action items page instead of nothing
* Closes https://github.com/18F/micropurchase/issues/915


harrisj:
LGTM. Merging

PR #920, opened 2016-07-20
Change Ruby buildpack

jessieay:
* Former buildpack is deprecated
* See https://github.com/ddollar/heroku-buildpack-multi
* This is what C2 uses


harrisj:
Yay, a new ~~jet~~ buildpack!

Issue #913, opened 2016-07-20
New layout for create/edit page


mtorres253:
This story is to just implement the style and layout of the new create/edit page.

Use this prototype as a guide:
https://pages.18f.gov/micropurchase/auctions/edit/

Notes:
- For now leave the auction type dropdown (I'll write a separate story to change that behavior)
- Don't include the unpublish checkbox for the time being
- Create the GitHub section with the current inputs
- Create the payment section with the current input for payment method and customer

This story is just to create the groupings and update the layout.

PR #934, opened 2016-07-22
Style adjustments to Auction form

jessieay:
* Move around fields
* Use selectize multi-select for skill field

* Part of (but doesn't close) https://github.com/18F/micropurchase/issues/913

jul-22-2016 10-13-50


vendor/assets/stylesheets/uswds_overrides/core/_fonts.scss
```diff
@@ -46,50 +46,50 @@
...
-  font-style: normal;
-  font-weight: 700;
-  src: font-url('merriweather-bold-webfont.eot');
-  src: font-url('merriweather-bold-webfont.eot?#iefix') format('embedded-opentype'),
-    font-url('merriweather-bold-webfont.woff2') format('woff2'),
-    font-url('merriweather-bold-webfont.woff') format('woff'),
-    font-url('merriweather-bold-webfont.ttf') format('truetype'),
-    font-url('merriweather-bold-webfont.svg#source_sans_probold') format('svg');
-}
+// @font-face {
```
harrisj: Should this be removed entirely?

jessieay: not sure -- the issue with keeping it is that we don't actually have these fonts in this project so the `font_url` rules are referencing files that do not exist.

The issue with *not* keeping it is that the `uswds` file that this overrides references these fonts at a different location where they do not exist.

Maybe best to just remove entirely since we are not using these font rules. Will do so now!

jessieay: Removed with a hopefully helpful comment at 8e64a3e

jessieay:
@harrisj 💚

harrisj:
LGTM!

Issue #942, opened 2016-07-22
Update deploy script


jessieay:
Right now our deploy depends on an outdated buildpack.

Instead, we can use a similar setup to https://github.com/18F/18f-cli/issues/20

This may require us to define *all* env vars via user-provided services (not sure if we can use manifest files as well) asked about this at https://github.com/18F/18f-cli/issues/20

more comments on this here https://github.com/18F/cg-docs/issues/328#issuecomment-234665011

PR #943, opened 2016-07-22
Revert buildpack changes

jessieay:
* Breaks our env var setup
* We should still update our deploy script
* Created an issue for this at https://github.com/18F/micropurchase/issues/942
* Closes https://github.com/18F/micropurchase/pull/943


harrisj:
Don't cry for me, Heroku!
The truth is I never left you...

merging

Issue #956, opened 2016-07-26
setting purchase card to "other" does not save properly


jessieay:
Just tried creating a new auction for an "other" purchase card and it still showed up as default

PR #961, opened 2016-07-26
Save purchase card updates properly

jessieay:
* was not being updated!
* closes https://github.com/18F/micropurchase/issues/956


harrisj:
Oh no! Good catch!

PR #968, opened 2016-07-28
Change cucumber output format

jessieay:
* This outputs dots instead of full scenarios
* Still get verbose error messages, but less scroll back to see them!

screen shot 2016-07-28 at 10 05 10 am


harrisj:
Big 👍 from me! LGTM

PR #973, opened 2016-07-29
Calculate delivery date async

### LINKED ISSUE #958: Get the business day calculation async
Right now we get the delivery date in business days only when the page is drawn. We want to do this asynch so that the admin can see updates in real time.

This story is  to implement an endpoint that can be called asynch when someone changes any of the default dates of the auction. That would be a better experience for users (rather than having to refresh the page to get the new delivery date)

jessieay:
* Closes #958

async


app/assets/javascripts/deliveryDueAt.js
```diff
@@ -1,13 +1,23 @@
 $(document).ready(function() {
   calculateNewDate = function() {
-    $('.estimated-delivery-date').text("Save your changes to see updated delivery date and time");
+    auctionEndedAt = $("#auction_ended_at").val();
```
adelevie: why don't we use `var` here? (maybe noobish js question)

jessieay: we should! will update

app/assets/javascripts/deliveryDueAt.js
```diff
@@ -1,13 +1,23 @@
...
+    $.getJSON(
+      '/api/v0/business_day',
+      { day_count: dueInDays, date: auctionEndedAt, time: (hour + ":" + minute + " " + amPm)  },
+      function(data) {
+        $('.delivery-date').text(data.date);
+    });
   };
 
-  $( ".auction_due_in_days").change(function() {
+  $(".auction_due_in_days").change(function() {
```
adelevie: very cool!

app/controllers/api/v0/business_days_controller.rb
```diff
@@ -0,0 +1,8 @@
+class Api::V0::BusinessDaysController < ApiController
+  def show
```
adelevie: so this endpoint is totally public, right?

adelevie:
Looks great! Just add the `var`s and I'll merge.

app/controllers/api/v0/business_days_controller.rb
jessieay: yes public! is that problematic? I also didn't add to swagger docs because I don't imagine external parties will care about it much

jessieay:
@adelevie updated at 71dabd4 !

app/controllers/api/v0/business_days_controller.rb
adelevie: no not at all ... I was just trying to understand what's going on. 👍

adelevie:
Travis :'(

jessieay:
@adelevie timezomes :( :( :(

jessieay:
ok @adelevie should be fixed via https://github.com/18F/micropurchase/pull/973/commits/0d6150c6ba2cefa69cebc01c475ff2124e2ce8db -- failure was related to not using DcTime everywhere

PR #979, opened 2016-07-29
Ensure sealed_bid is default in create auction form

adelevie:
fixes https://github.com/18F/micropurchase/issues/932


jessieay:
good one!

PR #987, opened 2016-08-02
Action items -> Needs attention auctions

adelevie:
Lays the groundwork for https://github.com/18F/micropurchase/issues/939 and the other "Needs attention" stories.


app/controllers/admin/needs_attention_auctions_controller.rb
```diff
@@ -0,0 +1,5 @@
+class Admin::NeedsAttentionAuctionsController < Admin::BaseController
```
jessieay: thoughts on `Admin::Auctions::NeedsAttentionController`to be more consistent with https://github.com/18F/micropurchase/blob/develop/app/controllers/admin/auctions/closed_controller.rb ?

jessieay:
One naming question otherwise LGTM! Nice thinking getting this all set up before moving into new functionality

app/controllers/admin/needs_attention_auctions_controller.rb
adelevie: I guess so! good call!

app/controllers/admin/auctions/needs_attention_controller.rb
```diff
@@ -0,0 +1,5 @@
+class Admin::Auctions::NeedsAttentionController < Admin::BaseController
+  def index
+    @view_model = Admin::NeedsAttentionAuctionsViewModel.new
```
adelevie: @jessieay should these be renamed, too? to something like `Admin::Auctions::NeedsAttentionViewModel`?

jessieay: yeah I did not do that for the other controller but probably makes sense. If you do it, mind also renaming https://github.com/18F/micropurchase/blob/develop/app/view_models/admin/closed_auctions_list_item_view_model.rb ?

adelevie: No problemo

jessieay:
Going to merge as-is -- we can rename those view models separately!

adelevie:
Thanks!

PR #989, opened 2016-08-02
Open auction messaging - vendor/reverse auction

jessieay:
* Closes #955 

Bid placed:

screen shot 2016-08-02 at 10 35 25 am

Vendor has not bid yet (no bids or bids present...mac bid):

screen shot 2016-08-02 at 10 35 06 am

Vendor is outbid:

screen shot 2016-08-02 at 10 34 44 am

PR #990, opened 2016-08-02
Open auction messaging - vendor/sealed-bid

### LINKED ISSUE #954: Open auction messaging - vendor/sealed-bid
As a vendor, I want to quickly get high-level information about the status of any open auction so that I don't have to derive it from looking at the auction details, and so that I can take action.

Prototype (see notes): https://pages.18f.gov/micropurchase/auctions/show/
(Select Auction Status: Published (Open), User: Vendor)

**Context**
Given I am a registered vendor
And there is an auction that is sealed-bid
And the auction is open (start date has passed, end date is in the future)

- [x] **Scenario: Vendor who has bid on a sealed-bid auction sees status message in the bid box**
- And I have bid on the auction
- When I go to the auction show page
- Then I should see this message in the bid box

> You bid $[vendor's bid amount] on [mm/dd/yyyy, hh:mm ET].

- [x] **Scenario: Vendor who has NOT bid on a sealed-bid auction sees status message in the bid box, and bid form**
- And I have NOT bid on the auction
- When I go to the auction show page
- Then I should see this message in the bid box

> This auction is accepting bids until [auction.endDate]. The maximum bid for this auction is [starting price - bid increment]

- and I should see a text input
- and I should see a button labeled "Place Bid" next to it

https://pages.18f.gov/micropurchase/auctions/show/?auctionWorkflowState=published-open-bid-rejected-vendors

# Note:
- Use the copy in the requirements. Consult @andrewmaier with any layout questions.

jessieay:
* Closes #954
* Merge AFTER #989

### DIFF
--- a/app/presenters/bid_status_flash_factory.rb
+++ b/app/presenters/bid_status_flash_factory.rb
@@ -7,8 +7,10 @@ class BidStatusFlashFactory
   end
 
   def create
-    if available?
+    if available? && auction.type == 'reverse'
       AvailableUserIsWinningBidder.new(bid_amount: lowest_user_bid.try(:amount))
+    elsif available? && auction.type == 'sealed_bid'
+      AvailableSealedUserIsBidder.new(bid: lowest_user_bid)
     else
       over_message
     end
@@ -110,5 +112,21 @@ class AvailableUserIsWinningBidder
 
   def body
     "You are currently the low bidder, with a bid of #{Currency.new(bid_amount)}"
+  end
+end
+
+class AvailableSealedUserIsBidder
+  attr_reader :bid
+
+  def initialize(bid:)
+    @bid = bid
+  end
+
+  def header
+    ''
+  end
+
+  def body
+    "You bid #{Currency.new(bid.amount)} on #{DcTimePresenter.convert_and_format(bid.created_at)}."
   end
 end
--- a/app/view_models/auction_show_view_model.rb
+++ b/app/view_models/auction_show_view_model.rb
@@ -73,7 +73,9 @@ class AuctionShowViewModel
   end
 
   def bid_status_partial
-    if auction.type == 'reverse' && (over? || available_and_user_is_winning_bidder?)
+    if reverse_auction_over? ||
+       reverse_auction_available_user_is_winner? ||
+       sealed_bid_auction_user_is_bidder?
       'auctions/bid_status'
     else
       'components/null'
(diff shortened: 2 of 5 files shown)


jessieay:
@adelevie this is ready for review

app/presenters/bid_status_flash_factory.rb
```diff
@@ -7,8 +7,10 @@ def initialize(auction:, user:)
   end
 
   def create
-    if available?
+    if available? && auction.type == 'reverse'
```
harrisj: I will admit I don't like it when Presenters are querying the auction.type like this. It feels a bit wrong. Is this somthing we could encapsulate as a method within the respective Rules objects?

app/view_models/auction_show_view_model.rb
```diff
@@ -73,7 +73,9 @@ def bid_status_class
   end
 
   def bid_status_partial
-    if auction.type == 'reverse' && (over? || available_and_user_is_winning_bidder?)
+    if reverse_auction_over? ||
```
harrisj: Similarly, I wonder if we should be moved to the Rules objects or if the Rules objects can instantiate their own corresponding presenters/view models?

config/admins.yml
```diff
@@ -13,4 +13,4 @@ github_ids:
   - 5840989  # Laura Gerhardt
   - 7645362  # Patrick Bateman
   - 4803473  # Brian Hedberg
-  - 601515   # Jessie Young
+  # - 601515   # Jessie Young
```
harrisj: Wait, no!!!

app/presenters/bid_status_flash_factory.rb
jessieay: In this case, it is the `Factory` that is querying the type and figuring out which presenter to return. We could make the factory a model and have it return the presenters -- does that feel more intuitive to you?

app/view_models/auction_show_view_model.rb
jessieay: this is related to the bid status for a user rather than the rules of the auction -- but do you mean we should create rules-type objects?

config/admins.yml
jessieay: oops hah was testing bidder behavior

undid at fd0a7f2

app/view_models/auction_show_view_model.rb
harrisj: I don't know for sure. When I changed the names of the auction types, it was annoying because these factories started failing silently around the place and it bother. That said, it was a rare thing to do, but it felt kinda smelly

app/presenters/bid_status_flash_factory.rb
harrisj: I guess we can leave it for now and revisit if it feels smelly in the future

harrisj:
Decided it's cool to merge this.

app/view_models/auction_show_view_model.rb
jessieay: interesting! if they are failing silently I wonder if there are tests or changes we can make that will make these more obvious...

One certain thing we could do would be to keep those types on a constant so that they are more greppable and raise when they are changed or do not exist.

PR #992, opened 2016-08-02
Update deploy setup

### LINKED ISSUE #942: Update deploy script

jessieay:
* Use 18F-cli instead of outdated buildpack
* We must now define all env vars via user-provided services
* See https://github.com/18F/18f-cli/issues/20
* Closes #942


harrisj:
LGTM. Merging!

PR #993, opened 2016-08-02
Coming soon auction messaging and unpublish button

### LINKED ISSUE #922: Coming soon auction messaging and unpublish button
As a visitor to the site, I want to quickly get high-level information about auctions that are coming soon so that I don't have to derive it from looking at the auction details.

Prototype: https://pages.18f.gov/micropurchase/auctions/show/
(Select Auction Status: Published (Closed, unpublishable), User: Admin

- [x] **Scenario: Unauthenticated visitor sees a status of coming soon auction on show page**
- Given I am a visitor to the site (un-authenticated)
- And there is an auction that has been published
- And that auction has not yet started
- When I go to the show page for that auction
- Then I should see a status box
- And the status box should say

> This auction starts on [auction.startDate]. Sign-in or Sign-up with GitHub to bid.

- And "Sign-in" should link to the current sign-in flow
- And "Sign-up" should link to the current sign-up flow

- [x] **Scenario: Authenticated vendor sees status of coming soon auction on show page**
- And there is an auction that has been published
- And that auction has not yet started
- When I go to the show page for that auction
- Then I should see a status box
- And the status box should say

> This auction starts on [auction.startDate]. 

- [x] **Scenario: Authenticated admin sees status of coming soon auction on show page**
- And there is an auction that has been published
- And that auction has not yet started
- When I go to the show page for that auction
- Then I should see a status box
- And the status box should say

> This auction is visible to the public but is not currently accepting bids. It will open on [auction.startDate]. If you need to take it down for whatever reason, press the unpublish button, below.

- And I should see an "Unpublish" button

- [x] **Scenario: Authenticated un-publishes auction that is coming soon from the show page**
- And there is an auction that has been published
- And that auction has not yet started
- And I am on the auction show page
- When I click on the "Unpublish" button
- And there are no errors
- Then the auction should be unpublished
- And I should be taken to the "Needs Attention" section of the "Auctions" page (#939)
- And I should see the following message

> Auction: [Auction title] has been successfully unpublished

- And I should see that auction in the "Drafts" section

jessieay:
* Closes #922

Authenticated vendor:
screen shot 2016-08-02 at 3 51 38 pm

Guest:
screen shot 2016-08-02 at 3 43 52 pm

Admin:
screen shot 2016-08-02 at 4 06 24 pm

Admin after clicking "un-publish" button:
screen shot 2016-08-02 at 4 25 55 pm

### DIFF
--- a/app/presenters/auction_status_presenter_factory.rb
+++ b/app/presenters/auction_status_presenter_factory.rb
@@ -7,11 +7,31 @@ class AuctionStatusPresenterFactory
   end
 
   def create
-    if current_user.decorate.admin?
+    if available? && current_user.decorate.admin?
       AvailableUserIsAdmin.new(auction: auction)
-    else
+    elsif available?
       AvailableUserIsGuest.new(auction: auction)
+    elsif current_user.is_a?(Guest)
+      FutureUserIsGuest.new(auction: auction)
+    elsif admin?
+      FutureUserIsAdmin.new(auction: auction)
+    else
+      FutureUserIsVendor.new(auction: auction)
     end
+  end
+
+  private
+
+  def available?
+    auction_status.available?
+  end
+
+  def admin?
+    current_user.decorate.admin?
+  end
+
+  def auction_status
+    AuctionStatus.new(auction)
   end
 end
 
@@ -22,7 +42,7 @@ class AvailableUserIsAdmin
     @auction = auction
   end
 
-  def body
+  def to_s
     "This auction is accepting bids until #{end_date}."
   end
 
@@ -34,16 +54,13 @@ class AvailableUserIsAdmin
 end
 
 class AvailableUserIsGuest
-  include Rails.application.routes.url_helpers
-  include ActionView::Helpers::UrlHelper
-
   attr_reader :auction
 
   def initialize(auction:)
     @auction = auction
   end
 
-  def body
+  def to_s
     "This auction is accepting bids until #{end_date}. #{sign_in_link} or
     #{sign_up_link} with GitHub to bid."
   end
@@ -55,10 +72,86 @@ class AvailableUserIsGuest
   end
 
   def sign_in_link
-    link_to 'Sign in', sign_in_path
+    Url.new(link_text: 'Sign in', path_name: 'sign_in')
   end
 
   def sign_up_link
-    link_to 'sign up', sign_up_path
+    Url.new(link_text: 'sign up', path_name: 'sign_up')
+  end
+end
+
+class FutureUserIsGuest
+  attr_reader :auction
+
+  def initialize(auction:)
+    @auction = auction
+  end
+
+  def to_s
+    "This auction starts on #{start_date}. #{sign_in_link} or
+    #{sign_up_link} with GitHub to bid."
+  end
+
+  private
+
+  def start_date
+    DcTimePresenter.convert_and_format(auction.started_at)
+  end
+
+  def sign_in_link
+    Url.new(link_text: 'Sign in', path_name: 'sign_in')
+  end
+
+  def sign_up_link
+    Url.new(link_text: 'sign up', path_name: 'sign_up')
+  end
+end
+
+class FutureUserIsVendor
+  attr_reader :auction
+
+  def initialize(auction:)
+    @auction = auction
+  end
+
+  def to_s
+    "This auction starts on #{start_date}."
+  end
+
+  private
+
+  def start_date
+    DcTimePresenter.convert_and_format(auction.started_at)
+  end
+end
+
+class FutureUserIsAdmin
+  include Rails.application.routes.url_helpers
+  include ActionView::Helpers::UrlHelper
+  attr_reader :auction
+
+  def initialize(auction:)
+    @auction = auction
+  end
+
+  def to_s
+    "This auction is visible to the public but is not currently accepting bids.
+    It will open on #{start_date}. If you need to take it down for whatever
+    reason, press the unpublish button below. #{link}"
+  end
+
+  private
+
+  def link
+    link_to(
+      "Un-publish",
+      admin_auction_published_path(auction),
+      method: :patch,
+      class: 'usa-button usa-button-outline auction-button'
+    )
+  end
+
+  def start_date
+    DcTimePresenter.convert_and_format(auction.started_at)
   end
 end
--- a/app/presenters/auction_status_presenter_factory.rb
+++ b/app/presenters/auction_status_presenter_factory.rb
@@ -7,7 +7,7 @@ class AuctionStatusPresenterFactory
   end
 
   def create
-    if available? && current_user.decorate.admin?
+    if available? && admin?
       AvailableUserIsAdmin.new(auction: auction)
     elsif available?
       AvailableUserIsGuest.new(auction: auction)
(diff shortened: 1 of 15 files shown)


app/presenters/auction_status_presenter_factory.rb
```diff
@@ -7,12 +7,32 @@ def initialize(auction:, current_user:)
   end
 
   def create
-    if current_user.decorate.admin?
+    if available? && current_user.decorate.admin?
```
harrisj: Looks like you could just use the `admin?` method below?

app/presenters/auction_status_presenter_factory.rb
```diff
@@ -22,7 +42,7 @@ def initialize(auction:)
     @auction = auction
   end
 
-  def body
+  def to_s
```
harrisj: Oooh Ruby idioms...

app/presenters/auction_status_presenter_factory.rb
```diff
@@ -55,10 +72,86 @@ def end_date
...
+end
+
+class FutureUserIsGuest
+  attr_reader :auction
+
+  def initialize(auction:)
+    @auction = auction
+  end
+
+  def to_s
```
harrisj: Should these strings be loaded from en.yml? (could be a future PR)

features/guest_views_open_auction.feature
```diff
@@ -22,7 +22,7 @@ Feature: Guest views open auction
     And I should see a current bid amount
     And there should be meta tags for the open auction
     And I should not see the bid form
-    And I should the open auction message for guests
+    And I should see the open auction message for guests
```
harrisj: Nice catch!

features/step_definitions/auction_status_steps.rb
```diff
@@ -13,3 +33,7 @@
 def end_date
   DcTimePresenter.convert_and_format(@auction.ended_at)
 end
+
+def start_date
+  DcTimePresenter.convert_and_format(@auction.started_at)
```
harrisj: Nice!

app/presenters/auction_status_presenter_factory.rb
dogweather: I'm hesitant to include i18n in Rails projects that will only be English by design because IMO it makes the code harder to maintain.

dogweather:
A design comment: I don't know where these decisions are made in the micropurchase project, but when creating a mixed prose+data UI, it improves the UX to distinctively style the dynamic data. E.g.:

screen shot 2016-08-02 at 8 59 12 pm

http://repocheck.com/#https%3A%2F%2Fgithub.com%2F18F%2Fmicropurchase

app/presenters/auction_status_presenter_factory.rb
```diff
@@ -55,10 +72,86 @@ def end_date
...
+  def sign_in_link
+    Url.new(link_text: 'Sign in', path_name: 'sign_in')
+  end
+
+  def sign_up_link
+    Url.new(link_text: 'sign up', path_name: 'sign_up')
+  end
+end
+
+class FutureUserIsVendor
```
dogweather: Personally, I'm not so hot for implicit duck typing like this because I feel it makes it harder for a new project member to get up to speed. So e.g. here, I'd mark these classes in some way to show the author's intent that they're Presenters. I bet this makes it harder for tools (Atom,Rubymine) to help as well.

Maybe I'd add the word Presenter to the class name of each. Or, I'd create a Presenter class which they'd subclass.

app/view_models/auction_show_view_model.rb
```diff
@@ -81,7 +81,7 @@ def bid_status_partial
   end
 
   def auction_status_partial
-    if (current_user.is_a?(Guest) || current_user.decorate.admin?) && available?
+    if (available? && (current_user.is_a?(Guest) || current_user.decorate.admin?)) || future?
       'auctions/status'
     else
       'components/null'
```
dogweather: I like the consistent use of the null object pattern.

app/presenters/auction_status_presenter_factory.rb
jessieay: you are right!

app/presenters/auction_status_presenter_factory.rb
jessieay: one thing I do like about I18n is that it makes changing copy easier in the land of testing. Rather than changing a word and then breaking a bunch of high-level tests (which has happened to me this week), you change the translation and the key stays the same so no test changes are needed.

jessieay: so yes, @harrisj I think it would be cool to i18nify these at some point

app/presenters/auction_status_presenter_factory.rb
jessieay: We only ever instantiate these classes from within this factory (`AuctionStatusPresenterFactory`) but I can see what you mean about wanting clearer names. I can work on that. The alternative to the factory is to have a massive switch statement -- is that what you would prefer?

app/view_models/auction_show_view_model.rb
jessieay: :)

app/presenters/auction_status_presenter_factory.rb
jessieay: updated at 8e1ed6e

app/presenters/auction_status_presenter_factory.rb
harrisj: Okay, we can make that a future task if you want

jessieay: added an issue https://github.com/18F/micropurchase/issues/996

app/presenters/auction_status_presenter_factory.rb
dogweather: I wasn't thinking of a change to the factory logic, only the actual implementation classes. And then, only if you're down with that kind of coding style in the whole project.

So e.g., the cost here is that i had to scan up and down through the file a few times, reading the code to determine that these simple classes really are the presenters; that there's nothing else included, and that these classes must be implementing some kind of duck type which is implicit.

If they were all just grouped in a module (using my third option above), that would have made it self-documenting.

jessieay: I like the idea of using an abstract base class

we are doing that here: https://github.com/18F/micropurchase/tree/develop/app/presenters/sam_status_presenter

to your point about confusion, we are using this pattern in many places in the app, but slightly different in each case. we should make an effort to make them more consistent:

https://github.com/18F/micropurchase/blob/develop/app/models/sam_status_presenter_factory.rb
https://github.com/18F/micropurchase/blob/develop/app/models/status_presenter_factory.rb
https://github.com/18F/micropurchase/blob/develop/app/presenters/bid_status_flash_factory.rb

jessieay: (I would opt to do that update separately since in this PR we are just extending an existing pattern)

jessieay:
@dogweather thanks for the design comment! cc @andrewmaier  on that, he's our designer!

harrisj:
LGTM but we will probably change a few things in other PRs as noted. Merging.

harrisj:
Oops! Wrong button

PR #999, opened 2016-08-03
Update deploy method II

jessieay:
* Include code to parse user-provided services and turn them into ENV
  vars
* Kind of reverts https://github.com/18F/micropurchase/pull/458
* Also kind of reverts https://github.com/18F/micropurchase/pull/992

### DIFF
--- a/app/credentials/smtp_credentials.rb
+++ b/app/credentials/smtp_credentials.rb
@@ -23,7 +23,7 @@ class SMTPCredentials
     if use_env_var?
       ENV['MICROPURCHASE_SMTP_DEFAULT_URL_HOST']
     else
-      credentials('micropurchase-smtp')['default_urL_host']
+      credentials('micropurchase-smtp')['default_url_host']
     end
   end
 
--- a/app/credentials/smtp_credentials.rb
+++ b/app/credentials/smtp_credentials.rb
@@ -1,17 +1,37 @@
+require 'concerns/user_provided_service'
+
 class SMTPCredentials
+  extend UserProvidedService
+
   def self.smtp_password
-    ENV["MICROPURCHASE_SMTP_SMTP_PASSWORD"]
+    if use_env_var?
+      ENV['MICROPURCHASE_SMTP_SMTP_PASSWORD']
+    else
+      credentials('micropurchase-smtp')['smtp_password']
+    end
   end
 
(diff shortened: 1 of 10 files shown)


jessieay:
cc @stvnrlly

app/credentials/smtp_credentials.rb
```diff
@@ -1,17 +1,37 @@
...
+      credentials('micropurchase-smtp')['smtp_username']
+    end
   end
 
   def self.default_url_host
-    ENV["MICROPURCHASE_SMTP_DEFAULT_URL_HOST"]
+    if use_env_var?
+      ENV['MICROPURCHASE_SMTP_DEFAULT_URL_HOST']
+    else
+      credentials('micropurchase-smtp')['default_urL_host']
```
stvnrlly: Should this be all lowercase?

jessieay: that's how we have our user-provided services now -- all lowercase. service names are dasherized, keys are underscore

manifest-staging.yml
```diff
@@ -0,0 +1,21 @@
+---
+applications:
+- name: micropurchase-staging
+  memory: 1G
+  instances: 1
+  domain: 18f.gov
+  host: micropurchase-staging
+  buildpack: https://github.com/cloudfoundry/ruby-buildpack.git
```
stvnrlly: We should use `ruby_buildpack` directly here.

manifest.yml
```diff
@@ -0,0 +1,21 @@
+---
+applications:
+- name: micropurchase
+  memory: 1G
+  instances: 2
+  domain: 18f.gov
+  host: micropurchase
+  buildpack: https://github.com/cloudfoundry/ruby-buildpack.git
```
stvnrlly: We should use `ruby_buildpack` directly here.

manifest-staging.yml
jessieay: what do you mean by that? isn't this the ruby buildpack?

jessieay: via Slack:

> i think that using `ruby_buildpack` will use the buildpack as installed on cloud.gov, as opposed to the cloud foundry version
> since the docs use both i’d say just leave it

stvnrlly:
Looks good to me!

PR #1002, opened 2016-08-03
Auto-set New Relic env vars on app load

### LINKED ISSUE #998: Fix broken deploy
updated deploy to use 18f-cli and now it is broken. Steven Reilly is working on it. Until then, deploying via Travis is broken.

Once we get it fixed, we should also add docs around how to deploy manually using new methods

jessieay:
* Closes #998
* New Relic expects env vars named `NEW_RELIC_APP_NAME` and
  `NEW_RELIC_LICENSE_KEY`
* Inspired by https://github.com/18F/discourse/blob/cloud-gov-stage/config/load_cf_env.rb


jessieay:
cc @stvnrlly -- also, I disagree with this code climate error!

stvnrlly:
Ah, I was under the impression that "environment variable" meant that it had to be set on the server itself with `export NEW_RELIC_THING=whatever` before the app launches, but this makes a lot more sense.

I don't necessarily know what I'm talking about here, but is it possible to extend `UserProvidedService` again instead of using a new approach? Something where if `ENV['NEW_RELIC_THING'` doesn't exist, `ENV['NEW_RELIC_THING'] = credentials('new-relic')['thing']`?

jessieay:
@stvnrlly as far as I know, there is not a way to create `ENV` vars in that way. New Relic will look for those values and if they are not present, it will not log data to NR for us.

One thing we could do to make this more in line with other env vars is explicitly set the values on `newrelic.yml` using a credentials class. See https://docs.newrelic.com/docs/agents/ruby-agent/installation-configuration/ruby-agent-configuration

you think that would be better?

jessieay:
@stvnrlly updated -- lmk what you think!

stvnrlly:
I like this approach! Should `app_name` in `newrelic.yml` be configured similarly?

jessieay:
@stvnrlly good call! updated at 3ec198e

adelevie:
LGTM

stvnrlly:
🚀

PR #1005, opened 2016-08-04
Refactor use of factory pattern in app

### LINKED ISSUE #997: Make factory pattern usage more consistent
See https://github.com/18F/micropurchase/pull/993#discussion_r73405681

jessieay:
* Use base class to define interface for duck types
* Consolidate presenter logic -- woo hoo!
* Closes #997


app/models/status_presenter_factory.rb
```diff
@@ -6,7 +6,7 @@ def initialize(auction)
   end
 
   def create
-    "StatusPresenter::#{status_name}".constantize.new(auction)
+    Object.const_get("StatusPresenter::#{status_name}").new(auction)
```
harrisj: Just curious, is this a Rubocop-directed change?

app/presenters/bid_status_presenter/available_user_is_admin.rb
```diff
@@ -0,0 +1,17 @@
...
+    @auction = auction
+  end
+
+  def body
+    "This auction is accepting bids until #{end_date}."
+  end
+
+  private
+
+  def end_date
```
harrisj: Do you think this would make sense to define in BidStatusPresenter::Base?

app/presenters/bid_status_presenter/future_user_is_admin.rb
```diff
@@ -0,0 +1,31 @@
...
+    "This auction is visible to the public but is not currently accepting bids.
+    It will open on #{start_date}. If you need to take it down for whatever
+    reason, press the unpublish button below. #{link}"
+  end
+
+  private
+
+  def link
+    link_to(
+      "Un-publish",
```
harrisj: I always wonder if we should put link_to or other `view` things in Presenters or try to put them up in ViewModels. It's a hard call though, since the latter often involves some sort of weird checking or stuff that belongs within the Presenter

app/presenters/bid_status_presenter/future_user_is_admin.rb
```diff
@@ -0,0 +1,31 @@
...
+      admin_auction_published_path(auction),
+      method: :patch,
+      class: 'usa-button usa-button-outline auction-button'
+    )
+  end
+
+  def start_date
```
harrisj: This could also be moved into the base potentially

harrisj:
Looks good, but I wonder if we can eliminate some of the duplication by moving some methods that are copied multiple times into different classes from the children to the `Base` parent class? What do you think about that.

app/models/status_presenter_factory.rb
jessieay: no, just trying to make this more consistent with https://github.com/18F/micropurchase/blob/develop/app/models/sam_status_presenter_factory.rb#L9

app/presenters/bid_status_presenter/future_user_is_admin.rb
jessieay: yes I thought about that! Tricky bc not used in *all* presenters but would be nice to remove duplication....

app/presenters/bid_status_presenter/future_user_is_admin.rb
jessieay: yeah one option would be to include a link partial with the button in it....what do you think?

jessieay:
@harrisj good idea! will do so shortly when I finish some of the unrelated stuff I am working on for MP

jessieay:
@harrisj made some updates at 2dd261f -- LMK what you think!

app/models/status_presenter_factory.rb
harrisj: Cool

app/presenters/bid_status_presenter/future_user_is_admin.rb
harrisj: I dunno, I will admit that neither option makes me too thrilled, so this is fine for now

harrisj:
LYK what I think: LGTM. Now just Y U SO SLOW, Travis?

harrisj:
Oh, it's Jasmine hanging. Going to just merge it because I trust it

Issue #996, opened 2016-08-03
Move copy to i18n


jessieay:
See https://github.com/18F/micropurchase/pull/993#discussion_r73394035

mtorres253:
I love this idea @jessieay and I think it would make it easier for us to adjust the copy (we won't have to bug you to do it)

mtorres253:
Maybe we should do this now before all the other status messaging stories?

jessieay:
sure we can prioritize this!

mtorres253:
Cool @jessieay It's at the top of the backlog. We can discuss at planning on monday.

mtorres253:
@jessieay is doing this in pieces, so touch base with her @harrisj and @adelevie to see if there are any pieces you want to take on

mtorres253:
This is an ongoing thing and everyone is aligned on this refactoring. When we refactor other parts of the project, we implement this. Closing this issue for now

PR #1011, opened 2016-08-05
Start to I18n-ify all the things

jessieay:
* Working my way through cucumber specs first
* Part of #996


adelevie:
LGTM :dragon_face:

PR #1020, opened 2016-08-08
Vendor messaging - Auction show page - Delivery pending

jessieay:
* Closes #924

Winning vendor, no delivery URL submitted:

screen shot 2016-08-08 at 11 20 50 am

Winning vendor, delivery URL submitted:

screen shot 2016-08-08 at 12 25 26 pm

PR #1021, opened 2016-08-08
Winning vendor delivers auction

### LINKED ISSUE #1000: Winning vendor delivers auction
As a winning vendor, I want to let 18F know when I'm done with the job, so there should be a way for me to do that via the UI

- [ ] **Scenario: Auction show page: Winning vendor marks an auction as being delivered**
- And I have won an auction
- And there is a Delivery URL for that auction
- And I am on the auction show page https://micropurchase.18f.gov/auctions/[id]
- When I click "I'm Done" in the status box
- Then the system should record that my work is ready to be evaluated
- And I should see the following messaging in the status box

> Pending acceptance
> 18F is working to determine if your pull request meets the auction’s acceptance criteria. You will receive an email once this determination is made.

- And "your pull request" should link to the Delivery URL for that auction

screen shot 2016-08-03 at 2 51 14 pm

## Notes
- What if the vendor doesn't have a payment URL yet? How do we want to handle this case?

- Status box appears in the top right hand column of the show page

screen shot 2016-08-03 at 2 53 30 pm

jessieay:
Merge AFTER https://github.com/18F/micropurchase/pull/1020
* Closes #1000 

after marking as done:

screen shot 2016-08-08 at 3 11 45 pm

### DIFF
--- a/app/models/auction.rb
+++ b/app/models/auction.rb
@@ -10,7 +10,7 @@ class Auction < ActiveRecord::Base
   has_many :bidders, through: :bids
   has_and_belongs_to_many :skills
 
-  enum result: { not_applicable: 0, accepted: 1, rejected: 2 }
+  enum result: { not_applicable: 0, accepted: 1, rejected: 2, pending_acceptance: 3 }
   enum type: { sealed_bid: 0, reverse: 1 }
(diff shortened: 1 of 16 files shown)


jessieay:
@adelevie rebased this one too

PR #1024, opened 2016-08-09
Update en.yml

mtorres253:
Adding copy to the winner email letting them know to submit their pull request and "start work" when they're ready.


andrewmaier:
I'd like @jessieay to confirm that we can use the `%{auction_url}` variable! This now matches the prototype: https://pages.18f.gov/micropurchase/auctions/show/?auctionWorkflowState=ready-for-work-vendors

PR #1027, opened 2016-08-09
Migrate footer markup and styles from prototype

andrewmaier:
Proof of concept. Part of a larger project of bringing our production and prototype styles in line to streamline reuse.

1. Updated `_footer.html.erb` to align markup
2. Updated USWDS_overrides to reuse existing CSS (converted to SCSS).

cc @mtorres253 @jessieay

### DIFF
--- a/vendor/assets/stylesheets/uswds_overrides/components/_footer.scss
+++ b/vendor/assets/stylesheets/uswds_overrides/components/_footer.scss
@@ -1,143 +1,15 @@
-.site-footer{
-  html, body , div, span, applet, object, iframe,
-  h1, h2, h3, h4, h5, h6, p, blockquote, pre,
-  a, abbr, acronym, address, big, cite, code,
-  del, dfn, em, img, ins, kbd, q, s, samp,
-  small, strike, strong, sub, sup, tt, var,
-  b, u, i,
-  dl, dt, dd, ol, ul, li,
-  fieldset, form, label, legend,
-  table, caption, tbody, tfoot, thead, tr, th, td,
-  article, aside, canvas, details, embed,
-  figure, figcaption, footer, header, menu, nav,
-  output, ruby, section, summary,
-  time, mark, audio, video {
-    margin: 0;
-    padding: 0;
-    border: 0;
-    font-size: 100%;
-    font: inherit;
-    vertical-align: baseline;
+.usa-footer-big {
+  .usa-footer-primary-link {
+    background-image: image-url('uswds/arrow-down.png');
+    background-image: image-url('uswds/arrow-down.svg');
   }
 
-  * { box-sizing: border-box; }
-  *:before, *:after { display: none; }
-
-  font-family: Source Sans Pro, Verdana, sans-serif;
-  font-size: 18px;
-  font-weight: 400;
-  background: #2b3e5c;
-  color: #fff;
-
-  padding-top: 18px;
-  padding-bottom: 36px;
-
-  .wrapper{
-    width: 95%;
-    max-width: 960px;
-    margin: 0 auto;
-    position: relative;
+  ul {
+    &.hidden {
+      .usa-footer-primary-link {
+        background-image: image-url('uswds/arrow-right.png');
+        background-image: image-url('uswds/arrow-right.svg');
+      }
+    }
   }
-
-  .wrapper::after{
-    content: "";
-    display: table;
-    clear: both;
-  }
-
-
-  .logo a,
-  .logo a img{ float: left; }
-
-  .logo{ margin-bottom: 16px; }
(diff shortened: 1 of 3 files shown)


andrewmaier:
@jessieay Mind if I convert this component to slim? http://slim-lang.com/

jessieay:
@andrewmaier I think we are going to stick with ERB for now since that is more familiar for the rest of the team

vendor/assets/stylesheets/uswds_overrides/components/_footer.scss
```diff
@@ -1,15 +1,143 @@
-.usa-footer-big {
-  .usa-footer-primary-link {
-    background-image: image-url('uswds/arrow-down.png');
-    background-image: image-url('uswds/arrow-down.svg');
+.site-footer{
+  html, body , div, span, applet, object, iframe,
```
jessieay: this feels a bit excessive to me -- can we override fewer elements?

andrewmaier:
@jessieay Okay! I highly recommend it — or HAML :)

jessieay:
Yes I have used slim and HAML in the past, the team decided on erb before I joined

jessieay:
Looks like this is not using the web design standards at all but instead overrides everything -- didn't we discuss using an approach that builds on top of the standards rather than overriding them?

vendor/assets/stylesheets/uswds_overrides/components/_footer.scss
andrewmaier: I'm thinking I'll do this in each of the sections until we actually change some of the defaults. For example, this site shouldn't use white as the background color; it should use the dark blue.

vendor/assets/stylesheets/uswds_overrides/components/_footer.scss
```diff
@@ -1,15 +1,143 @@
...
-      .usa-footer-primary-link {
-        background-image: image-url('uswds/arrow-right.png');
-        background-image: image-url('uswds/arrow-right.svg');
-      }
-    }
+  * { box-sizing: border-box; }
+  *:before, *:after { display: none; }
+
+  font-family: Source Sans Pro, Verdana, sans-serif;
+  font-size: 18px;
```
jessieay: these feel like magic numbers -- can we use sass vars?

andrewmaier:
This is definitely building on top of them, but I'm intentionally going about it in a conservative way. 

I'm hoping to hold off on some of the more substantive lifts (changing, for example, the background color or the `$site-max-width` variable until we're farther along in the migration.

jessieay:
Everything in `vendor` comes directly from USWDS -- we are overriding that in `app` -- can we include any custom scss in `app` and remove all changes from `vendor` dir?

vendor/assets/stylesheets/uswds_overrides/components/_footer.scss
andrewmaier: Absolutely! I can the `base-font-size` to 18px (to match the prototype), but that might break other things across the site.

andrewmaier:
@jessieay I'm not sure what you mean. Don't overrides come from the `uswds_overrides` directory?

jessieay:
No sorry I am sure this is confusing -- `vendor` is for code from other places. Ideally we'd have zero overrides in there but rails doesnt work well with some of the uswds code so we need to override for it to work at all. Our actual customizations are in `app/assets`, which is a standard rails way of dealing with assets

jessieay:
This is not using the uswds at all except for basic layout -- design standards footer styles are only applied if you use `class="usa-footer usa-footer-slim"`

andrewmaier:
@jessieay I don't follow that logic? We don't have to use the 2-3 layouts provided by the standards.

jessieay:
We discussed this over google hangouts.

**Points made in conversation:**
- the CSS submitted here does not use the USWDS CSS. It overrides all of the vendored assets in order to style the footer but it is inspired by the USWDS CSS
- Jessie philosophically thinks we should be using the USWDS (the stylesheets, not just as inspiration)
- Andrew thinks that the USWDS framework is too constraining when creating custom-design apps
- The CSS written for [the prototype](https://pages.18f.gov/micropurchase/) is all hand-written, so making the app look like the prototype would require not using the USWDS
- Because we do not have a front end developer on this project, Jessie is worried about maintaining and updating hand-written CSS that is not broken down into reusable components

**Follow up:**
- leave this PR open for now but do not merge
- Andrew is going to follow up with Dave et al on whether we should be approaching this differently (eg: not use USWDS)

jessieay:
I am going to close this for now but will keep the branch live in case we want to re-visit soon

PR #1030, opened 2016-08-10
Update README.md

mtorres253:
Updating link to API docs to point to the new location


jessieay:
good one thanks! we also link to the docs in the site footer FYI

PR #1034, opened 2016-08-10
Vendor messaging - Auction show page - Need verified DUNS number to bid

### LINKED ISSUE #1007: Vendor messaging - Auction show page - Need DUNS number to bid
As a vendor, I want to know if there are factors preventing me from bidding on an auction so that if I want to bid on it, I can address those factors. So if I'm prevented from bidding on an auction because I don't have a DUNS number, then I should see that message

- [ ] **Scenario: Vendor sees "DUNS number needed" message on auction show page**
- Given I am a registered vendor
- And I am viewing an auction that is coming soon, or is open
- And I don't have a valid DUNS number
- When I go to the auction show page https://micropurchase.18f.gov/auctions/[id]
- Then no longer see the flash message reminder to update my duns number
- And I should see the following message in the status box

> Please update your profile
> Your profile must contain a valid DUNS number before you can place a bid. Please update your profile.

screen shot 2016-08-03 at 7 10 47 pm
 
- And "update your profile" should link to that vendor's profile

## Notes
- This story is to just move and update existing messaging so that it appears in the new status box 

- Status box should be at the top of the right hand column

screen shot 2016-08-03 at 7 11 36 pm

jessieay:
* Closes #1007

For an open auction, a signed in vendor with either no DUNS or a non-verified DUNS will see this:

screen shot 2016-08-10 at 1 05 01 pm

### DIFF
--- a/app/view_models/auction_show_view_model.rb
+++ b/app/view_models/auction_show_view_model.rb
@@ -76,7 +76,8 @@ class AuctionShowViewModel
     if !available? ||
        user_not_vendor? ||
        reverse_auction_available_user_is_winner? ||
-       sealed_bid_auction_user_is_bidder?
+       sealed_bid_auction_user_is_bidder? ||
+       current_user.sam_status != 'sam_accepted'
       'auctions/bid_status'
     else
       'components/null'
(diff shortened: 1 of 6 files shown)


app/view_models/auction_show_view_model.rb
```diff
@@ -76,7 +76,8 @@ def bid_status_partial
     if !available? ||
        user_not_vendor? ||
        reverse_auction_available_user_is_winner? ||
-       sealed_bid_auction_user_is_bidder?
+       sealed_bid_auction_user_is_bidder? ||
+       current_user.sam_status != 'sam_accepted'
```
jessieay: this conditional is getting super ugly -- will make it better once we have a status for every type of user

PR #1041, opened 2016-08-15
Quick pass at caching the insights page

adelevie:
An attempt at page caching `/insights` using https://github.com/rails/actionpack-page_caching.

To test this locally, set `config.action_controller.perform_caching = true` in `config/environments/development.rb`, run the server and visit `/insights`. Check to see that `public/cache/insights.html` exists. To force cache expiration, open a rails console and run `ClearCache.new.perform` and check that `public/cache/insights.html` no longer exists.


app/services/clear_cache.rb
```diff
@@ -0,0 +1,5 @@
+class ClearCache
+  def perform
+    ApplicationController.expire_page(app.insights_path)
```
jessieay: cool! have we confirmed that this works?

config/clock.rb
```diff
@@ -18,4 +18,9 @@ module Clockwork
     puts "Checking for approved auctions"
     CheckApproval.new.delay.perform
   end
+
+  every(1.day, "insights_cache.clear") do
+    puts "Clearing the page cache for the Insights page"
+    ClearCache.new.delay.perform
```
jessieay: niiiice

jessieay:
looks good! code climate looks like an easy fix

adelevie:
@jessieay I confirmed this locally from a manual test, but does this seem like something that needs an automatic test? E.g. I have not yet simulated the clock.rb stuff, just ran `ClearCache#perform`.

jessieay:
@adelevie I do think one thing that could be cool here would be to add an integration test that looks for a stat, changes the auction data, clears cache, and looks for new stat. thoughts?

spec/services/clear_cache_spec.rb
```diff
@@ -0,0 +1,7 @@
+# require 'rails_helper'
```
jessieay: delete?

spec/factories/auctions.rb
```diff
@@ -116,6 +116,7 @@
     end
 
     trait :accepted do
+      ended_at { Time.now - 1.days }
```
jessieay: what is this related to?

features/support/env.rb
```diff
@@ -23,6 +23,16 @@
   Delayed::Worker.delay_jobs = false
 end
 
+Before '@enable_caching' do
+  ActionController::Base.perform_caching = true
```
jessieay: interesting -- is this not usually enabled in test env?

features/step_definitions/auction_create_steps.rb
```diff
@@ -174,11 +174,11 @@
...
-    :published,
-    result: :accepted,
-    accepted_at: nil,
-  )
+  @auction = FactoryGirl.create(:auction, :accepted, :with_bidders)
+end
+
+When(/^there is another accepted auction$/) do
+  # not creating another @auction-like instance variable
+  # just creating the record without saving to a variable
```
jessieay: I think think these comments are needed

spec/services/clear_cache_spec.rb
adelevie: sorry, still sketching ... I'm going to add a simple unit test, too.

spec/factories/auctions.rb
adelevie: For seed data to be displayed in /insights, I thought the auction needs to have ended to be counted. But `result: :accepted` might be sufficient here.

features/support/env.rb
adelevie: as of now, `config/environments/test.rb` does not have `perform_caching` anywhere. Seems safer to explicitly test only where needed? Or maybe have test match prod and apply caching globally in the test env?

jessieay: gotcha -- this seems oK for now!

spec/factories/auctions.rb
adelevie: @jessieay I think also, as a rule, wouldn't an accepted auction always have an ended_at timestamp? And if so, shouldn't the factory reflect this?

adelevie:
Not as great as I'd like it to be, but it's ready imo, and we should at least validate that this will work on staging before much more work is done.

spec/services/clear_cache_spec.rb
jessieay: did you want to write a unit test or not?

jessieay:
looks good @adelevie  -- can we just delete that empty spec file?

adelevie:
good catch--deleted.

jessieay:
merging!! 🌈

adelevie:
yay!

PR #1058, opened 2016-08-17
Only run analytics on prod

jessieay:
* Noticed while running dev offline that we were running g-analytics in
  all envs
* Could also run on staging but assume we don't want to?


adelevie:
very nice

PR #1059, opened 2016-08-17
Refactoring and renaming

jessieay:
* Rename status fields on auction
* Clarify and clean up naming of some view model methods


adelevie:
flexible enums ftw, lgtm

PR #1061, opened 2016-08-18
Fix typo in .env.example

divergentdave:
There was a typo in C2_HOST in .env.example which, when copied to .env, would cause the db:seed rake task to fail. Hardcoded sample C2 URLs are checked against the value of C2_HOST from .env.


adelevie:
Good catch!

PR #1064, opened 2016-08-18
Calculate savings based on complete AND successful auctions

### LINKED ISSUE #1037: Savings number on insights page seems too high?
The savings number on the insights page should be the difference between

- The sum of the starting prices for all successful auctions
- The sum of the total cost (award amount) for all successful auctions (auctions that have been paid)

Can we verify the number is correct? It should be around $20K I think

jessieay:
* Previously was looking at all complete auctions, resulting in ~$61k on prod
* Closes #1037

### DIFF
--- a/spec/models/statistics/total_savings_spec.rb
+++ b/spec/models/statistics/total_savings_spec.rb
@@ -2,7 +2,7 @@ require 'rails_helper'
 
 describe Statistics::TotalSavings do
   describe '#to_s' do
-    context 'there are completed auctions' do
+    context 'there are complete and successful auctions' do
       it 'returns correct amount' do
         auction = create(:auction, :complete_and_successful, start_price: 10)
         second_auction = create(:auction, :complete_and_successful, start_price: 20)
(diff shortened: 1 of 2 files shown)


spec/models/statistics/total_savings_spec.rb
```diff
@@ -2,7 +2,7 @@
 
 describe Statistics::TotalSavings do
   describe '#to_s' do
-    context 'there are completed auctions' do
+    context 'there are complete and successful auctions' do
```
harrisj: Can we create an extra auction that is complete but not successful to ensure it isn't counted?

jessieay: that is what I am doing in the spec below with `create(:auction, :completed, :with_bidders)` -- that factory is for an auction that has delivery past due but is no accepted. You think we should include more?

This brings up a good point, though, which is that I think it's time for us to clean up some of our factories. They are well-named in some cases but I frequently find myself visiting the factory file to understand what a factory is. Making an issue for that now

harrisj: Oh sorry. Yes, we need to clean up the factories in `auctions.rb` since they are sometimes a bit redundant and unclear

jessieay:
@harrisj you think this is ready?

harrisj:
Ooops yeah, let's merge

PR #1068, opened 2016-08-19
Use a more consistent naming convention for bid status presenters

jessieay:
* First namespace is auction status
* Second / third  namespace, if any, is the user type
* Only use role namespace if there are multiple presenters for that user type

### DIFF
--- a/app/models/bid_status_presenter_factory.rb
+++ b/app/models/bid_status_presenter_factory.rb
@@ -22,67 +22,67 @@ class BidStatusPresenterFactory
 
   def future_message
     if admin?
-      BidStatusPresenter::FutureUserIsAdmin
+      BidStatusPresenter::Future::Admin
     elsif guest?
-      BidStatusPresenter::FutureUserIsGuest
+      BidStatusPresenter::Future::Guest
     else
-      BidStatusPresenter::FutureUserIsVendor
+      BidStatusPresenter::Future::Vendor
     end
   end
 
   def over_winning_bidder_message
     if auction.pending_acceptance?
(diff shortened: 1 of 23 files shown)


app/models/bid_status_presenter_factory.rb
```diff
@@ -22,67 +22,67 @@ def create
...
     else
-      BidStatusPresenter::OverUserIsWinner
+      BidStatusPresenter::Over::Vendor::Winner::WorkNotStarted
     end
   end
 
   def over_message
     if user_is_bidder?
-      BidStatusPresenter::OverUserIsBidder
+      BidStatusPresenter::Over::Vendor::Bidder
```
harrisj: Holy namespaces, Batman!

harrisj:
At first, I found myself a bit 😕 about five layers of namespacing for some classes, but it really is equivalent to having a long and unwieldy class name, so I support this. But I want to give others a chance to weigh in

jessieay:
Yes it is a bit annoying when opening up a file, but the nice thing about namespacing like this is that it forces consistency because you don't want to create all kinds of dirs for different types of presenters that are for the same user type etc. Imagining that the next version of this will include some type of interpolation based on the auction status and person viewing the auction

harrisj:
Just make sure it's automagical interpolation and loading to keep it Railsy

jessieay:
ping ping @adelevie , jacob is requesting your 👀 on this one

adelevie:
Yeah I think this is fantastic. If stuff is unwieldy, better to name it and make it clear than to hide it in longer module/class names.

PR #1074, opened 2016-08-19
Add null states for needs attention dashboard

jessieay:
* There was a bullet about this in each story for these sections

screen shot 2016-08-19 at 2 54 00 pm


harrisj:
Nice!

PR #1075, opened 2016-08-19
Fix admin auction pagination

jessieay:
* Must use same instance var as kaminari does in order for pagination to
  work

### DIFF
--- a/app/view_models/admin/auctions_index_view_model.rb
+++ b/app/view_models/admin/auctions_index_view_model.rb
@@ -1,6 +1,12 @@
 class Admin::AuctionsIndexViewModel < Admin::BaseViewModel
-  def auctions
-    Auction.all.order(started_at: :desc).map do |auction|
+  attr_reader :auctions
+
+  def initialize(auctions: Auction.all.order(started_at: :desc))
+    @auctions = auctions
+  end
+
+  def auction_view_models
+    auctions.map do |auction|
       Admin::AuctionListItem.new(auction: auction)
     end
   end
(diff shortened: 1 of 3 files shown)


app/view_models/admin/auctions_index_view_model.rb
```diff
@@ -1,6 +1,12 @@
 class Admin::AuctionsIndexViewModel < Admin::BaseViewModel
-  def auctions
-    Auction.all.order(started_at: :desc).map do |auction|
+  attr_reader :auctions
+
+  def initialize(auctions: Auction.all.order(started_at: :desc))
```
harrisj: I wonder if we should put that default argument into a method like `default_auctions` so the args will look like `auctions: default_auctions`

app/views/admin/auctions/index.html.erb
```diff
@@ -1,7 +1,7 @@
 <%= render partial: 'admin/auctions_subnav', locals: { view_model: @view_model } %>
 
 <div class="admin-auction-wrapper auction-wrapper">
-  <% @view_model.auctions.each do |auction| %>
+  <% @auctions.each do |auction| %>
```
harrisj: Is this change because of Kaminari?

app/view_models/admin/auctions_index_view_model.rb
jessieay: not this exact one, was just trying to make this view mode more consistent with `AuctionsIndexViewModel` (non-admin)

app/views/admin/auctions/index.html.erb
jessieay: yes! this was why the pagination wasn't working before in the admin section

harrisj:
Okay. LGTM

PR #1076, opened 2016-08-20
Fix typo

divergentdave:


adelevie:
Another great catch from Mr. Dave!

PR #1078, opened 2016-08-22
Update clock.rb to check c2 every 10 minutes

jessieay:
* Use environment variable so we can easily change the interval


adelevie:
LGTM

harrisj:
You going to merge it then?

PR #1081, opened 2016-08-22
Delivery due at is required field

jessieay:
* There is no way to create an auction without this field via the form
* Removes chance of `nil` errors

### DIFF
--- a/app/models/auction.rb
+++ b/app/models/auction.rb
@@ -34,7 +34,7 @@ class Auction < ActiveRecord::Base
   self.inheritance_column = :__disabled
 
   validate :user_is_contracting_officer_if_above_micropurchase
-  validates :delivery_due_at, presence: true, if: :published?
+  validates :delivery_due_at, presence: true
   validates :description, presence: true, if: :published?
   validates :ended_at, presence: true
   validates :purchase_card, presence: true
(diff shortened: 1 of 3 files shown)


harrisj:
I think the reason why we allowed delivery_due_at to be nil was that we could work on drafts without having to set the time the delivery is due by. Otherwise, there is the temptation to just give it some value so you can save and then forgetting why you set it to that specific date later.

jessieay:
@harrisj yes  I think you are right about that but the way the form is set up now, admins actually cannot set the `delivery_due_at` timestamp to a nil value. So the only nil values we have are from console-created data.

harrisj:
Oh okay. Sometimes I wonder if we should allow people to set the delivery due at as a days offset and then actually evaluate it when the auction is published, but I might be overthinking this

harrisj:
Okay, going to merge then.

jessieay:
@harrisj that is what we do! (kind of -- this is just on the new form, not edit)
screen shot 2016-08-23 at 9 16 28 am

harrisj:
I know we do that, but we evaluate it immediately on save. It's kinda silly, but I sometimes wonder if we should just make end_time and delivery_due_at offsets that we store until the auction is published. Then they become actual datetimes.

Issue #1015, opened 2016-08-08
Admin messaging - Auction show page - Reject work


mtorres253:
As an admin, I want a way to easily mark an auction as rejected, without having to go into the edit form. So I should be able to do this on the show page of an auction

- [x] **Scenario: Admin marks an auction as rejected from the show page**
- Given I am an authenticated admin
- And a vendor has delivered code for an auction (clicked the "I'm Done" button)
- When I view the show page for that auction
- Then I should see the following messaging in the status box

> Pending acceptance
> [auction.winner] made a pull request and marked this auction as having been delivered. If their work meets this auction’s acceptance criteria, click “accept.” Otherwise, click “reject.”

- And I should see an "Accept" and a "Reject" button
- And "[Auction.Winner]" should link to the winning vendor's profile (https://micropurchase.18f.gov/admin/users/[id])
- And "pull request" should link to the Delivery URL for that auction

<img width="334" alt="screen shot 2016-08-08 at 6 45 48 am" src="https://cloud.githubusercontent.com/assets/100149/17482161/55612680-5d35-11e6-8115-c3fc4a094701.png">

- [x] **Scenario: Admin clicks "reject" button on auction show page**
- Given I am an authenticated admin
- And I am on the show page for an auction that has work delivered (the vendor has clicked "I'm Done")
- When I click to the "Reject" button
- Then the auction should be marked as rejected by me
- And I should see the following messaging in the status box

> Rejected
> [Vendor.name]'s work was rejected on [auction.rejectionDate].

- And "[Vendor.name]'s work" should link to the delivery URL

## Notes

- Status box appears at the top right of the auction show page

<img width="992" alt="screen shot 2016-08-08 at 6 59 48 am" src="https://cloud.githubusercontent.com/assets/100149/17482248/bc1daad8-5d35-11e6-8c60-c0bea497f373.png">

andrewmaier:
I believe we spoke yesterday about creating a separate status if the person who rejected this work is viewing it? The current copy is

>[RejectingAdmin] rejected [auction.winner]’s bid on [auction.rejectionDate].

@mtorres253 I believe you asked to split this into 
> You rejected ...
and 
> Someone rejected

Let me know what you decide and I can update the prototype/workflows.

jessieay:
waiting for merge of #1003 before I start on this

PR #1088, opened 2016-08-22
Add Pending Acceptance messaging in admin section

jessieay:
* Next step is to add buttons / functionality
* Part of https://github.com/18F/micropurchase/issues/1015

screen shot 2016-08-22 at 3 38 44 pm

### DIFF
--- a/app/view_models/admin/auction_show_view_model.rb
+++ b/app/view_models/admin/auction_show_view_model.rb
@@ -15,7 +15,8 @@ class Admin::AuctionShowViewModel < Admin::BaseViewModel
   end
 
   def c2_status_partial
-    if auction.purchase_card == 'default' && auction.c2_status != 'approved' || auction.pending_acceptance?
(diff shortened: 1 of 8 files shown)


app/view_models/admin/auction_show_view_model.rb
```diff
@@ -15,7 +15,7 @@ def csv_report_partial
   end
 
   def c2_status_partial
-    if auction.purchase_card == 'default' && auction.c2_status != 'approved'
+    if auction.purchase_card == 'default' && auction.c2_status != 'approved' || auction.pending_acceptance?
```
harrisj: I know it's not required, but I think I would like a () around the `auction.c2_status != 'approved' || auction.pending_acceptance?`

jessieay: done! at 720c945

adelevie:
LGTM

PR #1089, opened 2016-08-22
Update styles to prevent ribbon and title overlap

jessieay:
* Seeing this on currently live auction

show page before and after (not more ribbon overlap with title):

screen shot 2016-08-22 at 4 31 32 pm

index page before and after (no change):

screen shot 2016-08-22 at 4 27 49 pm


harrisj:
CSS!

Looks good to me

PR #1091, opened 2016-08-23
Display number of auctions that need attention

### LINKED ISSUE #1048: Total number of auctions shows in the auctions subnav
As an admin, I'd like to know how many auctions need my attention, so I should see an alert in the auctions section

- [x] **Scenario: Admin sees the total number of auctions needing attention in the auctions section subnav**
- And there are auctions that are either in draft state, pending delivery, needs acceptance, or needs payment (#939, #950, #948, #947)
- When I am in the auctions section (https://micropurchase.18f.gov/admin/auctions/)
- Then I should see the total number of auctions needing my attention next to the "Needs Attention" link in the subnav (see screenshot below)

screen shot 2016-08-16 at 10 40 08 am

harrisj:
Fixes #1048

localhost_3000_admin_auctions_needs_attention

This adds some functionality to display the number of auctions that need attention in the subnav of the admin interface for auctions. This needs some review of the CSS, etc.

### DIFF
--- a/app/assets/stylesheets/components/_secondary_nav.scss
+++ b/app/assets/stylesheets/components/_secondary_nav.scss
@@ -44,4 +44,12 @@ $height-nav-secondary: 5rem;
       font-weight: $font-bold;
     }
   }
+
+  .subnav-number-label {
+    border-radius: $border-radius;
+    margin-right: 0.3rem;
+    padding: 0.1rem 0.5rem;
+    background-color: $color-secondary-darkest;
+    color: $color-white;
+  }
 }
--- a/app/assets/stylesheets/components/_secondary_nav.scss
+++ b/app/assets/stylesheets/components/_secondary_nav.scss
@@ -46,9 +46,11 @@ $height-nav-secondary: 5rem;
   }
 
   .subnav-number-label {
-    border-radius: $border-radius;
-    margin-right: 0.3rem;
-    padding: 0.1rem 0.5rem;
+    border-radius: $round-border-radius;
+    font-size: 1.3rem;
+    margin-right: 0.5rem;
+    margin-left: 0.5rem;
+    padding: 0.1rem 0.7rem;
     background-color: $color-secondary-darkest;
     color: $color-white;
   }
--- a/app/view_models/admin/needs_attention_auctions_view_model.rb
+++ b/app/view_models/admin/needs_attention_auctions_view_model.rb
@@ -12,7 +12,7 @@ class Admin::NeedsAttentionAuctionsViewModel < Admin::BaseViewModel
   end
 
   def drafts
-    @_drafts ||= AuctionQuery.new.unpublished.map do |auction|
+    @_drafts ||= Auction.unpublished.map do |auction|
       Admin::DraftListItem.new(auction)
     end
   end
@@ -38,7 +38,7 @@ class Admin::NeedsAttentionAuctionsViewModel < Admin::BaseViewModel
   end
 
   def evaluation_needed
-    @_evaluation_needed ||= list_items(AuctionQuery.new.pending_acceptance)
+    @_evaluation_needed ||= list_items(Auction.pending_acceptance)
   end
 
   def payment_needed_partial
--- a/app/view_models/admin/needs_attention_auctions_view_model.rb
+++ b/app/view_models/admin/needs_attention_auctions_view_model.rb
@@ -12,7 +12,7 @@ class Admin::NeedsAttentionAuctionsViewModel < Admin::BaseViewModel
   end
 
   def drafts
-    @_drafts ||= Auction.unpublished.map do |auction|
+    @_drafts ||= AuctionQuery.new.unpublished.map do |auction|
       Admin::DraftListItem.new(auction)
     end
   end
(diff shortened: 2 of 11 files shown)


harrisj:
Specifically, I wasn't sure how much time/effort I should spend on rounding the corners of the label to match what is in the mockup. What do you all think?

spec/models/auction_query_spec.rb
```diff
@@ -120,6 +120,20 @@
     end
   end
 
+  describe '#needs_attention_count' do
+    it 'returns the sum of unpublished, pending_delivery, pending_acceptance and payment_needed auctions' do
```
jessieay: ❤️

app/views/admin/_auctions_subnav.html.erb
```diff
@@ -7,7 +7,11 @@
       class: view_model.auctions_nav_class %></li>
     <li><%= link_to t('links_and_buttons.auctions.needs_attention'),
       admin_auctions_needs_attention_path,
-      class: view_model.needs_attention_auctions_nav_class %></li>
+      class: view_model.needs_attention_auctions_nav_class %>
+      <% if view_model.needs_attention_auctions_count > 0 %>
```
jessieay: let's add this logic to the view model so we don't need to put logic in the view (use the `components/null` pattern)

app/assets/stylesheets/components/_secondary_nav.scss
```diff
@@ -44,4 +44,12 @@ $height-nav-secondary: 5rem;
       font-weight: $font-bold;
     }
   }
+
+  .subnav-number-label {
+    border-radius: $border-radius;
```
jessieay: if we made this a higher value, would it look more like the mock? I have certainly not been doing everything pixel perfect, but if it's just a matter of changing a value it would be cool to get it looking closer to the mock

harrisj: Oh yes, you are so right!

harrisj:
Thanks for the feedback!

app/view_models/admin/needs_attention_auctions_view_model.rb
```diff
@@ -12,7 +12,7 @@ def drafts_partial
   end
 
   def drafts
-    @_drafts ||= Auction.unpublished.map do |auction|
+    @_drafts ||= AuctionQuery.new.unpublished.map do |auction|
```
jessieay: not sure we need to call this on `AuctionQuery` -- we have it defined as a scope in `app/models/concerns/auction_scopes.rb` -- is there a good reason to define it on AuctionQuery too?

app/view_models/admin/needs_attention_auctions_view_model.rb
```diff
@@ -38,7 +38,7 @@ def evaluation_needed_partial
   end
 
   def evaluation_needed
-    @_evaluation_needed ||= list_items(Auction.pending_acceptance)
+    @_evaluation_needed ||= list_items(AuctionQuery.new.pending_acceptance)
```
jessieay: same comment as above

jessieay:
2 tiny comments -- new styles look 💯  ! you are a frontend developer!!

app/view_models/admin/needs_attention_auctions_view_model.rb
harrisj: I guess I was confused why we have some things in AuctionQuery and some directly in Auction. In the past I thought we moved all that stuff out of the model and now some of it is back in?

jessieay: we moved the code that used to be in that `Scope` module within `AuctionQuery` into `AuctionScopes` so the behavior of `AuctionQuery` was clearer since before we were doing that confusing `define_method` stuff https://github.com/18F/micropurchase/commit/5d8cfaec763b8027303f171f85b8ffc36ef3c8ca#diff-8802c0b39ff625b976e3dc520209d322

jessieay: I think the difference between `AuctionQuery` and `AuctionScope` in my mind is that `AuctionScope`

harrisj: Ah, got it. I just assumed that scopes and queries were handled by the Query object, since I assumed that the caller is abstracted away from knowing the internals of the Auction model (ie, whether something is a scope). But I also am way of too much abstraction for the sake of abstraction.

In short, I don't really have a particular axe to grind on this one, but we probably need to update our documentation about this.

jessieay: cool so shall I merge?

harrisj: Well do you want me to revert this back from using AuctionQuery? If we don't wrap scopes, I'm not necessarily inclined to do it for some

jessieay: sure, I do think using `Auction` rather than `AuctionQuery` would be more consistent, but I defer to you

harrisj: Well, I think it's a bit weird that sometimes I instantiate an AuctionQuery object and sometimes I call a class method on Auction but I. defer. to. you!

Seriously though, if we do decide to change this, it should not be in this PR so let me unroll

harrisj:
Okay, I reverted the changes to AuctionQuery to not wrap AuctionScopes. Thanks!

jessieay:
👍

PR #1092, opened 2016-08-23
Remove `running` trait from factories

jessieay:
* was just adding bids to the auction, which was not obvious based on the name
* we usually use the terms "future" "available" and "closed" to describe the temporal state of an auction


harrisj:
Yay!

PR #1097, opened 2016-08-24
Test that the API documentation renders correctly

### LINKED ISSUE #1082: Add controller test for /api docs method
I think this would be good to add (just that it would catch if it crashed for some reason)

harrisj:
Fixes #1082 

The `render_views` declarations forcibly renders the views and I tested that it would fail if I broke the template.


jessieay:
good one! merging

PR #1098, opened 2016-08-24
More factory cleaning

jessieay:
* Only include attributes that are actually required


harrisj:
This looks good, but it looks like it needs to be rebased

jessieay:
@harrisj rebased!

harrisj:
LGTM

Issue #1099, opened 2016-08-24
Vendor messaging - Payment URL needed - Auction Show Page


jessieay:
So that everyone knows that a payment URL is needed for a payment to be made, the status box messaging should reflect that a payment URL is needed.

- [ ] **Scenario: Winning vendor sees message in status box that payment URL is needed**
- Given I am a vendor who has won an auction
- And my work was accepted (the auction is marked as accepted)
- And I don't have a payment URL stored in my profile
- When I go to the auction show page (https://micropurchase.18f.gov/auctions/[id])
- Then I should see the following status in the status box for that auction

> Payment URL needed
Your pull request was accepted, but we cannot issue payment until you have specified a valid payment URL. Please update the payment URL field your profile, or contact us at micropurchase@gsa.gov with other instructions on how to pay you via credit card for this auction.

- And "pull request" should link to the delivery URL
- And "your profile" should link to the my profile page

## Notes
We have messaging for "this was accepted" 

https://github.com/18F/micropurchase/issues/1016

if it was accepted and the winning vendor does not have a payment url, they receive an email asking them to update their profile. In that case, we do not have an `accepted_at` timestamp because that is only set when the vendor updates their profile.

should we have specific messaging for the state between when an auction is accepted and the vendor provides a payment URL?

mtorres253:
Probably @jessieay. @andrewmaier let's talk about this one tomorrow

PR #1100, opened 2016-08-24
Admin can accept auction from admin/auction#show page

jessieay:
* Closes https://github.com/18F/micropurchase/issues/1016
* This raised an edge case, which I wrote up here https://github.com/18F/micropurchase/issues/1099

Before clicking accept:

screen shot 2016-08-24 at 11 40 22 am

After:

screen shot 2016-08-24 at 11 40 30 am

### DIFF
--- a/app/models/auction.rb
+++ b/app/models/auction.rb
@@ -20,7 +20,7 @@ class Auction < ActiveRecord::Base
   }
 
   enum status: {
-    not_applicable: 0,
+    pending_delivery: 0,
     pending_acceptance: 3,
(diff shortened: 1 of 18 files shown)


app/models/auction.rb
```diff
@@ -20,7 +20,7 @@ class Auction < ActiveRecord::Base
   }
 
   enum status: {
-    not_applicable: 0,
+    pending_delivery: 0,
```
harrisj: Much nicer name

harrisj:
LGTM

Issue #1083, opened 2016-08-22
add dev:prime task


jessieay:
right now we are using the seeds file to add dev data. seed data should be for a prod database. let's add that data to another task instead and just add a few skills or something to seeds.rb

harrisj:
I am trying so hard to avoid pasting an Optimus Prime GIF in this discussion

jessieay:
example task in Dolores rails app: https://github.com/18F/dolores-landingham-slack-bot/blob/master/lib/tasks/dev.rake

PR #1102, opened 2016-08-24
Add dev:prime task

jessieay:
* Use that in `setup` script
* This script can be re-run daily without worry (not indempodent, but
  not sure we want that anyway since it's good to get auctions at
  different states)
* Closes https://github.com/18F/micropurchase/issues/1083


jessieay:
tested this out locally a few times - 👌

harrisj:
LGTM

Issue #1018, opened 2016-08-08
Vendor email alert - auction delivery has been rejected


mtorres253:
As an admin, I want to be consistent about notifications to vendors whose work I have rejected and I want to do this automatically so I don't have to remember to do it, so the system should send an automated email letting the vendor know that their work was rejected

- [x] **Scenario: System sends automated notification to winning vendor that work was rejected**
- Given that an auction exists
- And the winning vendor has delivered code (clicked the "I'm Done" button)
- And the vendor has an email address
- When an admin rejects the work (clicks "Reject" on the show page, or sets the delivery status to "Rejected" on the auction edit form)
- Then an email is sent to the the vendor's email with this content

``` yaml
subject: "18F Micro-purchase Auction Status"
para_1: "We regret to inform you that your work for %{auction_title} was rejected
         on mm/dd/yyyy. For more information please see the auction details here  
         %{auction_url} or contact us at micropurchase@gsa.gov. Though this auction 
         did not work out, we hope you will continue to participate in the future."
sign_off: "Sincerely,"
from: "The 18F Micro-purchase team"
        
```

PR #1103, opened 2016-08-24
It's not you, it's me

jessieay:
* A rejection email
* Closes https://github.com/18F/micropurchase/issues/1018

screen shot 2016-08-24 at 2 39 13 pm

### DIFF
--- a/config/locales/mailers/en.yml
+++ b/config/locales/mailers/en.yml
@@ -60,3 +60,14 @@ en:
         para_2: "Again, great job! We look forward to working with you again in the future."
         sign_off: "Sincerely,"
         from: "The 18F Micro-purchase team"
+      auction_rejected:
+        subject: "18F Micro-purchase Auction Status"
+        para_1: >-
+          We regret to inform you that your work for %{auction_title} was
+          rejected on %{rejected_at}. For more information please see the GitHub issue:
+          %{issue_url} or contact us at micropurchase@gsa.gov. Please also let us
+          know if you would like us to schedule a retrospective to go over any questions
+          or concerns you may have about the auction. Though this auction did not work
+          out, we hope you will continue to participate in the future.
+        sign_off: "Sincerely,"
+        from: "The 18F Micro-purchase team"
--- a/app/controllers/admin/auction_rejections_controller.rb
+++ b/app/controllers/admin/auction_rejections_controller.rb
@@ -2,6 +2,7 @@ class Admin::AuctionRejectionsController < Admin::BaseController
   def update
     auction = Auction.find(params[:id])
     auction.update(status: :rejected, rejected_at: Time.current)
+    WinningBidderMailer.auction_rejected(auction: auction).deliver_later
     redirect_to admin_auction_path(auction)
   end
 end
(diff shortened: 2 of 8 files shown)


config/locales/mailers/en.yml
```diff
@@ -60,3 +60,12 @@ en:
         para_2: "Again, great job! We look forward to working with you again in the future."
         sign_off: "Sincerely,"
         from: "The 18F Micro-purchase team"
+      auction_rejected:
+        subject: "18F Micro-purchase Auction Status"
+        para_1: >-
+          We regret to inform you that your work for %{auction_title} was rejected
```
harrisj: This message suggests that we will provide a reason for rejection on the auction. Is that true?

harrisj:
So, this is an interesting thing, because while I am generally for automating things, I realize there are several reasons why we might reject an auction. What's weird is that in each of these cases, we would probably relay the reason _why_ we rejected the auction in a different way.

* No work submitted (this would just be in the email sent, check if auction rejected and no delivery_url)
* Work was submitted but substandard (here we would probably comment on the delivery_url PR)
* We can't actually pay the person because they are an excluded entity or such (I guess this would mean a special email)

Am I missing any other circumstances where a person would be rejected?

jessieay:
great points, @harrisj ! Looking to @mtorres253 for some direction on this issue...

mtorres253:
@jessieay @harrisj Yes I was just thinking about this today. It would be nice to add a comment or reason for the rejection. We can groom this on Monday and add some stories. For now we can handle manually (if necessary)

jessieay:
ok so should we not move forward with this? close #1018 without merging?

mtorres253:
@jessieay @harrisj I think we should merge this. At least they can get an email notification that the auction has been rejected. We don't need to systematize reasons for rejection at this point. It will probably be clear in the github issue why the deliverable was rejected.

jessieay:
ok @harrisj I will leave it up to you

adelevie:
I agree with @mtorres253. But @harrisj brings up a great point. Maybe we also say, feel free to email us back/get in touch and that we're always happy to discuss what didn't go well and what could be better. Also, maybe include a standing offer to retro???

Makes me also think: could/should our emails always include micropurchase@gsa.gov so that they become reply-able?

jessieay:
We have to send emails from an 18F domain (18f.gov) because of the way
mandrill works.

We do have micropurchsse@gsa as the reply to address on all emails, though

adelevie:
Couldn't we cc the @gsa.gov account? Or add it to the `to:` field? That way that account will also catch replies?

jessieay:
when you set the `reply-to` that is the address that is used for replies. Is there a different scenario you're thinking of that I am missing?

jessieay:
here is a very retro but informative overview of `reply-to` vs `from` http://emailuniverse.com/ezine-tips/?id=179

jessieay:
will wait to hear back from @mtorres253 on whether we should tweak copy before merging.

config/locales/mailers/en.yml
```diff
@@ -60,3 +60,12 @@ en:
...
+          on %{rejected_at}. For more information please see the auction details here
+          %{auction_url} or contact us at micropurchase@gsa.gov. Though this auction
```
mtorres253: For more information please see GitHub issue: %{link to GitHub issue}. Please also let us know if you would like us to schedule a retrospective to go over any questions or concerns you may have.

adelevie:
🍧

PR #1108, opened 2016-08-25
Change the failing feature step to use set var

harrisj:
For some reason, I think the issue has something to do with chaining two business time calculations back to back (maybe there is some rounding or other adjustment after save in the record). So I just add the business days offset for delivery to the `@end_day` variable. This seems to work, but I dunno


features/step_definitions/admin_auction_form_steps.rb
```diff
@@ -90,7 +90,7 @@
 
 Then(/^I should see the updated delivery deadline$/) do
   within('.estimated-delivery-date') do
-    delivery_date = DefaultDateTime.new(5.business_days.after(DcTimePresenter.convert(3.business_days.from_now))).convert
+    delivery_date = DefaultDateTime.new(5.business_days.after(@end_day)).convert
```
jessieay: it seems like this is reverting the last change you made to address this issue: https://github.com/18F/micropurchase/commit/1a18a17ddf3c7d0acee09937f9ed84d0579ea82f#diff-a20386f4c76aafe8471bb849a3b37f5f

Not opposed to revert, but since this issue popped up before that change was made, seems like this was likely not the culprit

config/initializers/business_time.rb
```diff
@@ -1,3 +1,3 @@
-Holidays.between(Date.today, 5.years.from_now, :us, :observed).map do
+Holidays.between(Date.today, 5.years.from_now, :us, :observed).each do
```
jessieay: why this change?

features/step_definitions/admin_auction_form_steps.rb
harrisj: Hmm, okay then i have nothing

config/initializers/business_time.rb
harrisj: I don't like using map when the goal is just iteration and not creating a new array

features/step_definitions/admin_auction_form_steps.rb
jessieay: ok, well I don't mind this reverting change then because at least we now know that this was neither the fix nor the reason it was broken ;)

config/initializers/business_time.rb
jessieay: ok cool

Gemfile
```diff
@@ -18,7 +18,7 @@ gem 'puma'
 gem 'samwise', '~> 0.4.0'
 gem 'octokit', '~> 4.0'
 gem 'rack-cors', require: 'rack/cors'
-gem 'active_model_serializers'
+gem 'active_model_serializers', '~> 0.9.5'
```
jessieay: GREAT call -- I spent about 30 mins trying to upgrade to `0.10` and it was such a disaster I gave up

PR #1109, opened 2016-08-25
Update C2 with receipt URL

adelevie:
Closes https://waffle.io/18F/micropurchase/cards/57a2812c9aee7b0e006b2821

big thanks to @jessieay for pairing

sample receipt:

screen shot 2016-08-25 at 2 54 26 pm

### DIFF
--- a/app/jobs/update_c2_proposal_job.rb
+++ b/app/jobs/update_c2_proposal_job.rb
@@ -1,8 +1,10 @@
 class UpdateC2ProposalJob < ActiveJob::Base
   queue_as :default
 
-  def perform(auction_id)
+  def perform(auction_id, attributes_classname)
     auction = Auction.find(auction_id)
-    UpdateC2Proposal.new(auction).perform
+    attributes_class = attributes_classname.constantize
+
+    UpdateC2Proposal.new(auction, attributes_class).perform
   end
 end
--- a/app/models/add_receipt_to_c2_proposal_attributes.rb
+++ b/app/models/add_receipt_to_c2_proposal_attributes.rb
@@ -3,7 +3,7 @@ class AddReceiptToC2ProposalAttributes
     @auction = auction
   end
 
-  def perform
+  def to_h
     {
       gsa18f_procurement: {
         link_to_product: link_to_product
--- a/app/services/add_receipt_to_c2_proposal_attributes.rb
+++ b/app/services/add_receipt_to_c2_proposal_attributes.rb
@@ -1,0 +1,21 @@
+class AddReceiptToC2ProposalAttributes
+  def initialize(auction)
+    @auction = auction
+  end
+
+  def perform
+    {
+      gsa18f_procurement: {
+        link_to_product: link_to_product
+      }
+    }
+  end
+
+  private
+
+  attr_reader :auction
+
+  def link_to_product
+    ReceiptUrl.new(auction).to_s
+  end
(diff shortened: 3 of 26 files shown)


spec/services/update_c2_proposal_spec.rb
```diff
@@ -9,14 +9,17 @@
 
       fake_c2_attributes = { fake_c2: 'fake' }
       attributes_double = double(perform: fake_c2_attributes)
-      allow(UpdateC2Attributes).to receive(:new).with(auction).and_return(attributes_double)
+
+      # e.g. UpdateC2Attributes
```
jessieay: I don't think we need this ;)

app/services/add_receipt_to_c2_proposal_attributes.rb
```diff
@@ -0,0 +1,21 @@
+class AddReceiptToC2ProposalAttributes
```
jessieay: I feel like this class and the other attribute class belong in `app/models` more than `app/services` -- what do you think?

adelevie: I can see that. On one hand, it has a `#perform` method and does a tiny bit of data transformation. On the other hand, it is sort of a business logic object. Your call I guess?

jessieay: I think `perform`does make it conform with other service classes....but I think in the end this is just returning a hash. So really, a nice interface might be a model in `app/models` with a `to_h` method that returns the attributes (the ones currently returned by `perform`)

harrisj: I'm on the fence myself

app/jobs/update_c2_proposal_job.rb
```diff
@@ -1,8 +1,10 @@
 class UpdateC2ProposalJob < ActiveJob::Base
   queue_as :default
 
-  def perform(auction_id)
+  def perform(auction_id, attributes_classname)
     auction = Auction.find(auction_id)
-    UpdateC2Proposal.new(auction).perform
+    attributes_class = attributes_classname.constantize
```
adelevie: Delayed::Job will not serialize classes, so we pass the class name as a string and constantize it.

app/models/add_receipt_to_c2_proposal_attributes.rb
```diff
@@ -0,0 +1,21 @@
+class AddReceiptToC2ProposalAttributes
+  def initialize(auction)
+    @auction = auction
+  end
+
+  def to_h
```
jessieay: ❤️

app/services/confirm_payment.rb
```diff
@@ -0,0 +1,19 @@
...
+    auction.update(c2_status: :payment_confirmed)
+    add_receipt_to_c2_proposal
+  end
+
+  private
+
+  attr_reader :auction
+
+  def add_receipt_to_c2_proposal
+    UpdateC2ProposalJob.perform_later(auction.id,
```
jessieay: this is super minor but this type of alignment feels weird to me. I'd expect either one line or

```
UpdateC2ProposalJob.perform_later(
  auction.id,
  'AddReceiptToC2ProposalAttributes'
)
```

jessieay:
This is so awesome. Merging!!

Issue #947, opened 2016-07-22
Admin Auctions - Needs Attention/Delivery needed


mtorres253:
As an admin I want to quickly see which auctions that need delivery, so when I go to the "Needs Attention" page, I should see them 

**Context**
- Given I'm an authenticated admin
- When I'm on the Needs Attention page 

- [x] **Scenario: Admin sees a "Delivery needed" section on Needs Attention page**
- Then I should see a section labeled "Delivery needed"

- [x] **Scenario: Admin sees auctions that need delivery on the Needs Attention page**
- Given there are auctions that are published
- And those auctions have ended
- And the auctions have winning vendors
- And the auction end date has passed
- And the auction has not yet been marked as delivered by the vendor
- When I look at the Needs Attention page
- Then I should see these auctions listed in the "Delivery needed" section
- And for each auction, I should see a column labeled < Column Name >
- And for each auction, I should see < Data > in that column

| Column Name | Data |
| ----- | ----- |
| Title | Auction Title |
| Delivery Deadline | Days/hours to delivery deadline, or Days/hours past delivery deadline |
| Delivery URL | The Delivery URL or N/A |
| Winning Vendor | Vendor Name |
| Edit | Edit link |

- [x]  **Scenario: Admin sees message that there are no auctions needing delivery**
- Given there are no auctions needing delivery
- When I'm on the "Needs Actions" page
- Then I should see a message "There are no auctions needing delivery" in the "Pending Delivery" section

- [x] **Scenario: Admin clicks on auction title and navigates to auction show page**
- When I click on the title of the auction needing delivery
- Then I should go to that auction's show page

- [x] **Scenario: Admin clicks on edit link and navigates to auction edit page**
- When I click on the edit link of the auction needing delivery
- Then I should go to that auction's edit page

jessieay:
@mtorres253 under what conditions do we put each of these statuses: 

"Delivery Pending", or "Delivery Past Due"

if it is just based on delivery deadline, this seems redundant since we also have delivery deadline in the table as a relative value (eg: 1 day ago)

jessieay:
This added the section: https://github.com/18F/micropurchase/pull/1073/commits

Will keep this open until above questions i resolved

jessieay:
hey @mtorres253 question for you on this (see comments)

mtorres253:
Delivery pending:
- Given there are auctions that are published
- And those auctions have ended
- And the auctions have winning vendors
- And there is not yet a delivery URL
[...]

PR #1111, opened 2016-08-25
Refine "pending delivery" logic on Needs Attention page

jessieay:
* Rename to "Delivery needed"
* Include only published actions
* Include delivery url field
* Closes https://github.com/18F/micropurchase/issues/947

screen shot 2016-08-25 at 4 04 48 pm

### DIFF
--- a/app/controllers/admin/auctions/needs_attention_controller.rb
+++ b/app/controllers/admin/auctions/needs_attention_controller.rb
@@ -1,5 +1,5 @@
 class Admin::Auctions::NeedsAttentionController < Admin::BaseController
   def index
-    @view_model = Admin::NeedsAttentionAuctionsViewModel.new
+    @view_model = Admin::NeedsAttentionIndexViewModel.new
   end
 end
--- a/app/models/auction_query.rb
+++ b/app/models/auction_query.rb
@@ -47,7 +47,7 @@ class AuctionQuery
   end
 
   def delivery_needed
-    relation.published.ended.delivery_needed
+    relation.published.ended.pending_delivery
   end
 
   def c2_payment_needed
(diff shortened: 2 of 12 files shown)


app/controllers/admin/auctions/needs_attention_controller.rb
```diff
@@ -1,5 +1,5 @@
 class Admin::Auctions::NeedsAttentionController < Admin::BaseController
   def index
-    @view_model = Admin::NeedsAttentionAuctionsViewModel.new
+    @view_model = Admin::NeedsAttentionIndexViewModel.new
```
harrisj: Ooops, good catch

app/models/auction_query.rb
```diff
@@ -26,7 +26,7 @@ def active_auction_count
 
   def needs_attention_count
     relation.unpublished.count +
-      pending_delivery.count +
+      delivery_needed.count +
```
harrisj: Yes, this is a better name

app/models/concerns/auction_scopes.rb
```diff
@@ -11,13 +11,13 @@ module AuctionScopes
     scope :default_purchase_card, -> { where(purchase_card: 0) }
     scope :delivered, -> { where.not(delivery_url: [nil, '']) }
     scope :delivery_due_at_expired, -> { where('delivery_due_at < ?', Time.current) }
+    scope :delivery_needed, -> { where(status: statuses['pending_delivery']) }
```
harrisj: I'm kinda tempted to keep this as pending_delivery, since we have delivery_needed being the union of that scope AND published

adelevie:
LGTM, but I'll defer to @harrisj. Also needs rebase.

app/models/concerns/auction_scopes.rb
jessieay: I was trying to keep the naming similar to how we have it on the needs attention page but I see your point - will update!

jessieay: updated at 89744f4

PR #1112, opened 2016-08-25
Allow non-default purchase card auctions to be published #878

### LINKED ISSUE #878: Allow non-default purchase card auctions to be published
This is combination UI improvement and bug fix.

Right now, we have logic in `app/view_models/admin/edit_auction_view_model.rb` that allows an admin to set an auction to `published` *only* if `c2_approved_at` is present. But this field will never be present for auctions for `purchase_card: :other` because those auctions do not have a c2 proposal to be approved.

Pieces of this story:

1. Allow an admin to publish an auction on edit without requiring a value for `c2_approved_at` when the auction is for a `other` purchase card.

2. Add validations at the `ActiveRecord` level that ensure auctions for the `default` purchase card that do not have a  `c2_approved_at` timestamp cannot be published

3. UI improvement to dynamically allow `published` if the `other` purchase card option is selected in the form.

Assuming an auction does *not* have a value for `c2_approved_at`, these are the following states available in the form (for a new auction or when editing):

State 1:

screen shot 2016-07-14 at 3 41 30 pm

screen shot 2016-07-14 at 3 41 48 pm

State 2:

screen shot 2016-07-14 at 3 39 56 pm

screen shot 2016-07-14 at 3 41 05 pm

 This will require a little bit of jquery. 

## General Acceptance Criteria
Does not decrease code climate grade
Tests run on Travis-CI and do not break the build.
New features do not degrade accessibility (use pa11y/HTML Code Sniffer).
New functionality is adequately documented.

saastree:
Closes #878

### DIFF
--- a/app/assets/javascripts/publishedToggle.js
+++ b/app/assets/javascripts/publishedToggle.js
@@ -1,0 +1,14 @@
+$(document).ready(function() {
+  $("#auction_purchase_card").on("change", function(event) {
+    published_select_element = $("#auction_published")
+
+    published_option = "<option value='published'>published</option>"
+    unpublished_option = "<option value='unpublished'>unpublished</option>"
+
+    if(this.value == "default") {
+      published_select_element.empty().append(unpublished_option)
+    } else {
+      published_select_element.empty().append(published_option, unpublished_option)
+    }
+  });
+});
--- a/app/assets/javascripts/publishedToggle.js
+++ b/app/assets/javascripts/publishedToggle.js
@@ -1,11 +1,12 @@
 $(document).ready(function() {
   $("#auction_purchase_card").on("change", function(event) {
+    c2_status = $("#auction_c2_status").val()
     published_select_element = $("#auction_published")
 
     published_option = "<option value='published'>published</option>"
     unpublished_option = "<option value='unpublished'>unpublished</option>"
 
-    if(this.value == "default") {
+    if(this.value == "default" && c2_status != "approved") {
       published_select_element.empty().append(unpublished_option)
     } else {
       published_select_element.empty().append(published_option, unpublished_option)
--- a/app/assets/javascripts/publishedToggle.js
+++ b/app/assets/javascripts/publishedToggle.js
@@ -6,7 +6,7 @@ $(document).ready(function() {
     published_option = "<option value='published'>published</option>"
     unpublished_option = "<option value='unpublished'>unpublished</option>"
 
-    if(this.value == "default" && c2_status != "approved") {
+    if(this.value === "default" && c2_status !== "approved") {
       published_select_element.empty().append(unpublished_option)
     } else {
       published_select_element.empty().append(published_option, unpublished_option)
--- a/app/models/auction.rb
+++ b/app/models/auction.rb
@@ -49,6 +49,14 @@ class Auction < ActiveRecord::Base
     format: { with: Regexp.new("#{C2_REGEX}[0-9]") },
     allow_blank: true
   )
+  validate :publishing_auction, on: :update, if: :published_changed?
+
+  def publishing_auction
+    if published_was == 'unpublished' && purchase_card == 'default' &&
+      c2_status != 'approved'
+      errors.add(:c2_status, " is not approved.")
+    end
+  end
 
   def sorted_skill_names
     skills.order(name: :asc).map(&:name)
--- a/app/models/auction.rb
+++ b/app/models/auction.rb
@@ -52,8 +52,7 @@ class Auction < ActiveRecord::Base
   validate :publishing_auction, on: :update, if: :published_changed?
 
   def publishing_auction
-    if published_was == 'unpublished' && purchase_card == 'default' &&
-      c2_status != 'approved'
+    if published_was == 'unpublished' && purchase_card == 'default' && c2_status != 'approved'
       errors.add(:c2_status, " is not approved.")
     end
   end
--- a/app/view_models/admin/edit_auction_view_model.rb
+++ b/app/view_models/admin/edit_auction_view_model.rb
@@ -38,11 +38,16 @@ class Admin::EditAuctionViewModel < Admin::BaseViewModel
   end
 
   def published_options
-    if closed? || auction.c2_status == 'approved'
+    if publishable?
       Auction.publisheds.keys.to_a
     else
       ['unpublished']
     end
+  end
+
+  def publishable?
+    closed? || auction.purchase_card == 'other' ||
+      auction.c2_status == 'approved'
   end
 
   def customer_options
--- a/app/view_models/admin/edit_auction_view_model.rb
+++ b/app/view_models/admin/edit_auction_view_model.rb
@@ -45,11 +45,6 @@ class Admin::EditAuctionViewModel < Admin::BaseViewModel
     end
   end
 
-  def publishable?
-    closed? || auction.purchase_card == 'other' ||
-      auction.c2_status == 'approved'
-  end
-
   def customer_options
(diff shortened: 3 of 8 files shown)


jessieay:
thanks for opening this! just leave a comment here when you're ready for some feedback and we will be on it 🌈

saastree:
sure! just about to fix the issues found by codeclimate, and we will be ready for feedback!

jessieay:
Few more code climate errors. Let us know if you want clarification on any of those

One high-level piece of feedback: it would be great to get a high level test for this. We have cucumber tests for publishing an auction here: https://github.com/18F/micropurchase/blob/develop/features/admin_publishes_auction.feature

It seems like a scenario with javascript turned on (see an example at https://github.com/18F/micropurchase/blob/develop/features/vendor_views_bid_details.feature) that checks that an admin can create an auction and publish it at the same time if the purchase card is `:other` would be 👍

saastree:
Sure, we will get that test added. We are also ready for any additional feedback.

app/view_models/admin/edit_auction_view_model.rb
```diff
@@ -38,13 +38,18 @@ def published
...
   def published_options
-    if closed? || auction.c2_status == 'approved'
+    if publishable?
       Auction.publisheds.keys.to_a
     else
       ['unpublished']
     end
   end
 
+  def publishable?
```
jessieay: since we are only using this within this class, let's make it a private method

spec/models/auction_spec.rb
```diff
@@ -70,6 +70,16 @@
 
         expect(auction).to be_invalid
       end
+
+      it 'validates presence of c2_status' do
+        auction = create(:auction, published: :unpublished,
```
jessieay: let's just put all of these attributes on one line (I think they will fit?). If they are too long, I usually format with one attr on each line:

```
auction = create(
  :auction,
  published: :unpublished,
  purchase_card: :default
)
```

we also have an `:unpublished` trait in our factories so this can be simplified as:

`auction = create(:auction, :unpublished, purchase_card: :default`

jessieay:
OK, provided a round of feedback. Biggest thing is just writing this higher-level tests I described above. Please let us know if there are any questions!

saastree:
Hi! The test is ready for review. The CI failure seems to be related to the the issue #1122 and nothing to do with our code. The build did pass on our side. Please let us know if there are any questions/concerns. Thanks!

adelevie:
Hi, @saastree. I re-ran the Travis build and it passed this time 🍏

features/step_definitions/admin_auction_form_steps.rb
```diff
@@ -35,6 +35,51 @@
   select("reverse", from: "auction_type")
 end
 
+When(/^I fill auction form with choosing "([^"]*)" purchase card$/) do |purchase_card|
```
jessieay: since we are only using this step for the `other` purchase card, and it does something that only the `other ` purchase card can do (publish on `create`) -- how about hard-coding the step name instead of using interpolation? Something like `"I create and publish an auction for the non-default purchase card"`

features/admin_publishes_auction.feature
```diff
@@ -27,3 +27,20 @@ Feature: Admin publishes auction in the admin panel
...
+    And I visit the auctions admin page
+    When I click on the add auction link
+    And I fill auction form with choosing "other" purchase card
+    Then I should see that the auction type is sealed bid
+    And I click to create an auction
+    Then I should see the auction's title
+
+    When I click on the auction's title
+    Then I should see the start time I set for the auction
+    And I should see the end time I set for the auction
```
jessieay: these expectations are in line with the other test here so I think they are good for now...it would be good to somehow have something around the auction being `published`

features/admin_publishes_auction.feature
```diff
@@ -27,3 +27,20 @@ Feature: Admin publishes auction in the admin panel
...
+    And there is a skill in the system
+    When I sign in
+    And I visit the auctions admin page
+    When I click on the add auction link
+    And I fill auction form with choosing "other" purchase card
+    Then I should see that the auction type is sealed bid
+    And I click to create an auction
+    Then I should see the auction's title
+
+    When I click on the auction's title
```
jessieay: Instead of these, what about a step that visits `root_path` and looks for the auction title? (that is how we know it is published)

app/models/auction.rb
```diff
@@ -49,6 +49,13 @@ class Auction < ActiveRecord::Base
     format: { with: Regexp.new("#{C2_REGEX}[0-9]") },
     allow_blank: true
   )
+  validate :publishing_auction, on: :update, if: :published_changed?
+
+  def publishing_auction
+    if published_was == 'unpublished' && purchase_card == 'default' && c2_status != 'approved'
```
jessieay: interesting -- why are we checking for `published_was`?

spec/models/auction_spec.rb
```diff
@@ -70,6 +70,15 @@
 
         expect(auction).to be_invalid
       end
+
+      it 'validates presence of c2_status' do
```
jessieay: maybe a second test that checks that an approved auction *can* be published?

features/step_definitions/admin_auction_form_steps.rb
```diff
@@ -35,6 +35,51 @@
...
+  @repo = 'https://github.com/18F/calc'
+  fill_in('auction_github_repo', with: @repo)
+
+  @summary = 'The Summary!'
+  fill_in('auction_summary', with: @summary)
+
+  @issue_url = 'https://github.com/18F/calc/issues/255'
+  fill_in('auction_issue_url', with: @issue_url)
+
+  @start_day = DcTimePresenter.convert(3.days.from_now)
```
jessieay: let's just leave `start_day` `end_day` and `delivery_deadline` as the defaults for this step -- we want to fill out all required fields but those have defaults so we can save ~10 lines of spec code by not filling those out. Sound OK to you?

app/models/auction.rb
nikolalsvk: Since we are doing this validation when `:published_changed?`, this is an extra step whether we should validate or not based on the value of `published` field.

For e.g. if you want to unpublish **published** auction, then this validation is just extra work to be done and it will be skipped since `published_was` was "published".

jessieay: got it! thanks for the explanation

nikolalsvk: No problem :)

app/assets/javascripts/publishedToggle.js
```diff
@@ -0,0 +1,15 @@
...
+    c2_status = $("#auction_c2_status").val()
+    published_select_element = $("#auction_published")
+
+    published_option = "<option value='published'>published</option>"
+    unpublished_option = "<option value='unpublished'>unpublished</option>"
+
+    if(this.value === "default" && c2_status !== "approved") {
+      published_select_element.empty().append(unpublished_option)
+    } else {
+      published_select_element.empty().append(published_option, unpublished_option)
```
jessieay: just tested this locally -- can we re-order these so that the default is still `unpublished` ? super small detail, but in general we do not publish auctions on create so it's nice to default to what we'd normally do

jessieay:
Just checking out your updates based on PR feedback. This is looking *great* Left one final piece of feedback based on my testing (sorry I didn't catch that earlier -- some things are so hard to realize until you try something out!)

jessieay:
This looks great! I will merge now. Can you please visit https://micropurchase.18f.gov/auctions/31 and mark this as done when you get a chance? That will enable us to pay you ;)

Issue #1066, opened 2016-08-19
Clarify factories (naming)


jessieay:
frequently find myself visiting the factory file to understand what a factory is.  How can we clean up our naming so that we do not need to do that? Wonder if we are relying too heavily on traits....

harrisj:
I personally like the ability to tweak my test auction via traits rather than knowing that a specific named factory also does this or that thing. But I think we have some redundant traits among other problems

jessieay:
yeah I see that, traits are much more compact

I think maybe rather than having so many traits with several attributes, we should focus on traits with just one or two so we aren't accidentally getting records that have more specific details than we need.  eg

```
    trait :payment_needed do
      delivery_due_at_expired
      accepted
      delivered
      c2_approved
      not_paid
    end

```

I feel like this could be more concise.

harrisj:
So, I think we decided that we should focus on a few small PRs to eliminate some redundant traits like `:running` and then see how we feel about it from there

jessieay:
I also think that removing / refacotring any traits that call other traits would be very helpful as an initial step. even if that just means adding a bunch of explicit attributes to traits

harrisj:
It seems like we have two cases where traits might call a bunch of other traits:

1. The traits are tightly coupled and it is an error condition to have trait A without trait B
2. The traits are usually together, so that a record with trait A will typically also have trait B

I think the second case are ones we should absolutely unroll to be more explicit where called, but I think for the first case, it might make more sense to keep the first case

jessieay:
Some PRs for this (not to complete it)

https://github.com/18F/micropurchase/pull/1092
https://github.com/18F/micropurchase/pull/1093

PR #1113, opened 2016-08-26
Factories: the trilogy, Part III

jessieay:
* Closes https://github.com/18F/micropurchase/issues/1066 ...for now

### DIFF
--- a/spec/view_models/admin/user_auction_view_model_spec.rb
+++ b/spec/view_models/admin/user_auction_view_model_spec.rb
@@ -24,7 +24,7 @@ describe Admin::UserAuctionViewModel do
   describe '#user_bid_count' do
     it 'should return the total number of bids the user has made' do
       user2 = create(:user)
-      auction = create(:auction, :with_bidders, bidder_ids: [user.id, user2.id, user.id, user2.id, user.id])
+      auction = create(:auction, bidders: [user, user2, user, user2, user])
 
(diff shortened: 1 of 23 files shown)


spec/view_models/admin/user_auction_view_model_spec.rb
```diff
@@ -24,7 +24,7 @@
   describe '#user_bid_count' do
     it 'should return the total number of bids the user has made' do
       user2 = create(:user)
-      auction = create(:auction, :with_bidders, bidder_ids: [user.id, user2.id, user.id, user2.id, user.id])
+      auction = create(:auction, bidders: [user, user2, user, user2, user])
```
harrisj: One thing I would love is the ability to pass it something like `nil` or `:new` and just let the factory create the other bidders I don't care about, but I might be overthinking

harrisj:
This looks great. Just needs a rebase

adelevie:
LGTM. Needs rebase.

jessieay:
@adelevie  @harrisj rebased!

spec/view_models/admin/user_auction_view_model_spec.rb
jessieay: the `with_bids` trait creates bids with bidders -- but it sounds like you're talking about something else?

PR #1114, opened 2016-08-26
Public Link to Draft Auctions

harrisj:
Closes https://github.com/18F/micropurchase/issues/1056

Did you know Rails 5 has a has_secure_token helper and that it has been backported to earlier versions of Rails?

This will allow us to potentially provide privileged views of information without standard auth via a URL containing an unguessable token

What page looks like when you view with a token:

screen shot 2016-08-29 at 5 15 34 pm

Share link before auction has been published (includes token so user can see it before publish without being an admin):

screen shot 2016-08-29 at 5 15 09 pm

Share link once auction has been published:

screen shot 2016-08-30 at 9 05 34 am


adelevie:
Looking good so far!

db/schema.rb
```diff
@@ -42,10 +42,12 @@
     t.datetime "rejected_at"
     t.integer  "customer_id"
     t.integer  "c2_status",       default: 0,    null: false
+    t.string   "token"
```
jessieay: `null: false` ? or does this get created after the record in the db?

app/models/auction.rb
```diff
@@ -62,6 +64,12 @@ def lowest_bids
     bids.select { |b| b.amount == lowest_amount }.sort_by(&:created_at)
   end
 
+  def self.find_by_token(token)
+    a = where(token: token).first
+    fail ActiveRecord::RecordNotFound if a.nil?
```
jessieay: I think that is what rails will do for us if we just do `find_by(token:token` above

spec/models/auction_spec.rb
```diff
@@ -114,4 +114,26 @@
       end
     end
   end
+
+  describe '#token' do
+    it 'should automatically be generated when the auction is saved' do
+      u = FactoryGirl.create(:user)
```
jessieay: I don't think we need a user?

spec/models/auction_spec.rb
```diff
@@ -114,4 +114,26 @@
...
+      a = Auction.new(started_at: Time.now, ended_at: Time.now, delivery_due_at: Time.now,
```
jessieay: use regular `(:auction)` factory?

db/migrate/20160826013527_add_token_to_auctions.rb
```diff
@@ -0,0 +1,6 @@
+class AddTokenToAuctions < ActiveRecord::Migration
+  def change
+    add_column :auctions, :token, :string
+    add_index :auctions, :token, unique: true
```
jessieay: yesss indexing

Gemfile
```diff
@@ -30,6 +30,7 @@ gem 'jquery-rails-cdn'
 gem 'jquery-rails'
 gem 'selectize-rails'
 gem 'hana'
+gem 'has_secure_token' # this can be removed in Rails 5
```
jessieay: nice! good idea to use this

spec/models/auction_spec.rb
harrisj: The validations got upset

jessieay: oh yeah that's because you aren't using the factory

harrisj:
So I uploaded some code that supports preview with a token. This what a request to `admin/auctions/7?token=GCbRaQmAc3q2hUt7josax6Be` looks like on my machine

action_controller__exception_caught

Notice that the Edit button is gone and I have my own preview text. Not sure what else we want for the user experience, but there are still a few more steps to be done on this PR:

1. A controller for token generation/revocation/regeneration for a given auction
2. A user interface for activating the previews or deactivating it.

app/models/web_authenticator.rb
```diff
@@ -1,6 +1,14 @@
-class WebAuthenticator < Struct.new(:controller)
+class WebAuthenticator < Struct.new(:controller, :params)
   def current_user
-    @current_user ||= User.where(id: controller.session[:user_id]).first || Guest.new
+    if @current_user.nil?
+      @current_user = if params[:token]
+                        GuestWithToken.new(params[:token])
```
adelevie: so if the token is invalid, this line will cause an AR::RecordNotFound to be raised. I don't think that will be rescued anywhere. Is that ok? Or should we do some kind of redirecting?

jessieay: I think that's fine. that is also what happens when trying to visit a published auction via regular `id`.

This is in line with the philosophy of "exceptions should be exceptional" -- I don't have a good blog post ref on that one, but the idea is that you want exceptions to raise most of the time because that means something has gone wrong and only by raising will you see it in the dev logs (perhaps wrong link in an email or something)

app/controllers/auctions_controller.rb
```diff
@@ -32,6 +32,14 @@ def update
 
   private
 
+  def find_auction
+    if params[:token]
+      Auction.find_by(token: params[:token])
```
adelevie: I'm wondering if we should name this query, and put it in `AuctionQuery` similiar to https://github.com/18F/micropurchase/blob/develop/app/models/auction_query.rb#L12-L14

jessieay: sure -- this seems pretty simple but I am down with doing that for consistency. What do you think we should call it?

adelevie: `token_find`, `draft_find`?

Also, just realized that the `:auction_id` in the URL is irrelevant since we're finding by the token. Is that ok?

adelevie: `Auction.where(token: params[:token]).where(id: params[:id]).first` might tidy things up

jessieay: created `AuctionQuery.unpublished_find` at 884834c

jessieay:
re-review me please @adelevie @harrisj

spec/models/auction_query_spec.rb
```diff
@@ -73,19 +73,28 @@
   end
 
   describe '#public_find' do
-    let(:unpublished_auction) { create(:auction, :unpublished) }
-    let(:published_auction) { create(:auction, :published) }
-
     it 'should not return an unpublished auction' do
+      unpublished_auction= create(:auction, :unpublished)
```
adelevie: :rotating_light: minor spacing issue :rotating_light:

spec/models/auction_query_spec.rb
```diff
@@ -73,19 +73,28 @@
...
     it 'should return a published auction' do
+      published_auction = create(:auction, :published)
+
       expect(query.public_find(published_auction.id)).to eq(published_auction)
     end
   end
 
+  describe '#unpublished_find' do
+    it 'returns an unpublished auction' do
+      unpublished_auction= create(:auction, :unpublished)
```
adelevie: moar spacing

spec/models/auction_query_spec.rb
jessieay: ughhhh hate it when I do stuff like this. fixing now

spec/models/auction_query_spec.rb
jessieay: fixed at dbfb9b3

adelevie:
Great work @harrisj and @jessieay. LGTM, merging when green.

Issue #1117, opened 2016-08-26
Only run clock on first instance


jessieay:
because we have two instances, we are running all clock tasks twice. we should only run them once
 
https://18f.slack.com/archives/ruby/p1472169849000856

PR #1118, opened 2016-08-26
Only run clock.rb on first cf instance

adelevie:
Attempts to fix https://github.com/18F/micropurchase/issues/1117

### DIFF
--- a/config/clock.rb
+++ b/config/clock.rb
@@ -1,6 +1,7 @@
 require "clockwork"
 require_relative "boot"
 require_relative "environment"
+require_relative "../lib/server_env"
 
 module Clockwork
   # disabled until https://github.com/18F/micropurchase/issues/951 is fixed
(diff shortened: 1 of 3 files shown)


lib/server_env.rb
```diff
@@ -4,4 +4,8 @@ module ServerEnv
   def self.instance_index
     CloudFoundry.instance_index || 0
   end
+
+  def self.first_instance?
+    self.instance_index == 0
```
jessieay: I would probably call this just as `instance_index` rather than using `self` but that's just me -- what do you think?

config/clock.rb
```diff
@@ -1,6 +1,7 @@
 require "clockwork"
 require_relative "boot"
 require_relative "environment"
+require File.expand_path('../../lib/server_env', __FILE__)
```
jessieay: `require_relative` to be consistent with other requires in here?

lib/server_env.rb
jessieay: looks like we are doing this check elsewhere (`unless ServerEnv.instance_index == 0`) -- update that call to use `first_instance?`

config/clock.rb
```diff
@@ -9,21 +10,22 @@ module Clockwork
   # TockImporter.new.delay.perform
   # end
   #
+  if ServerEnv.first_instance? || Rails.env.development?
```
jessieay: nice! so much better

jessieay:
Just a few small comments from me!

jessieay:
looks like code climate agrees with me comment about `self`

jessieay:
this looks great! thoughts on manually deploying to staging to see if it works before merging? I am ok with doing a regular merge (and staging deploy), too. In that case it will just more annoying to fix if it is somehow broken because we will need to do a revert PR

adelevie:
I'm, uh, feeling good about this one. So .... I'll try a manual deploy.

dogweather:
I agree with @cmc333333 's comment, that the code looks good. 

I think this general concurrent strategy is kind of sketchy; but that's a different issue.

dogweather:
I dunno, or maybe it's the best way, because there's no designated clock instance. And so, it may be a little more reliable. As long as there's at least one server, there'll be a running clock service.

jessieay:
Mergin'

Issue #1106, opened 2016-08-24
C2 integration failing?


mtorres253:
To duplicate:

Go to https://micropurchase.18f.gov/admin/auctions/33

Expected behavior
I should see the following messaging

> Pending C2 approval
This auction has been sent to C2 for approval. You can check the status of this approval here.

Actual behavior:
I see the following messaging

> Pending C2 approval
This auction has been sent to C2 for approval. 

Notes: 
The C2 proposal URL exists but I still see the Pending C2 approval message (without the "here" link)

https://cap.18f.gov/proposals/5351

Same thing is happening on https://micropurchase.18f.gov/admin/auctions/29

jessieay:
Pr to fix this: https://github.com/18F/micropurchase/pull/1121

was not a bug in C2 integration, was an error on our end. We were accidentally flipping the statuses of "sent to c2 but no URL yet" and "sent to c2 and we have a URL but not approved yet"

PR #1121, opened 2016-08-26
Use correct `c2_status` for auctions

jessieay:
* We were accidentally using `pending` when we meant to use `sent`
* Renamed `pending` to make it clearer that this means `pending approval
  from C2` whereas `sent` means 'sent request to c2 but no URL yet`
* Closes https://github.com/18F/micropurchase/issues/1106


harrisj:
LGTM. Merging

PR #1124, opened 2016-08-26
Move Timecop out of dev group

jessieay:
* For dev:prime, again


harrisj:
I am wondering if we want to say `require: false` for it and explicitly require it in the spec_helper and factory_girl auctions definition since I don't think we want to load it on production otherwise?

jessieay:
sure! good idea

jessieay:
Ok so I just tried this out -- problem with adding `require: false` is that then we need to explicitly require these gems in the test env which is kind of a pain.

Then I tried to include them without `require: false` in the test env in addition to listing them outside, I get this error from bundler:

```
Your Gemfile lists the gem timecop (>= 0) more than once.
You should probably keep only one of them.
```

what do you think we should do?

harrisj:
Yeah, we'd have to require explicitly within the rails_helper, the env.rb for cucumber and the dev prime as well. But let's wait on that I guess

PR #1127, opened 2016-08-26
Remove API redirects

jessieay:
* These were temporary
* We've updated other aspects of the API since then, so hopefully anyone
  using the API has updated their usage anyway


harrisj:
Thank you!

PR #1131, opened 2016-08-29
Fix translations

jessieay:
* Typo on one
* Add hyperlink on other

### DIFF
--- a/config/environments/development.rb
+++ b/config/environments/development.rb
@@ -2,6 +2,7 @@ Rails.application.configure do
   config.action_controller.perform_caching = false
   config.action_mailer.delivery_method = :letter_opener_web
   config.action_mailer.raise_delivery_errors = false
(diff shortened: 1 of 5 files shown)


config/environments/development.rb
```diff
@@ -2,6 +2,7 @@
   config.action_controller.perform_caching = false
   config.action_mailer.delivery_method = :letter_opener_web
   config.action_mailer.raise_delivery_errors = false
+  config.action_view.raise_on_missing_translations = true
```
jessieay: now we will see exceptions on dev when a translation is missing rather than the key (rails just uses key string if it can't find a translation...nice but hard to find missing keys)

see https://robots.thoughtbot.com/foolproof-i18n-setup-in-rails

adelevie:
🌊

jessieay:
🍏

PR #1137, opened 2016-08-30
:cry: DUNS doesn't support Chrome

tmcw:
Ran into the same issue on DUNS many times - code `CEPDEV101` - and called their number. Chrome is not supported, and I confirmed that Safari works.

### DIFF
--- a/app/views/application/faq.html.erb
+++ b/app/views/application/faq.html.erb
@@ -26,7 +26,9 @@
       Taxpayer Identification Number (TIN). If you do not have an active DUNS
       number, you may obtain one online at the <a
         href="http://fedgov.dnb.com/webform">Dun &amp; Bradstreet site</a>. This
-      process takes one-to-two business days. If you do not have a TIN, you may
+      process takes one-to-two business days, and the Dun &amp; Bradstreet website
+      is not compatible with Google Chrome: Safari or Internet Explorer are recommended.
+      If you do not have a TIN, you may
       complete an online application on the <a
         href="https://www.taxid-gov.us/">Tax ID site</a>.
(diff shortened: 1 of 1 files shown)


app/views/application/faq.html.erb
```diff
@@ -26,7 +26,9 @@
       Taxpayer Identification Number (TIN). If you do not have an active DUNS
       number, you may obtain one online at the <a
         href="http://fedgov.dnb.com/webform">Dun &amp; Bradstreet site</a>. This
-      process takes one-to-two business days. If you do not have a TIN, you may
+      process takes one-to-two business days, and the Dun &amp; Bradstreet website
+      is not compatible with Google Chrome: Safari or Internet Explorer are recommended.
```
mtorres253: @tmcw The site works in Chrome on a Mac. Are you using a PC? Maybe we should make that clear

tmcw: I'm using a Mac. Chrome `53.0.2785.80 beta (64-bit)`, uBlock + HTTPS everywhere on. Called DUNS and they immediately said to avoid using Chrome.

jessieay:
Thanks for opening this! Sorry you had to deal with that issue / had to call the people at DUNS 😢 

cc @andrewmaier can you check out these changes and merge if you're 👍  ?

app/views/application/faq.html.erb
mtorres253: Ah OK. I'm using  52.0.2743.116 (64-bit) on a Mac. Thanks!

andrewmaier:
@jessieay Looks good to me!

Issue #1122, opened 2016-08-26
Fix intermittently failing spec


jessieay:
I thought this would fix it, but saw failure again after https://github.com/18F/micropurchase/pull/1119

adelevie:
To reproduce:

```sh
$ bundle exec cucumber features/admin_creates_auction.feature:22
```

adelevie:
Fully expanded cucumber steps:

```ruby
Given(/^I am an administrator$/) do
  @user = FactoryGirl.create(:admin_user)
  @github_id = @user.github_id
  mock_sign_in(@user.github_id, @user.name)
end

Given(/^there is a client account to bill to$/) do
  @billable = FactoryGirl.create(:client_account)
end

Given(/^there is a skill in the system$/) do
  @skill = FactoryGirl.create(:skill)
end

When(/^I sign in$/) do
  step "I visit the home page"
  click_on "Sign in"
end

When(/^I visit the auctions admin page$/) do
  visit admin_auctions_path
end

When(/^I click on the add auction link$/) do
  add_link = I18n.t('links_and_buttons.auctions.add')
  step("I click on the \"#{add_link}\" link")
end

Then(/^I should see an estimated delivery deadline of 12 business days from now$/) do
  within('.estimated-delivery-date') do
    default_delivery_date = DefaultDateTime.new(12.business_days.from_now).convert
    expect(page).to have_content("Estimated delivery date #{DcTimePresenter.convert_and_format(default_delivery_date)}")
  end
end

When(/^I change the auction end date$/) do
  datetime = DefaultDateTime.new(3.business_days.from_now).convert
  @end_day = DcTimePresenter.convert(datetime)
  fill_in "auction_ended_at", with: @end_day.strftime('%Y-%m-%d')
end

Then(/^I should see the updated delivery deadline$/) do
  within('.estimated-delivery-date') do
    delivery_date = DefaultDateTime.new(5.business_days.after(@end_day)).convert
    formatted_delivery_date = DcTimePresenter.convert_and_format(delivery_date)
    expect(page).to have_content(
      "Estimated delivery date #{formatted_delivery_date}"
    )
  end
end
```

PR #1138, opened 2016-08-30
Fix business days testing error

adelevie:
closes https://github.com/18F/micropurchase/issues/1122

### DIFF
--- a/features/admin_creates_auction.feature
+++ b/features/admin_creates_auction.feature
@@ -18,6 +18,7 @@ Feature: Admin creates auctions in the admins panel
     Then I should see the start time I set for the auction
     And I should see the end time I set for the auction
 
+  @freeze_time
   @javascript
   Scenario: Adding an auction with a non-default delivery timeline
     Given I am an administrator
--- a/features/admin_creates_auction.feature
+++ b/features/admin_creates_auction.feature
@@ -18,7 +18,6 @@ Feature: Admin creates auctions in the admins panel
     Then I should see the start time I set for the auction
     And I should see the end time I set for the auction
 
-  @freeze_time
   @javascript
   Scenario: Adding an auction with a non-default delivery timeline
     Given I am an administrator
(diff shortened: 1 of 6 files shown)


features/support/env.rb
```diff
@@ -23,6 +23,14 @@
   Delayed::Worker.delay_jobs = false
 end
 
+Before('@freeze_time') do
+  Timecop.freeze(5.weeks.ago)
```
adelevie: I just realized how foolish this is. It will cause the errors to return in `5.weeks`. I will update.

features/admin_creates_auction.feature
```diff
@@ -18,6 +18,7 @@ Feature: Admin creates auctions in the admins panel
     Then I should see the start time I set for the auction
     And I should see the end time I set for the auction
 
+  @freeze_time
```
jessieay: it's cool that this fixes it -- but do we know *why*? If not, I am worried that we are missing an underlying bug or something that this exposes...

harrisj: Me too. I am looking into it so I can understand exactly.

Also, this needs a Dr. Horrible reference

adelevie: One trick might be to litter the code path with some `puts` statements, and change the `@freeze_time` date by a day and find the precise day that begins to cause the error. Then compare the `puts` output.

features/support/env.rb
```diff
@@ -23,6 +23,14 @@
   Delayed::Worker.delay_jobs = false
 end
 
+Before('@freeze_time') do
```
jessieay: I think these are now unused -- remove?

jessieay:
from commit message:
> think the problem is my fault compounding another error. Basically,
the failing test was checking if the business_time deadline estimation
called by javascript from the frontend was correct. So, the following
things seem to have happened:

>  * A 1pm time was parsed by controller as 1 pm UTC, which is 5 pm ET. I
  figured I might as well fix that.
>  * The end day was set to a default date time 3 business days from the
  current time the test was run (DefaultTime#convert resets the time of
  the final product to 1pm, but this is done after the addition of 3
  business days. So, if the test is run in UTC timezone in the early
  hours (late in EST timezone), it might be a day or more off.

jessieay:
Just one comment about some unused spec code -- otherwise LGTM!

harrisj:
Thanks! Removed!

jessieay:
Looks good! merging

Issue #1014, opened 2016-08-08
Vendor messaging - Auction show page -Work rejected


mtorres253:
As a vendor who has won an auction, so that I know that the code I delivered was rejected, I should see clear messaging when I look at the show page for that auction

- [x] **Scenario: Winning vendor sees message that auction deliverable was rejected"
- Given I am a registered vendor
- And the code I have delivered has been rejected (the admin has marked the auction as rejected)
- When I go to the show page for that auction
- Then I should see the following messaging in the status box

> Rejected
> Your code submission was rejected on [auction.rejectionDate].

## Notes
- The messaging in the prototype is not correct so I didn't paste a screenshot. Just put this in the blue status box and show to logged in vendors who have had their code rejected

PR #1140, opened 2016-08-30
Add "auction rejected" messaging for winning vendor on auctions#show

jessieay:
* Closes https://github.com/18F/micropurchase/issues/1014


adelevie:
LGTM merging when green

PR #1144, opened 2016-08-30
Adding basic robots.txt

adelevie:
Only blocks URLs with `token=` in the string.

### DIFF
--- a/public/robots.txt
+++ b/public/robots.txt
@@ -1,5 +1,5 @@
 # See http://www.robotstxt.org/robotstxt.html for documentation on how to use the robots.txt file
 #
 # To ban all spiders from the entire site uncomment the next two lines:
-# User-agent: *
-# Disallow: /
+User-agent: *
+Disallow: /*?token=*


harrisj:
According to the robots.txt spec

> Note also that globbing and regular expression are not supported in either the User-agent or Disallow lines. The '*' in the User-agent field is a special value meaning "any robot". Specifically, you cannot have lines like "User-agent: *bot*", "Disallow: /tmp/*" or "Disallow: *.gif".

So I think we'll have to use robots meta declarations on the token page. Note that we should have /admin/ in the robots.txt regardless, right?

jessieay:
I don't think we need to worry about ignoring anything in `admin` since that all requires auth and for the user to be an admin.

public/robots.txt
```diff
@@ -1,5 +1,5 @@
 # See http://www.robotstxt.org/robotstxt.html for documentation on how to use the robots.txt file
 #
 # To ban all spiders from the entire site uncomment the next two lines:
-# User-agent: *
-# Disallow: /
+User-agent: *
```
jessieay: do we want this un-commented? I don't really know what it means

harrisj: Basically, you can target certain directives to specific crawlers (which identify by their User-agent) or just define rules that apply to all crawlers.

jessieay: oh I see so the below directive is for all user-agents.

jessieay:
Does this look good to you, @harrisj ? Seems OK to me but I'd need to do a bit more research before feeling confident in merging myself. If someone else knows more and wants to merge, that is 💯

harrisj:
I don't think it works, according to the robots.txt spec there is no support for regexes or globbing

mtorres253:
Let's explore the meta tags solution as it seems more likely to work (given that bots aren't intent on exposing information we don't want them to expose)

jessieay:
another option is to redirect any auction that us published to the regular url (no token)

adelevie:
closing in favor of https://github.com/18F/micropurchase/pull/1181

PR #1148, opened 2016-08-31
Added a model for doing deadline calculations

harrisj:
I wasn't happy to see we were doing the delivery_due_at calculations separately in both the AuctionParser and the business_controller that is called by the frontend javascript, especially since the AuctionParser one seemed to default to the end of the workday. So I created a single class that is called by both of those for doing the date time calculations. This lets us unit test it.

While I was testing it, I discovered that the prior version of business_time we had pinned to was actually wrong with its timezone handling. So I upgraded to the latest version and all tests seem to pass now.


app/presenters/default_deadline_date_time.rb
```diff
@@ -0,0 +1,13 @@
+# used for computing the default deadline with an offset
```
jessieay: we aren't doing class-level comments throughout this codebase. you think we need one here?

app/presenters/default_deadline_date_time.rb
```diff
@@ -0,0 +1,13 @@
...
+class DefaultDeadlineDateTime
+  attr_reader :dc_time
+
+  def self.parse(start_time_str, day_offset)
```
jessieay: let's avoid naming vars after data types and just call this `start_time` ?

app/presenters/default_deadline_date_time.rb
```diff
@@ -0,0 +1,13 @@
...
+    time = DcTimePresenter.parse(start_time_str)
+    new(time, day_offset)
+  end
+
+  def initialize(start_time, day_offset)
+    @dc_time = day_offset.business_days.after(DcTimePresenter.new(start_time).convert)
```
jessieay: I wouldn't expect `initialize` to do calculations like this -- maybe have a `parse` method and call that in `self.parse` ? 

```
def self.parse(start_time_str, day_offset)
  new(start_time, day_offset).parse
end
```

app/presenters/default_deadline_date_time.rb
```diff
@@ -0,0 +1,13 @@
+# used for computing the default deadline with an offset
+class DefaultDeadlineDateTime
+  attr_reader :dc_time
+
+  def self.parse(start_time_str, day_offset)
+    time = DcTimePresenter.parse(start_time_str)
```
jessieay: why do we need to parse start time here and below?

spec/presenters/dc_time_presenter_spec.rb
```diff
@@ -19,6 +19,16 @@
         expect(converter.convert).to eq(nil)
       end
     end
+
+    context 'when it is already in the DC timezone' do
+      let(:time) { Time.parse('2016-09-01 13:00 -04:00') }
```
jessieay: I think we are moving away from let :)

app/presenters/dc_time_presenter.rb
```diff
@@ -13,6 +13,10 @@ def self.convert(time)
     new(time).convert
   end
 
+  def self.parse(time_str)
```
jessieay: `time_str` -> `time`

app/models/auction_parser.rb
```diff
@@ -45,8 +45,8 @@ def strong_params
 
   def delivery_due_at
     if due_in_days.present?
-      real_days = due_in_days.to_i.business_days
-      end_of_workday(real_days.after(ended_at))
+      real_days = due_in_days.to_i
+      DefaultDeadlineDateTime.new(ended_at, real_days).dc_time
```
jessieay: idea: let's used named args here rather than depending on ordering. So `DefaultDeadlineDateTime.new(start_time: ended_at, day_offset: real_days).dc_time` --I think that will make this easier to understand

app/presenters/default_deadline_date_time.rb
```diff
@@ -0,0 +1,13 @@
+# used for computing the default deadline with an offset
+class DefaultDeadlineDateTime
+  attr_reader :dc_time
```
jessieay: I'd make this a method rather than an `attr_reader` -- think it's clearer that way. What do you think?

spec/presenters/default_deadline_date_time_spec.rb
```diff
@@ -0,0 +1,26 @@
+require 'rails_helper'
+
+describe DefaultDeadlineDateTime do
+  before(:all) do
+    BusinessTime::Config.holidays << Date.parse('2016-09-05')
```
jessieay: let's do this in each one! I know it's more lines of code but when the setup is outside of the test it is harder to know what the dependencies are. Also, I am guessing we don't need this config for each test.

spec/presenters/default_deadline_date_time_spec.rb
```diff
@@ -0,0 +1,26 @@
...
+
+    it 'should handle holidays correctly' do
+      t = DefaultDateTime.new(Time.parse('2016-08-29 15:00 -0400')).convert
+      expect(DefaultDeadlineDateTime.new(t, 5).dc_time).to be_within(1).of(Time.parse('2016-09-06 13:00 -0400'))
+    end
+  end
+
+  describe 'parse' do
+    it 'should parse the time in the DC timezone and do the offset' do
+      expect(DefaultDeadlineDateTime.parse('2016-08-29 15:00', 3).dc_time).to be_within(1).of(Time.parse('2016-09-01 15:00 -0400'))
```
jessieay: instead of doing `be_within` what about freezing time and looking or exact matches?

spec/controllers/api/v0/business_days_controller_spec.rb
```diff
@@ -0,0 +1,17 @@
+require 'rails_helper'
+
+RSpec.describe Api::V0::BusinessDaysController do
+  describe 'show' do
+    it 'should return a date N business days from the date and time in the DC time zone' do
+      get :show, {date: '2016-08-30', time: '13:00', day_count: '1'}
```
jessieay: i think our ruby style guide, while it does not evaluate specs, calls for spaces within curly braces

jessieay:
So much better! Left a handful of smaller comments, all on the code style side. ;) Woot woot so glad you got to the bottom of this and cleaned up some of this parsing

app/presenters/default_deadline_date_time.rb
harrisj: I wonder if we should just do class methods and not even bother instantiating?

app/presenters/default_deadline_date_time.rb
harrisj: I wanted to make sure that the right type of parse (ie, the one in DcTimePresenter is called vs. Time.parse), although I could just call that.

spec/presenters/dc_time_presenter_spec.rb
harrisj: Not me! Let me do my thing

app/presenters/default_deadline_date_time.rb
jessieay: the model you have here is a pattern I've  seen done in the past (have a class method instantiate) but in this codebase we aren't doing much of it so I could go either way...

spec/presenters/dc_time_presenter_spec.rb
jessieay: haha ok but perhaps warrants a discussion since I've been spending cycles removing `let`

app/presenters/default_deadline_date_time.rb
```diff
@@ -0,0 +1,7 @@
...
+
+  def initialize(start_time:, day_offset:)
+    @dc_time = day_offset.business_days.after(DcTimePresenter.new(start_time).convert)
```
jessieay: I think we should have a method call `dc_time` that returns this. my rule of thumb is that `initialize` should only  take in args, no data transforms

spec/presenters/dc_time_presenter_spec.rb
```diff
@@ -19,6 +19,16 @@
...
+
+      it 'should stay in the DC timezone' do
+        c = converter.convert
```
jessieay: can we think of a better var name than `c`?

spec/presenters/default_deadline_date_time_spec.rb
```diff
@@ -0,0 +1,17 @@
+require 'rails_helper'
+
+describe DefaultDeadlineDateTime do
+  describe '#initialize' do
+    it 'should be N business days later' do
+      t = DcTimePresenter.parse('2016-08-29 15:00')
```
jessieay: can we use `time` rather than `t`? I try to not use one-letter vars when possible because they can get confusing

spec/presenters/default_deadline_date_time_spec.rb
```diff
@@ -0,0 +1,17 @@
...
+      expect(DcTimePresenter.convert(t)).to be_within(1).of(t)
```
jessieay: thoughts on my previous comment about freezing  time rather than using `within(1)` ?

harrisj: The problem isn't freezing time but float comparison. I could either just cast both sides `to_s` or make the interval a little tighter, but I'm not sure it matters?

jessieay: oh interesting -- why not remove seconds altogether? (have seconds count always be zero coming from `DefaultDeadlineDateDate`)

app/presenters/default_deadline_date_time.rb
harrisj: okay, will do

spec/presenters/default_deadline_date_time_spec.rb
harrisj: Well, the issue is that Ruby time objects are basically floats internally. I do want to know if we want to force the deadline to be a specific time (like 5 pm) like we do for default start times

jessieay: I don't think we want to set a specific time, but since there is an hour and minute selector but not a second selector in the form, shouldn't see assume this is always at 0 seconds?

jessieay:
🌈 🌈

Issue #1084, opened 2016-08-22
Admin can view different auction states


jessieay:
assumptions:

- This is just for dev and staging, not prod
- this should be read-only

Resources:
* https://robots.thoughtbot.com/how-to-masquerade-as-another-user-to-see-how-they-use-yo
* https://coderwall.com/p/e7iyxw/devise-fast-user-switching-using-a-custom-authentication-strategy
* http://www.justinweiss.com/articles/how-rails-sessions-work/
* https://github.com/flyerhzm/switch_user

see https://github.com/18F/micropurchase/issues/995 for more details

PR #1150, opened 2016-08-31
Run staging:prime on staging deploy

jessieay:
* This is to prep for https://github.com/18F/micropurchase/issues/1084
* In a world where we can view auctions as different people, we also
  want to be able to see auctions in many different states
* We can do this locally by running `bin/setup`
* On staging, we run this on each deploy via `staging:prime`
* Philosophical question: do we want to do this? Upside is we have more
  useful seed data on staging. Downside is that we are not running
  migrations on existing data like we do on prod, so there is less
  parity.

### DIFF
--- a/lib/tasks/dev.rake
+++ b/lib/tasks/dev.rake
@@ -1,9 +1,9 @@
-require "factory_girl"
-  require Rails.root.join('spec', 'support', 'helpers', 'factory_girl.rb')
+require 'factory_girl'
+require Rails.root.join('spec', 'support', 'helpers', 'factory_girl.rb')
 
 namespace :dev do
-  desc "Sample data for local development environment"
-  task prime: "db:setup" do
+  desc 'Sample data for local development environment'
+  task prime: 'db:setup' do
     include FactoryGirl::Syntax::Methods
 
     skills = [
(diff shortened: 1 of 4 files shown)


harrisj:
Does this clear out prior auctions?

jessieay:
@harrisj yes, that is the conundrum. I would hate to have a migration fail on prod because we didn't run it on existing data on staging. 

One alternative: instead of clearing out and updating, we could hard-code the titles of a few and do a `find_or create_by` on title. Then, we can update the timestamps of those so on staging we always have a few open, expiring, future, and closed auctions. What do you think of that plan?

harrisj:
I think that could work, but what if you are trying to debug something with an auction and it gets reset? Am I overthinking it?

jessieay:
I think we could update this to make the data not be deleted, just updated so we have data in all states on staging (eg : future, pending payment, pending approval etc) -- let me try that and see how it goes

jessieay:
ok so I just updated this to run the migrations on staging, then drop the DB and seed it. So if migrating existing data fails, we will see that on a staging deploy. Yay!

This does *not* address the concern of needing to debug sonething on staging and then deploy to try and fix. So far, I've never had to do something like that on this project. But I can totally see the concern there. 

Conflicted re: my desire to have helpful data on staging vs being able to know certain records will stick around. Right now, staging is just a graveyard of old auctions that are not in states that are helpful to us in terms of demo-ing new features. What do you think?

jessieay:
ping ping @harrisj

harrisj:
Well let's do it I guess and see if it burns us later

PR #1155, opened 2016-09-01
Added customer and client_account models to rake dev:prime

adelevie:
This creates enough data to create new auctions from scratch in a local development environment.


harrisj:
LGTM

PR #1157, opened 2016-09-02
work_in_progress status for admin auction show

### LINKED ISSUE #1132: Admin messaging - Auction Show Page - Work in progress
As an admin user, I want to quickly see whether work has been started on an auction, so the system should show me an informative message when I am on the show page for that auction

- [ ] **Scenario: Admin sees in progress messaging on auction show page**
- Given I am an authenticated admin user
- And there is a published auction
- And work has been started on that auction (vendor has clicked "Start Work")
- When I go to the admin auction show page (https://micropurchase.18f.gov/admin/auctions/[id])
- Then I should see the following messaging in the auction status box

> Work in progress
> [auction.winner] initiated a pull request on [auction.startWorkDate], and should complete their work no later than [auction.deliveryDeadline].

screen shot 2016-08-29 at 10 34 08 am

- And "[auction winner] initiated a pull request" should link to the delivery URL that was submitted

harrisj:
Fixes #1132 

I also made a few tweaks to add a dedicated `work_in_progress?` to `StatusPresenter` rather than recreating it in several places.

### DIFF
--- a/app/models/admin_auction_status_presenter_factory.rb
+++ b/app/models/admin_auction_status_presenter_factory.rb
@@ -10,6 +10,8 @@ class AdminAuctionStatusPresenterFactory
       Object.const_get("C2StatusPresenter::#{c2_status}").new(auction: auction)
     elsif future? && auction.published?
       AdminAuctionStatusPresenter::FuturePublished.new(auction: auction)
+    elsif work_in_progress?
+      AdminAuctionStatusPresenter::WorkInProgress.new(auction: auction)
     elsif !auction.pending_delivery?
       Object.const_get("AdminAuctionStatusPresenter::#{status}").new(auction: auction)
     else
@@ -21,6 +23,10 @@ class AdminAuctionStatusPresenterFactory
 
   def future?
     AuctionStatus.new(auction).future?
+  end
+
+  def work_in_progress?
+    AuctionStatus.new(auction).work_in_progress?
   end
 
   def status
--- a/app/models/auction_status.rb
+++ b/app/models/auction_status.rb
@@ -19,6 +19,10 @@ class AuctionStatus
     available? && auction.ended_at < (Time.current + 12.hours)
   end
 
+  def work_in_progress?
+    auction.delivery_url.present? && !(auction.pending_acceptance? || auction.accepted? || auction.rejected?)
+  end
+
   private
 
   attr_reader :auction
--- a/features/step_definitions/auction_create_steps.rb
+++ b/features/step_definitions/auction_create_steps.rb
@@ -40,6 +40,10 @@ Given(/^I am going to lose an auction$/) do
   end
 end
 
+Given(/^there is an auction with work in progress$/) do
+  @auction = FactoryGirl.create(:auction, :closed, :with_bids, :delivered)
+end
+
 When(/^the auction ends$/) do
   Timecop.return
   Timecop.travel(@auction.ended_at + 5.minutes)
--- a/features/step_definitions/auction_create_steps.rb
+++ b/features/step_definitions/auction_create_steps.rb
@@ -41,7 +41,7 @@ Given(/^I am going to lose an auction$/) do
 end
 
 Given(/^there is an auction with work in progress$/) do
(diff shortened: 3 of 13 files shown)


app/models/auction_status.rb
```diff
@@ -19,6 +19,10 @@ def expiring?
     available? && auction.ended_at < (Time.current + 12.hours)
   end
 
+  def work_in_progress?
```
jessieay: love this!!

jessieay: I think this is great for now, but we could also consider adding a `work_in_progress` enum so that we don't have to do this crazy stuff

features/step_definitions/auction_create_steps.rb
```diff
@@ -40,6 +40,10 @@
   end
 end
 
+Given(/^there is an auction with work in progress$/) do
+  @auction = FactoryGirl.create(:auction, :closed, :with_bids, :delivered)
```
jessieay: wondering if we should have a `work_in_progress` trait since `delivered` feels kind of misleading here

app/models/admin_auction_status_presenter_factory.rb
```diff
@@ -8,6 +8,8 @@ def initialize(auction:)
   def create
     if auction.payment_confirmed? || auction.c2_paid?
       Object.const_get("C2StatusPresenter::#{c2_status}").new(auction: auction)
+    elsif work_in_progress?
```
jessieay: the other benefit to a `work_in_progress` addition to the `status` enum is that we wouldn't need this conditional, we could just use ` Object.const_get("AdminAuctionStatusPresenter::#{status}").new(auction: auction)` below

jessieay:
This looks 🌈 🌈 

Included some questions about whether we might want to add a `work_in_progress` enum, but totally up to you. I think this is already very clear and a big improvement on what we had before.

app/models/auction_status.rb
harrisj: I was going to suggest that potentially, but we are kinda getting to a point where we are recapitulating a state machine too ;)

One thing that is annoying is that we have currently 3 different places where `status` is described for an auction:

1. the `status` enum column
2. the AuctionStatus class (normally just derived from other fields)
3. The `c2_status` field

I wasn't quite sure where to put this at first and thought for a while before I felt like it should go with the `future?` and other methods that are derived from the values of other fields in the auction. But we could consider some renaming (like AuctionStatus -> AuctionAvailability) that might mitigate some confusion

features/step_definitions/auction_create_steps.rb
harrisj: Well, delivered in this case is basically `has_delivery_url`, so I don't know if should rename to that?

jessieay: yeah maybe that would be clearer?

app/models/auction_status.rb
jessieay: I agree, the two fields are confusing and there feels like some overlap. I also like the idea of renaming `AuctionStatus` 

I do think that one challenge we have is that each status on the chain is not captured in `status` -- while we would be more state-machine-y, we are not forcing the statuses to go in a particular order so at least that is good.

jessieay:
@harrisj let me know what you think of my responses to your responses. If you want to make further changes, go for it. If not, I am more than happy to merge as-is!

harrisj:
Hmmm, I think maybe we should make `work_in_progress` enum a different PR, but will change name of trait

jessieay:
@harrisj alas, merge conflicts on this one

spec/controllers/api/v0/business_days_controller_spec.rb
```diff
@@ -9,6 +9,7 @@
     end
 
     it 'should handle holidays correctly' do
+      BusinessTime::Config.holidays << Date.parse('2016-09-05')
```
jessieay: can we put this in a separate PR? seems unrelated and I'd like to have it as its own thing

jessieay: yeah I really think we should do it separately so we can have a clean commit history to understand why this change was made. I will do it and we can merge that first so other things arent broken

PR #1158, opened 2016-09-02
fixing local docker setup

adelevie:
previously `foreman` was being (incorrectly) run as a rake command. this PR adds an `&&` to pull it out of rake and into its own command.


adelevie:
confirmed that this works, ready to review+merge

toolness:
Looks good to me!

PR #1159, opened 2016-09-02
Admin can unpublish a future auction

### LINKED ISSUE #1133: Admin messaging - Published Auction - Coming Soon
As an admin user, I want to be able to easily unpublish auctions in case I need to remove any of them before they go live. So, there should be a way for me to do this on the auction show page

- [ ] **Scenario: Admin sees unpublish message on auction show page"
- And there is a published auction that has not yet started (coming soon)
- When I go to the admin auction show page https://micropurchase.18f.gov/admin/auctions/[id]
- Then I should see the following messaging

> Coming soon
> This auction is visible to the public but is not currently accepting bids. It will open on [auction.startDate]. If you need to take it down for whatever reason, press the “un-publish” button, below.

- And I should see a button labeled "Unpublish"

## Notes
See https://github.com/18F/micropurchase/issues/922 for reference. Want to make behavior consistent and I want this messaging to appear on the admin show page as well as the default show page. Admin should be able to unpublish from this page as well.

harrisj:
Fixes #1133 

localhost_3000_admin_auctions_9

This allows admins to unpublish future auctions before they go live. A few additional notes:

1. I added a few missing tests for the AdminAuctionPresenterFactory class.
2. I really don't like cucumber features that look for the presence of a table by seeing if it's in the second div from the top. I added an id to the `drafts` table on the page because it seems less brittle than some approach that would break in an unclear fashion if we reordered the page.

### DIFF
--- a/app/models/admin_auction_status_presenter_factory.rb
+++ b/app/models/admin_auction_status_presenter_factory.rb
@@ -8,6 +8,8 @@ class AdminAuctionStatusPresenterFactory
   def create
     if auction.payment_confirmed? || auction.c2_paid?
       Object.const_get("C2StatusPresenter::#{c2_status}").new(auction: auction)
+    elsif future? && auction.published?
+      AdminAuctionStatusPresenter::FuturePublished.new(auction: auction)
     elsif !auction.pending_delivery?
       Object.const_get("AdminAuctionStatusPresenter::#{status}").new(auction: auction)
     else
@@ -16,6 +18,10 @@ class AdminAuctionStatusPresenterFactory
   end
 
   private
+
+  def future?
+    AuctionStatus.new(auction).future?
+  end
 
   def status
     auction.status.camelize
--- a/app/presenters/admin_auction_status_presenter/future_published.rb
+++ b/app/presenters/admin_auction_status_presenter/future_published.rb
@@ -1,0 +1,22 @@
+class AdminAuctionStatusPresenter::FuturePublished < AdminAuctionStatusPresenter::Base
+  def header
+    I18n.t('statuses.admin_auction_status_presenter.future_published.header')
+  end
+
+  def body
+    I18n.t(
+      'statuses.admin_auction_status_presenter.future_published.body',
+      start_date: start_date
+    )
+  end
+
+  def action_partial
+    'admin/auctions/unpublish'
+  end
+
+  private
+
+  def start_date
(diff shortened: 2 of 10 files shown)


app/models/admin_auction_status_presenter_factory.rb
```diff
@@ -8,6 +8,8 @@ def initialize(auction:)
   def create
     if auction.payment_confirmed? || auction.c2_paid?
       Object.const_get("C2StatusPresenter::#{c2_status}").new(auction: auction)
+    elsif future? && auction.published?
+      AdminAuctionStatusPresenter::FuturePublished.new(auction: auction)
```
jessieay: thoughts on re-using https://github.com/18F/micropurchase/blob/develop/app/presenters/bid_status_presenter/future/admin.rb ?

jessieay:
I think this is providing the same content as https://github.com/18F/micropurchase/blob/develop/app/presenters/bid_status_presenter/future/admin.rb which was added at https://github.com/18F/micropurchase/commit/4f742f4e5675eb3af2650ad64a50d916dd8aa371

do you think it might make sense to re-use that logic? I love that you've I18n'd stuff here, we should def do that for the other presenter too

jessieay:
Actually, I want to edit my last statement: I think what you have here is the same but *better* (eg: link in partial)-- maybe use this logic / presenter for that view as well? (in any case: remove the duplication)

app/models/admin_auction_status_presenter_factory.rb
harrisj: Ohhhhhh

harrisj:
So, to be clear you want me to clean up https://github.com/18F/micropurchase/blob/develop/app/presenters/bid_status_presenter/future/admin.rb with some of the logic here?

jessieay:
@harrisj I think we could use the same presenter for both! but your call

harrisj:
I see now. I would actually like to keep the BidStatusPresenter::Future::Admin scheme I guess but add support for the action_partial capacity to those presenters. This would allow me to eliminate the new presenter and perhaps even reorganize some of the other AdminAuctionStatusPresenter classes? But should we do that in this PR or move to a future PR?

harrisj:
Oh wait, I see that action_partial is already in there. Trying things out.

harrisj:
I think I want to make consolidation a future PR

jessieay:
👍  we will consolidate via #1167

Issue #1169, opened 2016-09-07
Intermittently failing test related to timezones


jessieay:
```
 1) Api::V0::BusinessDaysController show should handle holidays correctly
     Failure/Error: expect(json['date']).to eq("September 09, 2016 01:00:00 PM EDT")
     
       expected: "September 09, 2016 01:00:00 PM EDT"
            got: "September 08, 2016 01:00:00 PM EDT"
     
       (compared using ==)
     # ./spec/controllers/api/v0/business_days_controller_spec.rb:14:in `block (3 levels) in <top (required)>'
```

when this spec was run at 9:45pm PT, it failed

PR #1171, opened 2016-09-07
Fix timezone-related failure

jessieay:
* Our holiday loading is only for future holidays so I need to forcibly
  load it in those places because that test is in the past
* Closes https://github.com/18F/micropurchase/issues/1169


harrisj:
Will approve and merge

PR #1181, opened 2016-09-07
Added nofollow and noindex tags to unpublished auctions

adelevie:
merge this instead of https://github.com/18F/micropurchase/pull/1144


public/robots.txt
```diff
@@ -2,4 +2,4 @@
 #
 # To ban all spiders from the entire site uncomment the next two lines:
 # User-agent: *
-# Disallow: /
+# Disallow: *
```
jessieay: I think we should revert this back to what it was previously?

PR #1182, opened 2016-09-07
Remove code related to auction preview

### LINKED ISSUE #1115: Eliminate preview method in admin/auctions_controller?
Now that we have enhanced the display of auctions for `admin_auctions#show` it seems like there is no need to use the `preview` method, and although the tests are still there and the routes, I don't think there is a link to the preview action anymore. Am I correct? Can we remove that code?

jessieay:
* We no longer link to the preview
* Instead, we have the tokenized URL
* Closes #1115


harrisj:
YAY! Waiting on Travis

PR #1184, opened 2016-09-07
Show notes on admin/auctions#show

### LINKED ISSUE #1023: Admin view - Auction Show Page - Notes field
This story is to display the notes (if there are any) in a box beneath the status box on the auction show page. See the attached screenshot. This is only visible to admins, and it only shows for auctions that have notes. We may want to name this admin_notes or something a bit more obvious

micropurchase.18f.gov/admin/auctions/[id]

screen shot 2016-08-09 at 10 33 34 am

jessieay:
* Closes #1023

When there are notes:

screen shot 2016-09-07 at 3 01 13 pm

When there are no notes:

screen shot 2016-09-07 at 3 02 26 pm

### DIFF
--- a/app/view_models/admin/auction_show_view_model.rb
+++ b/app/view_models/admin/auction_show_view_model.rb
@@ -23,6 +23,14 @@ class Admin::AuctionShowViewModel < Admin::BaseViewModel
     end
   end
 
+  def admin_notes_partial
+    if auction.notes.present?
+      'admin/auctions/notes'
+    else
+      'components/null'
+    end
+  end
+
   def admin_auction_status_presenter
     AdminAuctionStatusPresenterFactory.new(auction: auction).create
   end
@@ -60,10 +68,6 @@ class Admin::AuctionShowViewModel < Admin::BaseViewModel
     auction.title
   end
 
-  def summary
-    auction.summary
-  end
-
   def relative_time
     status_presenter.relative_time
   end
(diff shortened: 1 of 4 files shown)


app/view_models/admin/auction_show_view_model.rb
```diff
@@ -60,10 +68,6 @@ def title
     auction.title
   end
 
-  def summary
```
harrisj: Getting rid of these because they aren't being used?

jessieay: yep!

app/view_models/admin/auction_show_view_model.rb
```diff
@@ -82,10 +86,6 @@ def html_summary
     MarkdownRender.new(auction.summary).to_s
   end
 
-  def admin_edit_auction_partial
```
harrisj: So we don't have the Edit button on admin auction show?

jessieay: we do, I just inlined this.  (` <%= render partial: 'auctions/edit_auction_link', locals: { view_model: @view_model } %>` ) I think this method was leftover from a world where we shared view models between admin and regular auction  show pages.

harrisj: Got it. Thanks!

harrisj:
LGTM. Merging

Issue #1174, opened 2016-09-07
New Page - Getting Started


mtorres253:
As a visitor to the site, I want to know how to get started with micropurchase (what I need to do to participate), so there should be a page on the site that provides that information

- [ ] **Scenario: Visitor clicks navigates to /docs/getting_started/**
- Given I am a visitor to the site
- When I navigate to micropurchase.18f.gov/docs/getting_started/
- Then I should see this content on that page: https://pages.18f.gov/micropurchase/docs/getting-started/

jessieay:
once this is done, add link to about page (see https://github.com/18F/micropurchase/issues/1161)

PR #1186, opened 2016-09-07
Add about page

### LINKED ISSUE #1161: New Page  - About page
As a visitor to the micropurchase site, I want to know about the project itself. So, there should be a page where I can go to read about it.

- [ ] **Scenario: Visitor goes to about page via "About" link in footer**
- Given I am a visitor to the site
- When I click on the about link in the footer
- Then I should navigate to /about/
- And the this page should contain the following content

https://pages.18f.gov/micropurchase/about/

## Note
There is some duplication on this page of numbers that already exist on the insights page, information in the about.yml file and blog posts. Let's discuss the best way to handle this

Per our discussion today: 

1. Instead of by the number section, just link to insights
2. Hard code the rest of it

jessieay:
* Use high voltage https://github.com/thoughtbot/high_voltage
* Closes #1161
* Removed sentence about getting started guide, but added note to #1174
  to add that when we create the page

screen shot 2016-09-07 at 4 39 30 pm


harrisj:
LGTM!

PR #1187, opened 2016-09-08
Fix staging seed task

jessieay:
* `delivered` is no longer a trait
* Deploy to staging failed because this would not run
* Next, we should write a test for this :)


harrisj:
LGTM. Sorry for breaking this!

Issue #1167, opened 2016-09-07
Consolidation of logic in BidStatusPresenter/AdminAuctionStatusPresenter classes


harrisj:
Per the comments on #1159, there is some redundancy between the BidStatusPresenter::*::Admin classes and AdminAuctionStatusPresenter classes. I think @jessieay makes a very good point on this, but I would like to make this the subject of a future PR. It seems like we could 

1. Migrate some of the functionality in the AdminAuctionStatusPresenter classes (i18n, action_partial) to their equivalents/new classes in BidStatusPresenter
2. Eliminate the AdminAuctionStatusPresenter hierarchy entirely

Thoughts?

jessieay:
right now https://github.com/18F/micropurchase/blob/develop/app/presenters/admin_auction_status_presenter/future_published.rb and https://github.com/18F/micropurchase/blob/develop/app/presenters/bid_status_presenter/future/admin.rb are returning the same content -- can we fix that in this PR and keep the rest of the changes for some future time?

jessieay:
@harrisj did you intend #1199 to close this or do we want to do more work on it?

harrisj:
I guess that is fine since we aren't eliminating the AdminAuctionStautsPresenter hierarchy. I haven't renamed any C2StatusPresenters

jessieay:
works for me - closing

PR #1188, opened 2016-09-08
WIP: Consolidate view models for auction admin status

harrisj:
Following the suggestion in #1167, this is an effort to clean up logic and consolidate the AuctionAdminStatusPresenter and BidStatusPresenter view models into one place.

The rough plan of work is as follows:

- [X] AdminAuctionStatusPresenter::Future -> BidStatusPresenter::Future::Admin
- [X] AdminAuctionStatusPresenter::WorkInProgress -> BidStatusPresenter::Over::Admin::WorkInProgress
- [x] AdminAuctionStatusPresenter::PendingAcceptance -> BidStatusPresenter::Over::Admin::PendingAcceptance
- [x] AdminAuctionStatusPresenter::Accepted -> BidStatusPresenter::Over::Admin::Accepted
- [x] AdminAuctionStatusPresenter::Rejected -> BidStatusPresenter::Over::Admin::Rejected
- [ ] Remove AdminAuctionStatusPresenterFactory

The end game is to remove the AdminAuctionStatusPresenter hierachy and its Factory entirely (although we might rename it a C2StatusFactory or something). This makes it a little clear what is wear and makes it so the factory for BidStatusPresenter mirrors our matrix of supported states


jessieay:
I wonder if we should consolidate these a bit less or differently. 

The way I've been doing it in the past is:

`BidStatusPresenter` is the namespace for anything that the person (guest/admin/vendor) sees on the public auctions#show page
`C2StatusPresenter` is for what the admin seeds on admin/auctions#show
`AdminAuctionStatus` is for what the admin sees on admin/auctions#show when not related to C2

Maybe we keep two namespaces to clarify what is for the admin-only view and what is for the regular show page (even though some of those statuses are specific to admins) ?

The one exception to this is that we now have one view, (the "unpublish" button view) that is for both the regular auctions#show page and for the admin/auctions#show page. I think it would be good to share logic between those but still have them in separate namespaces.

harrisj:
Ooops, I suppose I should read things before I go ahead on this. I don't know how I feel on this one, is it better to have one ginormous factory so you can see exactly how complex it is or to try a bunch of specific factories (where we would probably also have a different one for guests)

jessieay:
to be clear, we still have two giant factories after this PR, they are just returning presenters in the same namespace rather than view-specific namespaces.

harrisj:
I see, so it seems like I should start over perhaps and try a different approach. That makes sense but I would like to figure out if we can rename C2Status -> AuctionAdminStatus at least. Suggestions?

harrisj:
Deprecating in favor of #1199

PR #1189, opened 2016-09-08
Remove staging prime task

jessieay:
* Was causing problems deploying to staging
* Not sure what the problems are, but clearly this is more worry than it
  is worth


harrisj:
You were too beautiful for this world, my little rake task

PR #1199, opened 2016-09-09
Remove BidStatusPresenter::Future::Admin

harrisj:
Following the advice in issue #1167, I am having BidStatusPresenter::Future::Admin use the AdminAuctionStatusPresenter::Future (formerly FuturePublished) presenter instead. Just committing this as a single PR to replace #1188 

A next step might be to rename C2StatusPresenter presenters to be AdminAuctionStatusPresenter, but that can be a future PR


features/admin_views_future_auction.feature
```diff
@@ -2,14 +2,26 @@ Feature: Admin views future auction
...
     Then I should see a "Coming Soon" label
     And I should not see the bid form
     And I should see the future auction message for admins
-    When I click on the "Un-publish" button
+
+    When I click on the unpublish button
+    Then I should be on the admin needs attention auctions page
+    And I should see a success message confirming that the auction was unpublished
+
+  Scenario: Viewing the admin auction page
```
jessieay: nice! I think moving this makes a lot of sense

jessieay: (moving the spec, that is)

jessieay:
great one!! 🌈

PR #1203, opened 2016-09-12
Admins cannot bid

jessieay:
* Add logic to rules that doesn't allow admins to bid
* Closes https://github.com/18F/micropurchase/issues/995


harrisj:
Should we have an `admin_bids_on_an_auction.feature`?

jessieay:
@harrisj I prefer to leave negative cases to lower-level unit tests. Mostly so our feature specs can remain smaller in number and our test suite can stay fast

jessieay:
thoughts on that @harrisj  ?

harrisj:
Okay, that works for me. Merging

PR #1205, opened 2016-09-12
Admin can unpublish "other" purchase card auctions

jessieay:
* Closes https://github.com/18F/micropurchase/issues/1162

### DIFF
--- a/app/models/auction_status.rb
+++ b/app/models/auction_status.rb
@@ -20,7 +20,7 @@ class AuctionStatus
   end
 
   def work_in_progress?
-    auction.delivery_url.present? && !(auction.pending_acceptance? || auction.accepted? || auction.rejected?)
+    auction.delivery_url.present? && auction.pending_delivery?
   end
 
   private
--- a/app/view_models/admin/auction_show_view_model.rb
+++ b/app/view_models/admin/auction_show_view_model.rb
@@ -14,15 +14,6 @@ class Admin::AuctionShowViewModel < Admin::BaseViewModel
     end
   end
 
-  def admin_auction_status_partial
-    if !auction.pending_delivery? ||
-       auction.purchase_card == 'default'
-      'auctions/status'
-    else
-      'components/null'
-    end
-  end
-
   def admin_notes_partial
     if auction.notes.present?
       'admin/auctions/notes'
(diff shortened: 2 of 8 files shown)


jessieay:
hold off on merging for a bit -- I want to fix some edge cases first

app/view_models/admin/auction_show_view_model.rb
```diff
@@ -14,15 +14,6 @@ def csv_report_partial
     end
   end
 
-  def admin_auction_status_partial
```
jessieay: disclaimer: because we are removing this, we are showing statuses for all auctions at all states now in the admin interface. We need to fill some gaps with that (eg: right now, if an "other pcard" auction is pending delivery, we show its c2 status...which makes no sense) but should happen soon (created cards for the cases)

app/models/auction_status.rb
```diff
@@ -20,7 +20,7 @@ def expiring?
   end
 
   def work_in_progress?
-    auction.delivery_url.present? && !(auction.pending_acceptance? || auction.accepted? || auction.rejected?)
+    auction.delivery_url.present? && auction.pending_delivery?
```
harrisj: Yeah, that is better

app/view_models/admin/auction_show_view_model.rb
harrisj: I like removing code we don't need

features/admin_views_future_auction.feature
```diff
@@ -2,12 +2,10 @@ Feature: Admin views future auction
   As an admin of the Micro-purchase system
   I want to be able to unpublish future auctions
 
-  Background:
+  Scenario: Viewing the auction page
```
harrisj: Argument against Background?

jessieay: same as `before` statements -- it takes the dependencies of the test outside of the test. When I read a spec, I want to know all of the required pre-steps when reading it. Scrolling up to a `background` or `before` statement isn't so bad in a case like this, but when you become overly dependent on those things you suddenly have a very complex setup and are not sure which tests require it or not. Rather than carefully deciding when things have gone "too far", I find it easier to just not use `background` or `before`

harrisj:
LGTM. Merging

PR #1206, opened 2016-09-12
Add "Getting Started" page

jessieay:
* Add some styles so we have bullets on lists (USWDS removes them)
* Link to page on "About"
* Closes https://github.com/18F/micropurchase/issues/1174

screen shot 2016-09-12 at 12 16 44 pm


harrisj:
YAY!

PR #1208, opened 2016-09-12
Fix issue with receipt URL

jessieay:
* We were getting errors when we tried to update the Receipt URL for
  auctions that had received payment confirmation
* This was because completed requests cannot be updated on C2
* This PR removes the code that tried to update the request
* Instead, we are surfacing the receipt URL in the admin interface
* Closes https://github.com/18F/micropurchase/issues/1177

when payment is confirmed:

screen shot 2016-09-12 at 3 48 50 pm

when payment is not confirmed:

screen shot 2016-09-12 at 3 48 17 pm


harrisj:
Argh, C2, why you gotta play that way?

PR #1209, opened 2016-09-12
Move FAQ page to `pages` dir

jessieay:
* That is where we keep static pages now


harrisj:
Looks like it needs to be rebased

jessieay:
@harrisj updated!

harrisj:
We are go then!

PR #1220, opened 2016-09-14
Clarify status confirmation for sealed-bid auction

### LINKED ISSUE #985: Vendor workflow and messaging - Auction show page - Bid on auction
As a vendor, I want to see clear messaging around whether my bid was accepted or not (in either type of auctino), so the system should show me

**Context:**
- And there is a sealed-bid auction
- And I am qualified to bid on that auction
- And I am on the auction show page (micropurchase.18f.gov/auctions/[id])

- [x] **Scenario: Vendor sees javascript bid confirmation after placing bid - sealed-bid**
- And I have NOT bid on the auction
- And I have entered a bid amount in the bid box
- When I click the "Place Bid" button
- And there are no errors
- Then I should see a javascript confirmation popup

> micropurchase.18f.gov says
> Are you sure you want place a bid for $[bid_amount]?

screen shot 2016-09-02 at 10 50 23 am

- [x] **Scenario: Vendor confirms bid - sealed-bid**
- When I click on the "OK" button in the javascript confirmation
- And there are no errors
- Then my bid should be placed
- And I should see a success message

> Bid placed
> You placed a bid of $[auction.currentBid] on [auction.bidDate].

screen shot 2016-08-08 at 2 48 27 pm

- [x] **Scenario: Vendor cancels out of javascript confirmation - sealed-bid**
- When I click on the "Cancel" button in the javascript confirmation
- Then my bid should not be placed
- And the javascript confirmation box should close

harrisj:
Fixes #985

This is the last checkbox on there. But it seems like it already is implemented. So I just renamed a few steps

### DIFF
--- a/features/step_definitions/auction_status_steps.rb
+++ b/features/step_definitions/auction_status_steps.rb
@@ -53,8 +53,10 @@ Then(/^I should the open auction message for admins$/) do
   )
 end
 
-Then(/^I should see the time I placed my bid$/) do
+Then(/^I should see a status confirmation for my sealed bid$/) do
   bid = @auction.bids.last
+  expect(bid.amount).to eq(@bid_amount.to_i)
+  expect(bid.bidder).to eq(@user)
 
   expect(page).to have_content(
     I18n.t(
--- a/features/step_definitions/auction_status_steps.rb
+++ b/features/step_definitions/auction_status_steps.rb
@@ -53,7 +53,7 @@ Then(/^I should the open auction message for admins$/) do
   )
 end
 
-Then(/^I should see a status confirmation for my sealed bid$/) do
+Then(/^I should see a status message that confirms I placed a sealed bid$/) do
   bid = @auction.bids.last
   expect(bid.amount).to eq(@bid_amount.to_i)
   expect(bid.bidder).to eq(@user)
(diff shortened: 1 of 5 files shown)


jessieay:
Nice! are we doing multi-bid later

features/step_definitions/auction_status_steps.rb
```diff
@@ -53,8 +53,10 @@
   )
 end
 
-Then(/^I should see the time I placed my bid$/) do
+Then(/^I should see a status confirmation for my sealed bid$/) do
```
jessieay: I wonder if there is a clearer name that "status confirmation" -- feels a bit vague to me

harrisj:
Oops, I should actually read the issue. The status box should be green and the text is changed

18f_micro-purchase_-_enhanced_bifurcated_infrastructure

One thing that bothers me is there is nothing in the steps right now that says "This status box should be green" or such compared to the usual status. This is something I am confirming manually, but it's something we might want to codify before doing same for reverse auctions, since we might want it to be green if you have lowest bid, blue if you don't

features/step_definitions/auction_status_steps.rb
harrisj: What do we call this? The status box?

jessieay: the format we are using in the rest of this file is a bit inconsistent but mostly this format:

> I should see the unpublished auction message for guests

I think we are not calling it 'status confirmation' anywhere else. status box? maybe

> I should see a status message that confirms I placed a bid

?

harrisj: Okay, will try that

jessieay:
Nice one!

PR #1221, opened 2016-09-14
Vendor messaging - Payment URL needed - Auction Show Page

jessieay:
* if an auction is accepted and the winning vendor does not have a payment url,
  they receive an email asking them to update their profile.
* when they view the auction, they will now also see this

screen shot 2016-09-13 at 8 30 12 pm

* Closes https://github.com/18F/micropurchase/issues/1099

### DIFF
--- a/app/models/bid_status_presenter_factory.rb
+++ b/app/models/bid_status_presenter_factory.rb
@@ -34,7 +34,9 @@ class BidStatusPresenterFactory
   end
 
   def over_winning_bidder_message
-    if auction.pending_acceptance?
+    if auction.accepted? && auction.accepted_at.nil?
+      BidStatusPresenter::Over::Vendor::Winner::PendingPaymentUrl
+    elsif auction.pending_acceptance?
       BidStatusPresenter::Over::Vendor::Winner::PendingAcceptance
     elsif auction.accepted?
       BidStatusPresenter::Over::Vendor::Winner::PendingPayment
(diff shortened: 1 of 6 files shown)


jessieay:
cyclomatic complexty 💭

app/models/bid_status_presenter_factory.rb
```diff
@@ -34,7 +34,9 @@ def future_message
   end
 
   def over_winning_bidder_message
-    if auction.pending_acceptance?
+    if auction.accepted? && auction.accepted_at.nil?
```
harrisj: So, we use this instead of looking at the payment_url field?

features/step_definitions/auction_create_steps.rb
```diff
@@ -202,6 +202,7 @@
   @auction = FactoryGirl.create(
     :auction,
     :with_bids,
+    :closed,
```
harrisj: good catch

app/models/bid_status_presenter_factory.rb
jessieay: yes, this is how we know something was accepted but the payment URL is missing. Kinda weird, perhaps something to change in a future PR. Added issue here: https://github.com/18F/micropurchase/issues/1222

jessieay:
@harrisj re-review when you get a chance?

harrisj:
I thought we were going to implement a PR for #1222 first?

harrisj:
Going to merge now but we will have to revise if we add another enum state

jessieay:
@harrisj yes, will def revise as I finish #1222

PR #1224, opened 2016-09-14
Small fixes on static pages

jessieay:
* Used named routes so we can an error if file paths change
* Use straight apostrophe not curly


harrisj:
LGTM!

PR #1226, opened 2016-09-14
Fix the tiniest problem

jessieay:
* `mail_to` needs 2 links in rails :(


adelevie:
makes so much sense, rails ...

merging when 🍏

jessieay:
@adelevie  🍏

Issue #1057, opened 2016-08-17
Curated docs page and help pages


mtorres253:
As a vendor or a customer, I want to be able to easily access documentation that can help me understand how micropurchase works and how I can get involved so there should be a clear way for me to access that information via the UI

**Context:**
- Given I am a visitor to the micropurchase site (micropurchase.18f.gov)

- [x] **Scenario: Visitor accesses [/docs/](https://pages.18f.gov/micropurchase/docs/)  page via footer**
- When I am on any page on the site
- Then I should see a link labeled "Docs" in the footer
- And that link should go to micropurchase.18f.gov/docs/
- And I should see the following content on that page: https://pages.18f.gov/micropurchase/docs/

- [x] **Scenario: Visitor clicks "Policies" link on /docs/ page**
- When I click on the "Policies" link
- Then I should go to /policies/

- [x] **Scenario: Visitor clicks "Placing a Bid" link on /docs/ page*
- When I click on the "Placing a Bid" link in the Vendor section of the docs page
- Then I should go to /docs/placing_a_bid/
- And I should see a link labeled "All Documentation" at the top of that page that goes back to the /docs/ page 

- [x] **Scenario: Vendor clicks "Getting Paid" link on /docs/ page**
- When I click on the "Getting Paid" link in the Vendor section of the docs page
- Then I should go to /docs/getting_paid/
- And I should see a link labeled "All Documentation" at the top of that page that goes back to the /docs/ page 

- [x] **Scenario: Visitor clicks "Getting Started" link on /docs/ page and sees getting started content**
- When I click on the "Getting Started" link
- Then I should to to /docs/getting_started/
- And I should see a link labeled "All Documentation" at the top of that page that goes back to the /docs/ page 

- [x] **Scenario: Customer clicks "Creating an Auction" link on /docs/creating_an_auction/ page**
- When I click on the "Creating an Auction" link
- Then I should go to /docs/creating_an_auction/
- And I should see a link labeled "All Documentation" at the top of that page that goes back to the /docs/ page 

Footer links are now:
About -> https://micropurchase.18f.gov/about
Policies -> https://micropurchase.18f.gov/policies
Docs -> https://micropurchase.18f.gov/docs
Insights -> https://micropurchase.18f.gov/insights
API -> https://micropurchase.18f.gov/api

## Notes
Does this involve creating a pages controller?

PR #1229, opened 2016-09-14
Create curated docs page

jessieay:
* Remove old `faq` page that we no longer link to
* Closes https://github.com/18F/micropurchase/issues/1057

page:

screen shot 2016-09-14 at 4 26 08 pm

footer:

screen shot 2016-09-14 at 4 31 32 pm


harrisj:
Should I be merging these?

jessieay:
I think so! Per conversation yesterday, 100% design parity with prototype
is not possible. Always a good idea to check back and compare with
prototype for any obvious and changeable differences (that's why I include
screenshots) but eyeballing it is fine.

harrisj:
LGTM then

Issue #1222, opened 2016-09-14
Add enum to represent "accepted but no payment url"


jessieay:
Right now, we just look at whether there is an `accepted_at` timestamp (that is not set if the winner doesnt have a payment URL). That is confusing and could be clarified with an enum under `status`

jessieay:
see https://github.com/18F/micropurchase/pull/1221#discussion_r78751741 for more context

harrisj:
I approve of this approach

harrisj:
I almost feel like this should be another `c2_status` to be honest

jessieay:
This state could exist for either purchase card.

PR #1230, opened 2016-09-15
Admin auction status for pending payment URL

harrisj:
Addresses one part of #1153 

Note that the original issue includes the `approver_name` and `approved_at` time but the first is not recorded and the second is not set if the payment_url is nil

Note that this will need to be revised when #1222 is implemented. At that time, we could also add `accepted_at` to the status message

### DIFF
--- a/app/models/admin_auction_status_presenter_factory.rb
+++ b/app/models/admin_auction_status_presenter_factory.rb
@@ -12,6 +12,8 @@ class AdminAuctionStatusPresenterFactory
       AdminAuctionStatusPresenter::Future.new(auction: auction)
     elsif work_in_progress?
       AdminAuctionStatusPresenter::WorkInProgress.new(auction: auction)
+    elsif auction.accepted? && auction.accepted_at.nil?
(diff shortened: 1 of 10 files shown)


app/models/admin_auction_status_presenter_factory.rb
```diff
@@ -12,6 +12,8 @@ def create
       AdminAuctionStatusPresenter::Future.new(auction: auction)
     elsif work_in_progress?
       AdminAuctionStatusPresenter::WorkInProgress.new(auction: auction)
+    elsif auction.accepted? && auction.accepted_at.nil?
```
jessieay: yess can't wait to update these l8r

app/presenters/admin_auction_status_presenter/base.rb
```diff
@@ -17,6 +17,14 @@ def alert_css_class
 
   protected
 
+  def winner_url
+    Url.new(
```
jessieay: nice!

jessieay:
This looks gr8! mergin'

PR #1231, opened 2016-09-15
Create enum status for accepted auctions where winner was no payment URL

jessieay:
* Closes https://github.com/18F/micropurchase/issues/1222


app/controllers/admin/auction_acceptances_controller.rb
```diff
@@ -1,15 +1,13 @@
 class Admin::AuctionAcceptancesController < Admin::BaseController
   def update
     @auction = Auction.find(params[:id])
-    @auction.status = :accepted
```
harrisj: Yes!

app/models/admin_auction_status_presenter_factory.rb
```diff
@@ -7,18 +7,16 @@ def initialize(auction:)
...
-      AdminAuctionStatusPresenter::PendingPaymentUrl.new(auction: auction)
+      AdminAuctionStatusPresenter::WorkInProgress
     elsif !auction.pending_delivery?
-      Object.const_get("AdminAuctionStatusPresenter::#{status}").new(auction: auction)
+      Object.const_get("AdminAuctionStatusPresenter::#{status}")
     else # if auction.purchase_card == 'default'
-      Object.const_get("C2StatusPresenter::#{c2_status}").new(auction: auction)
-    end
+      Object.const_get("C2StatusPresenter::#{c2_status}")
+    end.new(auction: auction)
```
harrisj: This is consistent with BidStatusPresenter, so I approve.

My only complaint about this approach is that it forces you to have the same arguments for all presenters even if they don't use them. For instance, when I was doing the BidStatusPresenter::Available::Vendor::BidError presenter, I really didn't need the user but I did need an bid error. To make it work, I had to have it so all auction presenters accepted a bid_error argument, even though none of the other ones used them. It seems like this could be an issue down the road, not enough to affect this PR, but I wanted to note it.

app/models/bid_status_presenter_factory.rb
```diff
@@ -34,14 +34,8 @@ def future_message
...
-    if auction.accepted? && auction.accepted_at.nil?
-      BidStatusPresenter::Over::Vendor::Winner::PendingPaymentUrl
-    elsif auction.pending_acceptance?
-      BidStatusPresenter::Over::Vendor::Winner::PendingAcceptance
-    elsif auction.accepted?
-      BidStatusPresenter::Over::Vendor::Winner::PendingPayment
-    elsif auction.rejected?
-      BidStatusPresenter::Over::Vendor::Winner::Rejected
+    if !auction.pending_delivery?
+      Object.const_get("BidStatusPresenter::Over::Vendor::Winner::#{auction.status.camelize}")
```
harrisj: Ooooh

app/models/concerns/auction_scopes.rb
```diff
@@ -3,6 +3,7 @@ module AuctionScopes
 
   included do
     scope :accepted, -> { where(status: statuses['accepted']) }
+    scope :accepted_pending_payment_url, -> { where(status: statuses['accepted_pending_payment_url']) }
     scope :delivery_accepted, -> { where(status: statuses['accepted']).where.not(accepted_at: nil) }
```
harrisj: This scope is slightly different now. It seems like before it meant ("delivery accepted and we have a payment URL") and now the second clause is redundant. I would look at if this is still used and we might need to modify name/definition

app/models/admin_auction_status_presenter_factory.rb
jessieay: Having a common interface to all presenters is nice, which is why we have the `Base` classes. As of right now, they can all be initialized the same way. If we get to a point where they are very different, that might be the code telling us the we need separate factories

app/models/concerns/auction_scopes.rb
jessieay: you are right! updated at 261a2fb to remove all references of `delivery_accepted` and replace with `accepted` since now we know that all `accepted` auctions have a payment URL

jessieay:
@harrisj ready for re-review!

harrisj:
Do you mind looking at `UpdateAuction` and `AuctionShowViewModel` to make sure they aren't using `accepted_at` in a wrong way now?

jessieay:
@harrisj yes, those are correct. In `UpdateAuction` we've already assigned attributes so we want to know if the `accepted` param was passed for the first time (we know that by looking for a nil `accepted_at` timestamp

in `AuctionShowViewModel` we are only showing the accepted at timestamp when it is present, which makes sense

harrisj:
I thought so, but I wanted to check! Thanks! Merging

Issue #931, opened 2016-07-21
Create new auction flow - other p-card


mtorres253:
As an admin, when I'm creating a new auction that does NOT use the default p-card, I want to have clear instructions on how to publish it, so the system should tell me/show me.

- [ ] **Scenario: Admin sees correct options for creating auction NOT using default p-card**
- Given that I am an authenticated admin user
- And I am creating a new auction (https://micropurchase.18f.gov/admin/auctions/new)
- And the auction I'm creating does NOT use the default p-card payment option
- When I click on the "Create New Auction" button
- And there are no errors
- Then my auction should be saved
- And I should be taken to the auction show page for the newly created auction (https://micropurchase.18f.gov/admin/auctions/[id])
- And I should see a message in the status box

> Ready to publish
> This auction is ready to publish!

- And I should also see a "Publish" button

- [x] **Scenario: Admin publishes new auction using other p-card that is coming soon**
- Given I am an authenticated admin
- And there is an auction that is publishable (newly created auction using external p-card, auction using 18F p-card that has C2 approval)
- And the auction start date is in the future
- And I am on the show page for that auction (https://micropurchase.18f.gov/admin/auctions/[id])
- When I click the "Publish Auction" button
- Then the auction should be published
- And the page should reload
- And I should see the following message in the status box

> Coming soon
> This auction is visible to the public but is not currently accepting bids. It will open on [auction.startDate]. If you need to take it down for whatever reason, press the “Un-publish” button, below.

![screen shot 2016-08-30 at 3 12 42 pm](https://cloud.githubusercontent.com/assets/100149/18109077/44966388-6ec4-11e6-9a6a-f1dd6b7714ad.png)

- And I should see an Un-publish button

jessieay:
this seems like it overlaps with  #878

mtorres253:
Keep in the backlog until #878 is done

jessieay:
do ready to publish for both pcards

PR #1232, opened 2016-09-15
Add "ready to publish" status

jessieay:
* When default pcard, this happens when c2 request is approved
* Otherwise, it is for any future auction that is unpublished.
* Closes https://github.com/18F/micropurchase/issues/931

screen shot 2016-09-15 at 4 17 24 pm


harrisj:
LGTM! I wonder if we should do something about the Cyclomatic complexity dings for those factories

jessieay:
@harrisj Yes! I am scheming and hope to do something to improve this today.

PR #1233, opened 2016-09-15
Update c2_status when auction is paid

jessieay:
* Previous we just updated `paid_at` timestamp
* This created a bug (closes https://github.com/18F/micropurchase/issues/1217)


harrisj:
Oooh, good catch!

PR #1236, opened 2016-09-16
Auctions pending payment URL need attention

### LINKED ISSUE #1153: Admin messaging - Payment URL needed - Auction Show Page and Needs Attention
As an admin user, if I've accepted an auction but I'm unable to pay the vendor because there is no payment URL, I should see that in the UI where I would normally look for auction status.

- [x] **Scenario: Admin sees message in status box for auction that payment URL is needed before payment can be made**
- When I'm on the auction show page (https://micropurchase.18f.gov/admin/auctions/[id])

> Payment URL needed
> [auction.acceptanceAdmin] accepted [auction.winner]’s pull request on [auction.acceptanceDate]. However, this vendor cannot be paid until they’ve specified a valid payment URL.

- And "[auction.winner]'s pull request" should link to the delivery URL for that auction

screen shot 2016-08-31 at 4 37 04 pm

- [ ] **Scenario: Admin sees status in the "Needs Payment" section of needs attention page that a payment URL is needed**
- And I am on the needs attention page (https://micropurchase.18f.gov/admin/auctions/needs_attention)
- Then I should see the auction listed in the "Payment Needed" section
- And in the column labeled "Status" I should see "Payment URL needed"

screen shot 2016-08-31 at 4 40 46 pm

harrisj:
Fixes #1153 

Make sure that auctions that are accepted but missing the payment URL show up in the needs_attention column

### DIFF
--- a/app/models/auction_query.rb
+++ b/app/models/auction_query.rb
@@ -63,7 +63,7 @@ class AuctionQuery
 
   def payment_needed
     relation
-      .accepted_and_pending
+      .accepted_or_accepted_and_pending_payment_url
       .not_paid
   end
 
--- a/app/models/auction_query.rb
+++ b/app/models/auction_query.rb
@@ -63,7 +63,7 @@ class AuctionQuery
 
   def payment_needed
     relation
-      .accepted
+      .accepted_and_pending
       .not_paid
   end
 
--- a/app/models/concerns/auction_scopes.rb
+++ b/app/models/concerns/auction_scopes.rb
@@ -4,7 +4,7 @@ module AuctionScopes
   included do
     scope :accepted, -> { where(status: statuses['accepted']) }
     scope :accepted_pending_payment_url, -> { where(status: statuses['accepted_pending_payment_url']) }
-    scope :accepted_and_pending, lambda {
+    scope :accepted_or_accepted_and_pending_payment_url, lambda {
       where(status: [statuses['accepted'],
                      statuses['accepted_pending_payment_url']])
     }
--- a/app/models/concerns/auction_scopes.rb
+++ b/app/models/concerns/auction_scopes.rb
@@ -4,6 +4,10 @@ module AuctionScopes
   included do
     scope :accepted, -> { where(status: statuses['accepted']) }
     scope :accepted_pending_payment_url, -> { where(status: statuses['accepted_pending_payment_url']) }
+    scope :accepted_and_pending, lambda {
+      where(status: [statuses['accepted'],
+                     statuses['accepted_pending_payment_url']])
+    }
     scope :default_purchase_card, -> { where(purchase_card: 0) }
     scope :delivery_url, -> { where.not(delivery_url: [nil, '']) }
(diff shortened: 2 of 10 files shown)


app/models/concerns/auction_scopes.rb
```diff
@@ -4,6 +4,10 @@ module AuctionScopes
   included do
     scope :accepted, -> { where(status: statuses['accepted']) }
     scope :accepted_pending_payment_url, -> { where(status: statuses['accepted_pending_payment_url']) }
+    scope :accepted_and_pending, lambda {
```
jessieay: I don't **think** this `lambda` is needed (it's also not the style we are using in the rest of this module but I might be missing something)

app/models/auction_query.rb
```diff
@@ -63,7 +63,7 @@ def c2_payment_needed
 
   def payment_needed
     relation
-      .accepted
+      .accepted_and_pending
```
jessieay: maybe `accepted_or_accepted_and_pending_payment_url` would be clearer?

app/views/admin/auctions/needs_attention/_payment_needed.html.erb
```diff
@@ -1,4 +1,4 @@
-<table class="usa-table-borderless">
+<table class="usa-table-borderless" id="table-payment">
```
jessieay: I'm not in love with adding css for the sole purpose of tests -- any way we can use xpath instead?

app/models/concerns/auction_scopes.rb
harrisj: According to our Rubocop, we use lambda for multiple-line lambdas (the `->` is just syntactic sugar), so that's why I added it

app/models/auction_query.rb
harrisj: Will do

app/models/concerns/auction_scopes.rb
jessieay: I think this doesn't need to be multiple lines

app/views/admin/auctions/needs_attention/_payment_needed.html.erb
jessieay: yeah I agree. hard to know. I just know that I generally avoid `ids` in any case in css unless it's for jquery so maybe change to a class instead?

features/step_definitions/admin_views_drafts_steps.rb
```diff
@@ -1,22 +1,21 @@
...
+
   within(:xpath, table_xpath) do
     expect(page).to have_content(@auction.title)
   end
 end
 
 Then(/^I should see a table listing all payment needed auctions$/) do
-  table_xpath = '//table[@id="table-payment"]'
-  expect(page).to have_xpath(table_xpath)
+  expect(page).to have_xpath(needs_attention_table_xpath('payment_needed'))
```
jessieay: ❤️

jessieay:
@harrisj merge conflicts, otherwise LGTM!

harrisj:
Rebased

PR #1239, opened 2016-09-16
Fix c2 status in drafts section of Needs Attention

jessieay:
* Were were using `AdminAuctionStatusPresenterFactory` which returns
  various presenters
* We want to return the c2 status presenter status if default pcard
* Otherwise, return "N/A"

Before:

screen shot 2016-09-16 at 2 11 17 pm

After:
screen shot 2016-09-16 at 2 09 47 pm


harrisj:
LGTM

PR #1240, opened 2016-09-16
Improve dev seeds

jessieay:
* Bidders should be `sam_verified` like in real life
* Create a client account so we can create new auctions easily (required)


harrisj:
LGTM

Issue #1248, opened 2016-09-19
Potential masking of C2 Status on show page for future auctions


harrisj:
In the AdminAuctionStatusController, we have 

```
  def default_purchase_card_presenter
    if auction.payment_confirmed? || auction.c2_paid?
      Object.const_get("C2StatusPresenter::#{c2_status}")
    elsif future? && auction.unpublished?
      AdminAuctionStatusPresenter::ReadyToPublish
    elsif future? && auction.published?
      AdminAuctionStatusPresenter::Future
    elsif work_in_progress?
      AdminAuctionStatusPresenter::WorkInProgress
    elsif !auction.pending_delivery?
      Object.const_get("AdminAuctionStatusPresenter::#{status}")
    else
      Object.const_get("C2StatusPresenter::#{c2_status}")
    end
  end
```

Some of the C2 status boxes would be realistically applied to future auctions (you approve it before you publish it). However, these would not be displayed on the admin screen because the factory looks for `future?` before checking `c2_status`. This situation passed the tests because they used a current available auction with C2 fields, even though this is not really possible in a realistic workflow.

After consultation with @jessieay, we decided on the following likely course of action:

1. Change C2 FactoryGirl traits to set other auction fields to likely situations (ie, this auction will be future if it's this C2_status. This might mean finding FG creates in the tests and revising them. But it should reveal some test failures that were false positives.
2. Eliminate the Object.const_get fallbacks in these constructors and order the conditions in the likely sequence they would naturally occur with high priority atemporal overrides first (ie, future first, then available, then work in progress, but if there is something that must be checked first, put it up top)

PR #1249, opened 2016-09-19
Clarify presenter logic for admins

jessieay:
* Move away from `Obj.const_get`
* Make sure our traits have realistic attributes (eg: unpublished
  auctions should be future by default)
* Fixes https://github.com/18F/micropurchase/issues/1248

### DIFF
--- a/app/models/admin_auction_status_presenter_factory.rb
+++ b/app/models/admin_auction_status_presenter_factory.rb
@@ -16,18 +16,26 @@ class AdminAuctionStatusPresenterFactory
   private
 
   def default_purchase_card_presenter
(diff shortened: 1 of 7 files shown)


jessieay:
sorry I made you mad code climate, but this is what we need for clarity right now

harrisj:
This looks great, but do we actually have Cucumber tests for when the C2 statuses should be displayed?

adelevie:
LGTM

jessieay:
@harrisj yes we do! they use the factories, so updating the factories made the cucumber tests break, which is what I used to guide the refactor process

harrisj:
LGTM then!

Issue #1251, opened 2016-09-19
Bug: no pending delivery presenter


jessieay:
when there is an "other" purchase card auction, there is an error at this line in the admin auction status factory:

 Object.const_get("AdminAuctionStatusPresenter::#{status}")

https://rpm.newrelic.com/accounts/1345535/applications/18638947?tw%5Bstart%5D=1474312290&tw%5Bend%5D=1474314090

Issue #1235, opened 2016-09-15
Admin messaging - Auction show page - Auction is accepting bids


mtorres253:
As an admin, when I go to the show page for an auction, and that auction is active (accepting bids) then I should see appropriate messaging in the status box

- [x]  **Scenario: Admin sees that an auction is accepting bids (but has no bids yet)- any type of auction/any payment method**
- Given I am an authenticated admin
- And there is an auction that is published
- And that auction is in the bid period (accepting bids)
- And there are NO bids yet
- When I go to the show page for that auction (micropurchase.18f.gov/admin/auctions/[id])
- Then I should see the following messaging in the status box

> Accepting bids
> This auction is accepting bids until [auction.endDate].

![screen shot 2016-09-15 at 4 54 57 pm](https://cloud.githubusercontent.com/assets/100149/18571168/254ef280-7b65-11e6-8e85-57b0de6e9c15.png)

- [ ]  **Scenario: Admin sees that an auction is accepting bids (bids exist) - any type of auction/any payment method**
- Given I am an authenticated admin
- And there is an auction that is published
- And that auction is in the bid period (accepting bids)
- And there are bids
- When I go to the show page for that auction (micropurchase.18f.gov/admin/auctions/[id])
- Then I should see the following messaging in the status box

> Accepting bids
> There are [auction.totalBids] bids on this auction, and [winningVendor.Name] is currently winning. This auction is accepting bids until [auction.endDate].

![screen shot 2016-09-15 at 4 57 03 pm](https://cloud.githubusercontent.com/assets/100149/18571198/723d3d5e-7b65-11e6-930d-cda88c7da322.png)

- And [auction.totalBids] should link to the bid history for that auction
- And [winningVendor.Name] should link to the vendor's page (https://micropurchase.18f.gov/admin/users/[id])

jessieay:
First part of this is completed via https://github.com/18F/micropurchase/pull/1253

harrisj:
Noting that @jessieay's work shows this message for available auctions at `/auctions/:id` path as well. I think that is good, but I wanted to just check. Do we want a similar parity for the second case when the admin looks at `/auctions/:id` and `/admin/auctions/:id`?

jessieay:
Note: the available auction message was already on `auctions/:id`, I just added it to `admin/auctions/:id`

PR #1253, opened 2016-09-19
Staging bug: no pending delivery presenter

jessieay:
* Closes https://github.com/18F/micropurchase/issues/1251
* Ensure we don't try to use the `PendingDelivery` presenter
* This should also help with
  https://github.com/18F/micropurchase/issues/1235 (but does not
  complete it)


jessieay:
@adelevie

harrisj:
LGTM

PR #1256, opened 2016-09-20
Fix issue with undefined method `end_date`

jessieay:
* Was not failing in spec because methods in cucumber were part of
  global namespace
* Add helper methods to module, which results in failure
* Fix failure
* Boom


harrisj:
Cucumber!!!

LGTM

Issue #1243, opened 2016-09-19
Remove `status` dropdown from edit form


jessieay:
We want an admin to be able to change the status, but through buttons and other actions rather than using the form.

jessieay:
the system won't work properly if admins change status in the form

mtorres253:
Good catch @jessieay we can discuss in planning today

PR #1257, opened 2016-09-20
Remove `status` dropdown for auction form

jessieay:
* Update status via bid box only
* This revealed a bug, which was that we didn't check that a vendor was
  eligible when accepting an auction via the bid box
* This PR moves the "vendor eligible" code to `AcceptAuction`
* Closes https://github.com/18F/micropurchase/issues/1243


app/controllers/admin/auction_acceptances_controller.rb
```diff
@@ -2,12 +2,18 @@ class Admin::AuctionAcceptancesController < Admin::BaseController
...
     @auction = Auction.find(params[:id])
 
-    AcceptAuction.new(
+    accept_auction = AcceptAuction.new(
       auction: @auction,
       payment_url: winning_bidder.payment_url
-    ).perform
+    )
+
+    if accept_auction.perform
```
harrisj: ooof, good catch

app/services/accept_auction.rb
```diff
@@ -7,6 +7,17 @@ def initialize(auction:, payment_url: '')
   end
 
   def perform
+    if vendor_ineligible?
+      auction.errors.add(:base, I18n.t('errors.update_auction.vendor_ineligible'))
+      false
+    else
```
harrisj: Oh man, well done

app/services/accept_auction.rb
```diff
@@ -19,7 +30,36 @@ def perform
...
 
-  private
+  def vendor_ineligible?
+    !winning_bidder_is_eligible_to_be_paid?
+  end
+
+  def winning_bidder_is_eligible_to_be_paid?
+    if auction_is_small_business?
+      reckoner = SamAccountReckoner.new(winning_bidder)
+      reckoner.set!
```
harrisj: This is basically a saving throw to see if their SAM status is good before bid right? Does the reckoner have switch from true -> false?

harrisj:
LGTM. Merging

app/services/accept_auction.rb
jessieay: this is all happening when an auction is being accepted. I think the assumption we are making is that someone cannot become sam-unverified, but they can go from small business to not small business between the time they bid and when they win.... perhaps we will want to re-visit that assumption at some point

Issue #1245, opened 2016-09-19
Show vendor GitHub name instead of email address


mtorres253:
When an auction is pending acceptance, or pending a payment URL, the admin should see the vendor's GitHub name in the status box rather than the vendor email. Right now it's showing the email which is  a bit of a confusing experience.

We agreed to show vendor name (if it exists) and fallback to GitHub name if not. The name/GitHub name should still link to that vendor's

PR #1258, opened 2016-09-20
Show winner name or github login, not email

jessieay:
* Closes https://github.com/18F/micropurchase/issues/1245

### DIFF
--- a/app/presenters/admin_auction_status_presenter/rejected.rb
+++ b/app/presenters/admin_auction_status_presenter/rejected.rb
@@ -8,7 +8,7 @@ class AdminAuctionStatusPresenter::Rejected < AdminAuctionStatusPresenter::Base
       'statuses.admin_auction_status_presenter.rejected.body',
       delivery_url: auction.delivery_url,
       rejected_at: rejected_at,
-      winner_email: winner.email
+      winner_name: winner_name
     )
   end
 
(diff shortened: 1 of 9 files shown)


app/presenters/admin_auction_status_presenter/rejected.rb
```diff
@@ -8,7 +8,7 @@ def body
       'statuses.admin_auction_status_presenter.rejected.body',
       delivery_url: auction.delivery_url,
       rejected_at: rejected_at,
-      winner_email: winner.email
+      winner_name: winner_name
```
harrisj: Nice, but why doesn't this use the `winner_url` method too? Should we fix that while we're doing this?

config/locales/statuses/en.yml
```diff
@@ -143,10 +143,10 @@ en:
           be paid until they’ve specified a valid payment URL.
       accepted:
         header: "Accepted"
-        body: "%{winner_email}'s delivered code was accepted on %{accepted_at}."
+        body: "%{winner_url}'s delivered code was accepted on %{accepted_at}."
       rejected:
         header: "Rejected"
-        body: "<a href=%{delivery_url}>%{winner_email}'s work</a> was rejected on %{rejected_at}."
+        body: "<a href=%{delivery_url}>%{winner_name}'s work</a> was rejected on %{rejected_at}."
```
harrisj: Oh, I guess that's why. I wonder if we should have winner_url and their work as two separate links but I guess that is a bit confusing

harrisj:
LGTM. Merging

PR #1259, opened 2016-09-20
Show information about bids for available auctions

### LINKED ISSUE #1235: Admin messaging - Auction show page - Auction is accepting bids

harrisj:
Fixes #1235 

18f_micro-purchase_-_universal_explicit_throughput

This adds a box like this to available auctions for the both the `/auctions/:id` and `/admin/auctions/:id` when there are bids for an auction and the user is an admin. Note that this deprecates the `BidStatusPresenter::Available::Admin` presenter and I should delete that before this is merged if we think this is a right step.

Also, the message does not include the amount of the winning bid. Should it?

### DIFF
--- a/app/models/admin_auction_status_presenter_factory.rb
+++ b/app/models/admin_auction_status_presenter_factory.rb
@@ -46,6 +46,8 @@ class AdminAuctionStatusPresenterFactory
       AdminAuctionStatusPresenter::Future
     elsif future? && auction.unpublished?
       AdminAuctionStatusPresenter::ReadyToPublish
+    elsif available?
+      AdminAuctionStatusPresenter::Available
     elsif work_in_progress?
       AdminAuctionStatusPresenter::WorkInProgress
     elsif !auction.pending_delivery?
--- a/app/models/admin_auction_status_presenter_factory.rb
+++ b/app/models/admin_auction_status_presenter_factory.rb
@@ -46,8 +46,6 @@ class AdminAuctionStatusPresenterFactory
       AdminAuctionStatusPresenter::Future
     elsif future? && auction.unpublished?
       AdminAuctionStatusPresenter::ReadyToPublish
-    elsif available?
-      AdminAuctionStatusPresenter::Available
     elsif work_in_progress?
       AdminAuctionStatusPresenter::WorkInProgress
     elsif !auction.pending_delivery?
--- a/app/presenters/admin_auction_status_presenter/available.rb
+++ b/app/presenters/admin_auction_status_presenter/available.rb
@@ -5,11 +5,12 @@ class AdminAuctionStatusPresenter::Available < AdminAuctionStatusPresenter::Base
 
   def body
     if total_bids > 0
-      I18n.t('statuses.bid_status_presenter.available.admin.has_bids',
-             end_date: end_date,
-             winner_url: winner_url,
-             total_bids: total_bids
-            )
+      I18n.t(
+        'statuses.bid_status_presenter.available.admin.has_bids',
+        end_date: end_date,
+        winner_url: winner_url,
+        total_bids: total_bids
+      )
     else
       I18n.t('statuses.bid_status_presenter.available.admin.body', end_date: end_date)
     end
@@ -25,8 +26,6 @@ class AdminAuctionStatusPresenter::Available < AdminAuctionStatusPresenter::Base
   def winner
     auction.lowest_bid.bidder
   end
-
-  private
 
   def end_date
     DcTimePresenter.convert_and_format(auction.ended_at)
--- a/app/presenters/admin_auction_status_presenter/available.rb
+++ b/app/presenters/admin_auction_status_presenter/available.rb
@@ -4,7 +4,26 @@ class AdminAuctionStatusPresenter::Available < AdminAuctionStatusPresenter::Base
   end
 
   def body
-    I18n.t('statuses.bid_status_presenter.available.admin.body', end_date: end_date)
+    if total_bids > 0
+      I18n.t('statuses.bid_status_presenter.available.admin.has_bids',
+             end_date: end_date,
+             winner_url: winner_url,
+             total_bids: total_bids
+            )
+    else
+      I18n.t('statuses.bid_status_presenter.available.admin.body', end_date: end_date)
+    end
+  end
+
+  private
+
+  def total_bids
+    auction.bids.count
+  end
+
+  # winning_bid is a NullBid when sealed-bid auctions are available
+  def winner
+    auction.lowest_bid.bidder
   end
 
   private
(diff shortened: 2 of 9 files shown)


app/models/admin_auction_status_presenter_factory.rb
```diff
@@ -46,6 +46,8 @@ def other_purchase_card_presenter
       AdminAuctionStatusPresenter::Future
     elsif future? && auction.unpublished?
       AdminAuctionStatusPresenter::ReadyToPublish
+    elsif available?
```
jessieay: seems like we are now checking for available twice in this conditional -- thoughts on whether we should just check once?

app/presenters/admin_auction_status_presenter/available.rb
```diff
@@ -4,7 +4,26 @@ def header
   end
 
   def body
-    I18n.t('statuses.bid_status_presenter.available.admin.body', end_date: end_date)
+    if total_bids > 0
```
jessieay: I would probably default to doing this check in out big giant conditional rather than here...but I can see why we'd want it here. I think that we need to work on breaking up that giant `if` statement so it is easier to read and understand. So having this here for now is easier to comprehend. In any ideal world, I think we'd keep all our conditionals outside of presenters

app/presenters/admin_auction_status_presenter/available.rb
```diff
@@ -4,7 +4,26 @@ def header
...
+    end
+  end
+
+  private
+
+  def total_bids
+    auction.bids.count
+  end
+
+  # winning_bid is a NullBid when sealed-bid auctions are available
```
jessieay: interesting -- maybe we should name this method differently or create a method somewhere else that clarifies this rather than needing a comment, which will potentially get out of date?

features/step_definitions/auction_create_steps.rb
```diff
@@ -88,10 +88,14 @@
   @auction.skills << skills
 end
 
-Given(/^there is a budget approved auction$/) do
+Given(/^there is a budget approved auction with bids$/) do
   @auction = FactoryGirl.create(:auction, :c2_approved, :with_bids)
 end
 
+Given(/^there is a budget approved auction$/) do
```
jessieay: maybe add `with no bids` to clarify?

features/support/env.rb
```diff
@@ -45,6 +45,18 @@ def start_date
   def winning_bid
     WinningBid.new(@auction).find
   end
+
+  def winner
+    winning_bid.bidder
+  end
+
+  def winner_url
```
jessieay: nice! I think if you rebase this is in `develop` now

app/presenters/admin_auction_status_presenter/available.rb
harrisj: Yeah, I assume you mean moving this into the `AdminAuctionStatusPresenterFactory` and `BidStatusPresenter` factories and have them look at the number of bids? I thought about that, but I kinda want to limit the factories to only needing to know about the status fields (I would probably merge the two `Future` presenters by this same logic.

app/presenters/admin_auction_status_presenter/available.rb
harrisj: Well, this lets me just use the same `winner_url` method, but yeah I understand it's a little unclear

jessieay: yeah I think what you are doing here makes total sense, I think naming the method something like `unmasked_winner_name_for_admin` would be an improvement over having a comment 💥

harrisj: So I would basically ignore and redefine winner_url as well then?

features/step_definitions/auction_create_steps.rb
harrisj: THis is an existing step I am reusing

features/support/env.rb
harrisj: Yep! Thanks!

app/models/admin_auction_status_presenter_factory.rb
harrisj: Good catch!

app/presenters/admin_auction_status_presenter/available.rb
jessieay: Yeah I think this is fine to leave as-is for now and then we can think about moving our crazy factory conditional into smaller parts and at the point we might want to surface this logic outside of this presenter.

app/presenters/admin_auction_status_presenter/available.rb
jessieay: yeah that might be clearer than having the comment here.

features/step_definitions/auction_create_steps.rb
jessieay: yeah but now that we have a similar step "with bids" perhaps renaming this would be a good idea?

app/presenters/admin_auction_status_presenter/available.rb
harrisj: Okay, fixed

features/step_definitions/auction_create_steps.rb
harrisj: Fixed

app/presenters/admin_auction_status_presenter/available.rb
```diff
@@ -4,11 +4,40 @@ def header
...
+      path_name: 'admin_user',
+      params: { id: unmasked_winner_for_admin.id }
+    )
+  end
+
+  def winner_name
+    unmasked_winner_for_admin.name || unmasked_winner_for_admin.github_login
+  end
+
+  def unmasked_winner_for_admin
```
jessieay: 😻

jessieay:
woot woot! merging once 💚

PR #1264, opened 2016-09-20
Set secret_key_base

jessieay:
* This will keep people logged into the app between deploys
* See https://gist.github.com/jessieay/25ac100206b9a40a191494fb4fc82d56


jessieay:
note: I already setup these user-provided services on staging and prod

harrisj:
OOOH 👍

Issue #1252, opened 2016-09-19
Tock band-aid


mtorres253:
Manual job to pull in projects from Tock so that we can create micro-purchases that are billed against the correct IAA and so our C2 integration continues to work

PR #1265, opened 2016-09-20
Tock band-aid

jessieay:
* Closes https://github.com/18F/micropurchase/issues/1252
* This update will allow us to ssh into staging and prod to update tock
  items in our DB
* Once new auth is implemented on Tock, we will add stuff to access
  auth-d API
* For now, will run this as a one-off task


harrisj:
LGTM

PR #1271, opened 2016-09-21
Rename AuctionStatus to BiddingStatus

harrisj:
Fixes part of #1247

This PR renames the AuctionStatus class to BiddingStatus. I still have to rename the StatusPresenter classes, but that should be a different PR

### DIFF
--- a/app/models/bid_status_presenter_factory.rb
+++ b/app/models/bid_status_presenter_factory.rb
@@ -100,23 +100,23 @@ class BidStatusPresenterFactory
   end
 
   def work_in_progress?
-    auction_status.work_in_progress?
+    bidding_status.work_in_progress?
   end
 
   def over?
-    auction_status.over?
+    bidding_status.over?
   end
 
   def available?
-    auction_status.available?
+    bidding_status.available?
   end
 
   def future?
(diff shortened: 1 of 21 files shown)


app/models/bid_status_presenter_factory.rb
```diff
@@ -100,23 +100,23 @@ def ineligible?
   end
 
   def work_in_progress?
```
jessieay: now that we call this `bidding_status`, `work_in_progress` no longer feels like it belongs here. thoughts?

harrisj: Yes, it was a bit of a bd match before. I wasn't sure if it should go in the Auction model or not since we have been trting to minimize what is in there. Thoughts?

jessieay: It almost seems like this should be an enum on `delivery_status` -- perhaps we can create an issue to do that l8r?

harrisj: Moved to Auction model for now. Thanks!

jessieay:
This looks great! will merge once 🍏

PR #1272, opened 2016-09-21
Rename auction status field to delivery_status

harrisj:
Fixes part of #1247

This is to make more clear how it is different from the bidding status (formerly known as AuctionStatus) and c2_status

### DIFF
--- a/app/models/admin_auction_status_presenter_factory.rb
+++ b/app/models/admin_auction_status_presenter_factory.rb
@@ -35,7 +35,7 @@ class AdminAuctionStatusPresenterFactory
     elsif auction.payment_confirmed?
       C2StatusPresenter::PaymentConfirmed
     elsif !auction.pending_delivery?
-      Object.const_get("AdminAuctionStatusPresenter::#{status}")
+      Object.const_get("AdminAuctionStatusPresenter::#{delivery_status}")
     else # auction.approved? && auction.pending_delivery?
       C2StatusPresenter::Approved
     end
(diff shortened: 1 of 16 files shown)


app/models/admin_auction_status_presenter_factory.rb
```diff
@@ -67,11 +67,11 @@ def work_in_progress?
...
-  def status
-    auction.status.camelize
+  def delivery_status
+    auction.delivery_status.camelize
   end
 
   def c2_status
     auction.c2_status.camelize
-  end
+ps  end
```
jessieay: typo?

spec/factories/auctions.rb
```diff
@@ -141,12 +141,12 @@
       end
     end
 
-   trait :pending_c2_approval do
-     future
-     c2_status :pending_approval
-     purchase_card :default
-     unpublished
-   end
+    trait :pending_c2_approval do
```
jessieay: I can't tell if this new indentation is good or bad?

app/models/admin_auction_status_presenter_factory.rb
harrisj: Yes fixed

spec/factories/auctions.rb
harrisj: Actually, the issue was the indenting was bad before and I fixed it

jessieay:
Perfecto -- mergin'! 😸

PR #1273, opened 2016-09-21
Rename status presenters to bidding_status presenters

### LINKED ISSUE #1247: Rename AuctionStatus
We have a `status` field on auctions, but we also have `AuctionStatus` which returns the temporal nature of an auction -- eg whether it is future, available, or closed. We should rename `AuctionStatus` to more clearly differentiate between the time issue and `status` field on auctions

harrisj:
**Merge after #1271**

This renames the StatusPresenter classes to be BiddingStatusPresenter (unfortunately, we now also have BidStatusPresenter too) as a followup to #1271.

Fixes #1247

### DIFF
--- a/app/models/bidding_status_presenter_factory.rb
+++ b/app/models/bidding_status_presenter_factory.rb
@@ -1,4 +1,4 @@
-class StatusPresenterFactory
+class BiddingStatusPresenterFactory
   attr_reader :auction
 
   def initialize(auction)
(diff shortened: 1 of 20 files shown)


jessieay:
Not seeing updates in `develop` showing up here -- perhaps will be fixed with a quick rebase?

harrisj:
Rebased!

jessieay:
Looks great! Mergin'

PR #1275, opened 2016-09-21
Only load live ClientAccounts when creating/editing an auction.

### LINKED ISSUE #1246: Only show live Tock projects in "Billable to" dropdown
So that I don't accidentally assign an expired project to a new auction, I should only see Tock projects that are "live" (that are on or after the project start date, and on or before the project end date) when I am creating an auction.

- [ ] **Scenario: Admin can only assign live Tock projects to new auction**
- And I'm creating an auction (https://micropurchase.18f.gov/admin/auctions/new)
- When I click the dropdown for "Billable to"
- Then I should only see Tock projects that are "live" (see above)

we should make sure that only active tock projects show up on dropdown, but we should not delete old tock projects since they might be associated with existing auctions (closed) and should still be displayed

adelevie:
closes #1246

### DIFF
--- a/app/services/tock_importer.rb
+++ b/app/services/tock_importer.rb
@@ -6,6 +6,7 @@ class TockImporter
       tock_project = ClientAccount.find_or_initialize_by(tock_id: project["id"])
       tock_project.name = project["name"]
       tock_project.billable = project["billable"]
+      tock_project.active = project["active"]
       tock_project.save!
     end
   end
--- a/db/schema.rb
+++ b/db/schema.rb
@@ -73,6 +73,7 @@ ActiveRecord::Schema.define(version: 20160921135321) do
     t.string   "name",       null: false
     t.boolean  "billable",   null: false
     t.integer  "tock_id",    null: false
+    t.boolean  "active"
   end
 
   add_index "client_accounts", ["tock_id"], name: "index_client_accounts_on_tock_id", unique: true, using: :btree
--- a/db/schema.rb
+++ b/db/schema.rb
@@ -68,12 +68,12 @@ ActiveRecord::Schema.define(version: 20160921135321) do
   add_index "bids", ["bidder_id"], name: "index_bids_on_bidder_id", using: :btree
 
   create_table "client_accounts", force: :cascade do |t|
-    t.datetime "created_at", null: false
-    t.datetime "updated_at", null: false
-    t.string   "name",       null: false
-    t.boolean  "billable",   null: false
-    t.integer  "tock_id",    null: false
-    t.boolean  "active"
+    t.datetime "created_at",                 null: false
+    t.datetime "updated_at",                 null: false
+    t.string   "name",                       null: false
+    t.boolean  "billable",                   null: false
+    t.integer  "tock_id",                    null: false
+    t.boolean  "active",     default: false
   end
 
   add_index "client_accounts", ["tock_id"], name: "index_client_accounts_on_tock_id", unique: true, using: :btree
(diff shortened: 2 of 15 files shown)


db/schema.rb
```diff
@@ -73,6 +72,7 @@
     t.string   "name",       null: false
     t.boolean  "billable",   null: false
     t.integer  "tock_id",    null: false
+    t.boolean  "live"
```
jessieay: default for this?

jessieay:
I feel like there should be some updates to the TockImporter that updates ClientAccounts' `live` status? Might be easier to just save `start_date` and `end_date` from Tock API instead and only show items that are "live" based on those dates rather than using a boolean

adelevie:
why is it better for micro-purchase to calculate "live"-ness if Tock already is doing that? Or could we have a scenario where live == false?

db/schema.rb
adelevie: I guess we could make it false? But we're getting this value from Tock, so I am unsure.

jessieay: Oh I didn't realize that. Where in this PR are we getting that from Tock?

app/services/tock_importer.rb
```diff
@@ -6,6 +6,7 @@ def perform
       tock_project = ClientAccount.find_or_initialize_by(tock_id: project["id"])
       tock_project.name = project["name"]
       tock_project.billable = project["billable"]
+      tock_project.live = project["live"]
```
jessieay: oh here it is! not sure how I missed that before

db/schema.rb
jessieay: yeah let's default to `false` and then let tock tell us when things are live

app/services/tock_importer.rb
jessieay: looks like tock calls this `"active"` -- should we do same?

jessieay: spec for this importer update?

jessieay:
This is looking super close, @adelevie !

app/services/tock_importer.rb
adelevie: more than this? https://github.com/18F/micropurchase/blob/cea84c1dad7f1c11006005bae860f22f7c740065/spec/services/tock_importer_spec.rb

spec/support/fixtures/tock_projects.json
```diff
@@ -1 +1,37 @@
...
+    },
+    {
+      "id": 96,
+      "client": "General Services Administration - 18F (Non-Billable)",
+      "name": "18F Guides",
+      "description": " Aims to disseminate software development and bureaucracy-hacking best practices across 18F and for our partners' digital service teams",
+      "billable": false,
+      "start_date": null,
+      "end_date": null,
+      "active": false
```
jessieay: this field is not in the real API -- is it coming soon? https://tock.18f.gov/api/projects.json

app/services/tock_importer.rb
jessieay: I think that rather than `respond_to` I'd make sure that `ClientAccount` records are being set to `active` properly

jessieay: eg: look for a specific client account that should be active according to fixtures and make sure that on our side it is `active` too

spec/support/fixtures/tock_projects.json
adelevie: Working on everything else until I get a new API key to Tock to see what's actually in the API.

adelevie: Confirmed that `active` is in the production API:
```json
		{
			"id": 120,
			"client": "General Services Administration - 18F (Non-Billable)",
			"name": "18F Branding",
			"description": "Working on branding for 18F—internal branding",
			"billable": false,
			"start_date": null,
			"end_date": null,
		}
```

adelevie:
The production API now has token-based auth. I'll set up a user-provided service to persist the API key and update `TockImporter` to use the key.

jessieay:
Hey @adelevie so do you want to merge this as-is since it covers #1246 and then work on token stuff separately? or do it here? getting token auth working has its own card, which I am happy for you to take over if you have time: https://github.com/18F/micropurchase/issues/951

Issue #1107, opened 2016-08-24
List Font Weights are different than normal p weights


lauraGgit:
in auctions, lists do not have bullets so it looks odd that they are a different font weight than the text that is in a p tag.

andrewmaier:
Hey @lauraGgit this is something we're addressing in [the updated design](https://pages.18f.gov/micropurchase/auctions/show/?auctionWorkflowState=coming-soon-vendors) — see [this](https://pages.18f.gov/micropurchase/docs/getting-started/) content-heavy page, for example. It's just going to take us some time to migrate these over to production :)

mtorres253:
Let's try updating the USWDS so that we can fix this bug.

jessieay:
WIP here https://github.com/18F/micropurchase/compare/jy-test-uswds-fonts?expand=1

PR #1277, opened 2016-09-22
Upgrade to latest USWDS

jessieay:
* Use the gem! https://github.com/18F/uswds-rails-gem
* Biggest changes needed to upgrade were related to list and link styles
* USWDS now defaults to underline style links, which we want only on hover
* They also used to default to no li style on unodered lists
* Closes https://github.com/18F/micropurchase/issues/1107

## Header before:

screen shot 2016-09-22 at 10 34 44 am

## Header after: (slightly darker, color taken from 18f.gsa.gov)

screen shot 2016-09-22 at 10 34 49 am

## List in auction body before:

screen shot 2016-09-22 at 11 05 43 am

## List in auction body after:

screen shot 2016-09-22 at 11 04 24 am


harrisj:
I APPROVE

Issue #1270, opened 2016-09-20
Footer is back in admin view


mtorres253:
This story is just to bring the footer back to the admin view so I don't need to go to the vendor view to access it!

PR #1278, opened 2016-09-22
Add footer to admin views

jessieay:
* Also re-add font-awesome css
  I removed it a while ago thinking it was for fonts, but it was for
  icons like the github icon
* Closes https://github.com/18F/micropurchase/issues/1270

### DIFF
--- a/app/views/layouts/admin.html.erb
+++ b/app/views/layouts/admin.html.erb
@@ -3,23 +3,26 @@
 <head>
   <meta charset="utf-8">
   <meta name="ROBOTS" content="NOODP" />
-  <%= stylesheet_link_tag    'application', media: 'all', debug: true %>
+  <link rel="stylesheet" href="https://maxcdn.bootstrapcdn.com/font-awesome/4.5.0/css/font-awesome.min.css">
+  <%= stylesheet_link_tag 'application', media: 'all', debug: true %>
   <%= jquery_include_tag :google %>
   <%= javascript_include_tag "application", debug: true %>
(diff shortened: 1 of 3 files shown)


app/views/layouts/admin.html.erb
```diff
@@ -3,23 +3,26 @@
 <head>
   <meta charset="utf-8">
   <meta name="ROBOTS" content="NOODP" />
-  <%= stylesheet_link_tag    'application', media: 'all', debug: true %>
+  <link rel="stylesheet" href="https://maxcdn.bootstrapcdn.com/font-awesome/4.5.0/css/font-awesome.min.css">
```
harrisj: Not really blocking this, but I wonder if we have any policy about using third-party CDNs or not

config/initializers/assets.rb
```diff
@@ -1,3 +1,3 @@
 Rails.application.config.assets.version = (ENV["ASSETS_VERSION"] || "1.0")
 Rails.application.config.assets.precompile += %w( jquery.js )
-Rails.application.config.assets.precompile += %w( uswds/us_flag_small )
+Rails.application.config.assets.precompile += %w( uswds/us_flag_small.png )
```
harrisj: Can you fix this Rubocop issue?

harrisj: Eh, never mind

app/views/layouts/admin.html.erb
jessieay: goooood question! we are using a jquery cdn too! (google)

PR #1283, opened 2016-09-23
Modify admin accepted text to link to C2 URL

### LINKED ISSUE #1197: Admin messsaging - Auction show page - Payment needed using 18F p-card
Given I am an admin managing auctions, if an auction has been accepted, and 18F needs to pay for it, I should see the messaging reflecting that status in on the auction show page

- [ ] **Scenario: Admin sees that accepted auction requires payment via 18F p-card**
- And there is an auction using the default 18F p-card
- And the auction has been accepted
- And there is a payment URL for the winning vendor on file
- When I navigate to the show page for that auction

> Pending payment from 18F
> The p-card holder has been asked (via C2) to remit payment. 

- And "C2" should link to the C2 proposal URL for that auction

harrisj:
Fixes #1197 

We decided to just edit the current message that admins see for accepted auctions to also include the C2 URL. Trying this

localhost_3000_admin_auctions_12

### DIFF
--- a/app/presenters/admin_auction_status_presenter/accepted.rb
+++ b/app/presenters/admin_auction_status_presenter/accepted.rb
@@ -7,11 +7,16 @@ class AdminAuctionStatusPresenter::Accepted < AdminAuctionStatusPresenter::Base
     I18n.t(
       'statuses.admin_auction_status_presenter.accepted.body',
       accepted_at: accepted_at,
-      winner_url: winner_url
+      winner_url: winner_url,
+      c2_url: c2_url
     )
   end
 
   private
+
+  def c2_url
+    auction.c2_proposal_url
+  end
 
   def accepted_at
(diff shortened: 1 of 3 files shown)


jessieay:
Hello! Now that I am seeing this, I am worried that I may have led you astray. This presenter is used for BOTH pcards, but this copy would not make sense in the context of the "other" pcard. See: https://github.com/18F/micropurchase/blob/develop/app/models/admin_auction_status_presenter_factory.rb#L51

harrisj:
Can we merge this? I know it doesn't quite make sense in the context of other purchase cards, but that will be fixed in a future PR

jessieay:
sure! but can we fix ci first?

harrisj:
Argh, it worked locally after a rebase. Let me see what is going on

PR #1284, opened 2016-09-23
Payment needed for Other P-Card customers

### LINKED ISSUE #1197: Admin messsaging - Auction show page - Payment needed using 18F p-card

### LINKED ISSUE #1198: Admin messsaging - Auction show page - Payment needed using "other" p-card
Given I am an admin managing auctions, if an auction has been accepted, and payment needs to happen on another p-card (partner agency's p-card for example), I should see the messaging reflecting that status in on the auction show page

- [x] **Scenario: Admin sees that accepted auction requires payment via other p-card (not 18F)**
- And there is an auction using an other p-card (other p-card payment method) 
- And the auction has been accepted
- And there is a payment URL for the winning vendor on file
- When I navigate to the show page for that auction

> Pending payment from customer
> [auction.customer] has been asked to remit payment. You must manually mark this auction as paid when it is paid by the customer.

- And I should see a button labeled "Mark as paid"

screen shot 2016-09-19 at 1 30 17 pm

- [ ] **Scenario: Admin marks an auction as paid on the auction page**
- When I click on the Mark as Paid button
- Then the auction should be marked as paid

harrisj:
Fixes #1198 

But I need to do some additional work before this is ready to be merged:
- [x] MarkPaidAuction should generate an email to the vendor like CheckPayment does
- [x] Need status message for when paid_at is not nil, can finish second feature then

### DIFF
--- a/app/controllers/admin/auction_mark_payments_controller.rb
+++ b/app/controllers/admin/auction_mark_payments_controller.rb
@@ -2,11 +2,12 @@ class Admin::AuctionMarkPaymentsController < Admin::BaseController
   def update
     @auction = Auction.find(params[:id])
 
-    paid_auction = MarkPaidAuction.new(
+    paid_auction = MarkOtherPcardAuctionAsPaid.new(
       auction: @auction
     )
 
     if paid_auction.perform
+      flash[:success] = I18n.t('controllers.admin.auction_mark_payments.update.success')
       @auction.save!
     else
       error_messages = @auction.errors.full_messages.to_sentence
--- a/app/controllers/admin/auction_mark_payments_controller.rb
+++ b/app/controllers/admin/auction_mark_payments_controller.rb
@@ -1,0 +1,18 @@
+class Admin::AuctionMarkPaymentsController < Admin::BaseController
+  def update
+    @auction = Auction.find(params[:id])
+
+    paid_auction = MarkPaidAuction.new(
+      auction: @auction
+    )
+
+    if paid_auction.perform
+      @auction.save!
+    else
+      error_messages = @auction.errors.full_messages.to_sentence
+      flash[:error] = error_messages
+    end
+
+    redirect_to admin_auction_path(@auction)
+  end
+end
--- a/app/mailers/auction_mailer.rb
+++ b/app/mailers/auction_mailer.rb
@@ -47,4 +47,19 @@ class AuctionMailer < ActionMailer::Base
       reply_to: 'micropurchase@gsa.gov'
     )
   end
+
+  def auction_paid_winning_vendor_other_pcard(auction:)
+    @auction = auction
+    @winning_bid = WinningBid.new(@auction).find
+
+    mail(
+      to: @winning_bid.bidder.email,
+      subject: I18n.t(
+        'mailers.auction_mailer.auction_paid_winning_vendor_other_pcard.subject',
+        auction_title: @auction.title
+      ),
+      from: SMTPCredentials.default_from,
+      reply_to: 'micropurchase@gsa.gov'
+    )
+  end
 end
--- a/app/models/admin_auction_status_presenter_factory.rb
+++ b/app/models/admin_auction_status_presenter_factory.rb
@@ -61,9 +61,9 @@ class AdminAuctionStatusPresenterFactory
     elsif auction.accepted_pending_payment_url?
       AdminAuctionStatusPresenter::AcceptedPendingPaymentUrl
     elsif auction.accepted? && auction.paid_at.nil?
-      AdminAuctionStatusPresenter::AcceptedOtherPcard
+      AdminAuctionStatusPresenter::OtherPcard::Accepted
     elsif auction.accepted? && auction.paid_at.present?
-      AdminAuctionStatusPresenter::PaidOtherPcard
+      AdminAuctionStatusPresenter::OtherPcard::Paid
     else # auction.rejected?
       AdminAuctionStatusPresenter::Rejected
     end
--- a/app/models/admin_auction_status_presenter_factory.rb
+++ b/app/models/admin_auction_status_presenter_factory.rb
@@ -60,8 +60,10 @@ class AdminAuctionStatusPresenterFactory
       AdminAuctionStatusPresenter::PendingAcceptance
     elsif auction.accepted_pending_payment_url?
       AdminAuctionStatusPresenter::AcceptedPendingPaymentUrl
-    elsif auction.accepted?
+    elsif auction.accepted? && auction.paid_at.nil?
       AdminAuctionStatusPresenter::AcceptedOtherPcard
+    elsif auction.accepted? && auction.paid_at.present?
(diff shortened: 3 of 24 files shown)


jessieay:
Oh i think this is for #1198  rather than #1197 right?

harrisj:
So, I need to know what should happen when the user says mark paid. Since it was requested in the original feature that this should email the user to confirm payment, it seems like we need to support two more states (the user has been emailed, the user has confirmed) down the road.

There are a few options for what we could do:
1. Have the mark_paid set `c2_paid` and reuse the existing path for user C2 confirmations via email. This is a little awkward since this isn't a C2 auction otherwise, but it means not having to duplicate functionality.
2. Create a separate path that replicates these two options. This would probably entail turning `c2_status` into a generic `payment_status` and figuring out if we could just add a few fields for C2-specific data rather than a separate state machine
3. Have a separate email to the customer that doesn't involve a click confirmation but tells them we are marking this as paid and to just get in touch if it's not. Also add a message for this situation where `pcard: :other, paid_at: not nil`

I am inclined to just do the third option for now since it is the easiest, but it means I will need the following bit of information to proceed:

- [ ] the text of the email to send to the vendor
- [ ] the text of the status to display when paid_at is true

Thoughts?

mtorres253:
I think #3 is fine for now because we don't have to do pegasus reporting on payments involving other p-cards (correct @adelevie?)

mtorres253:
For the language in the status box, I would just update to the following:

> Paid
[auction.winner.name] was paid $[amount] on [date].

harrisj:
Thank you. And did you have anything you want the email to say differently?

harrisj:
I put in my own text for the email based on the other email text. Let me know if not appropriate

mtorres253:
oh man! sorry @harrisj my day has gotten away from me. Looks good 👍

harrisj:
Oops, had a bad test. This is ready to go now

app/controllers/admin/auction_mark_payments_controller.rb
```diff
@@ -0,0 +1,18 @@
+class Admin::AuctionMarkPaymentsController < Admin::BaseController
+  def update
+    @auction = Auction.find(params[:id])
+
+    paid_auction = MarkPaidAuction.new(
```
jessieay: I am not opposed to having this service class, but since it does so little and we are just using it in one place, thoughts on marking as paid and sending the email in this controller?

harrisj: I hear you, but this keeps it consistent with the other actions and it lets me test the functionality with a unit test

app/controllers/admin/auction_mark_payments_controller.rb
```diff
@@ -0,0 +1,18 @@
...
+      auction: @auction
+    )
+
+    if paid_auction.perform
+      @auction.save!
```
harrisj: I guess we could do a flash!

app/mailers/auction_mailer.rb
```diff
@@ -47,4 +47,19 @@ def auction_paid_winning_vendor_notification(auction:)
       reply_to: 'micropurchase@gsa.gov'
     )
   end
+
+  def auction_paid_winning_vendor_other_pcard(auction:)
```
harrisj: Well, I was just matching the other mailer above it. I'm not sure why some of these are here vs. the other one.

app/models/admin_auction_status_presenter_factory.rb
```diff
@@ -60,8 +60,10 @@ def other_purchase_card_presenter
       AdminAuctionStatusPresenter::PendingAcceptance
     elsif auction.accepted_pending_payment_url?
       AdminAuctionStatusPresenter::AcceptedPendingPaymentUrl
-    elsif auction.accepted?
-      AdminAuctionStatusPresenter::Accepted
+    elsif auction.accepted? && auction.paid_at.nil?
+      AdminAuctionStatusPresenter::AcceptedOtherPcard
```
harrisj: How do we want to devise the namespaces though?

app/mailers/auction_mailer.rb
jessieay: The mailers that go to winners should all be moved to `WinnerMailer`, eventually. It was created after some of these mailers were, so not everything is over there yet

app/models/admin_auction_status_presenter_factory.rb
jessieay: `AdminAuctionStatusPresenter::OtherPurchaseCard::Accepted` ?

app/controllers/admin/auction_mark_payments_controller.rb
jessieay: ok cool, then let's just make sure we rename it so it is obvious that this is how we mark "Other P Card" auctions as paid, not all auctions

app/mailers/auction_mailer.rb
harrisj: I'd like to hold off on this one for a future PR if that's okay

app/services/mark_paid_auction.rb
```diff
@@ -0,0 +1,12 @@
+class MarkPaidAuction
+  attr_reader :auction
+
+  def initialize(auction:)
+    @auction = auction
+  end
+
+  def perform
+    auction.update(paid_at: Time.now)
+    AuctionMailer.auction_paid_winning_vendor_other_pcard(auction: auction).deliver_later
```
harrisj: Renamed

app/mailers/auction_mailer.rb
jessieay: OK I added an issue to next up for this: https://github.com/18F/micropurchase/issues/1316

app/models/admin_auction_status_presenter_factory.rb
```diff
@@ -60,8 +60,10 @@ def other_purchase_card_presenter
       AdminAuctionStatusPresenter::PendingAcceptance
     elsif auction.accepted_pending_payment_url?
       AdminAuctionStatusPresenter::AcceptedPendingPaymentUrl
-    elsif auction.accepted?
-      AdminAuctionStatusPresenter::Accepted
+    elsif auction.accepted? && auction.paid_at.nil?
+      AdminAuctionStatusPresenter::OtherPcard::Accepted
```
jessieay: Nice! added an issue for also namespacing the other presenters https://github.com/18F/micropurchase/issues/1317

jessieay:
Looks great! Mergin' 🌈 🌈

PR #1286, opened 2016-09-23
Update TockImporter to set API token in header

jessieay:
* Closes https://github.com/18F/micropurchase/issues/951 (!!)
* Tested this locally and it works!


adelevie:
tiny code climate nitpick

jessieay:
@adelevie fixed! :)

PR #1294, opened 2016-09-26
Add test for `dev:prime` rake task

jessieay:
* So we know when `bin/setup` is broken


adelevie:
Looks good ... is it ok in this instance for the test to lack an assertion? I know rspec doesn't like it when you assert that an error isn't raised, so I can see why assertions might be tough here.

jessieay:
When the task is invoked, it runs. All I am concerned with here is that it runs successfully. If you invoke the task when an invalid trait, for example (the state the task was in before this PR), the spec will raise.

adelevie:
Sounds good to me!

Issue #1165, opened 2016-09-06
Admin sets auction dates when publishing an auction on the show page


mtorres253:
As an admin, when I'm ready to publish an auction from the auction show page, I want to be able to quickly update the auction dates without having to go into the edit form.

- [x] **Scenario: Admin sees form to set auction dates in status box of show page**
- Given I am an authenticated admin user
- And there is an auction that is in an unpublished state
- When I am on the show page for that auction (micropurchase.18f.gov/admin/auctions/show)
- Then in the status box, I should see the following message

> Ready to publish
> This auction is ready to publish! Please select a start date, end date, and delivery period for this auction.

- And I should see inputs for start date/time, end date/time and days to delivery
- And if these dates are already assigned, I should see them pre-populated in the inputs

![screen shot 2016-09-06 at 4 43 44 pm](https://cloud.githubusercontent.com/assets/100149/18294836/1d34a6fa-7451-11e6-9eae-c2e9011a04f3.png)

- [x] **Scenario: Admin sets auction dates of an unpublished auction on the auction show page and publishes auction**
- And there are values for the start date, end date and delivery deadline
- When I click the publish button
- And there are no errors
- Then the auction should be published
- And the start date, end date and delivery deadline should be set to what I specified
- And the page should reload
- And I should see the "Coming Soon" messaging for that auction (#931, #1133)

jessieay:
can we remove `published` field from the form once this is added? I ask because I am going to build functionality around this button and want to know if I need to build it for the form, too

andrewmaier:
@jessieay I would _personally_ prefer we remove the published field from the form — that feels superfluous and it's accounted for in the existing prototype/vignettes. 

@jessieay @mtorres253 For what it's worth, I've made implementation suggestions around javascript support during publishing in [this vignette](https://docs.google.com/document/d/1lWPYtQLdyZ42ENPIMxITcAuALXr2ADswu0XUCB7hYb4/edit#heading=h.2mc12waszjsa).

PR #1296, opened 2016-09-26
Admin can set dates when publishing auction

jessieay:
* Add fields on blue status box
* Remove publish fields from auction form
* Final cleanup will be to move from date select to day count select for
  delivery deadline
* Closes https://github.com/18F/micropurchase/issues/1165

screen shot 2016-09-26 at 12 00 23 pm

screen shot 2016-09-26 at 12 00 28 pm


adelevie:
💯

PR #1302, opened 2016-09-26
Put a sequence on the client_account name

harrisj:
There was an issue where the tests were randomly using the same name for active and inactive client accounts leading to a false positive for some tests that was randomly triggered.


adelevie:
LGTM

PR #1310, opened 2016-09-28
Admin can now mark finished an auction that is a work in progress

### LINKED ISSUE #1299: Admin view - Status box on auction show page - Finish work
As an admin, I want to be able to mark an auction as done if a vendor forgets to do that themselves, so there should be a way for me to do this in the UI

- [x] **Scenario: Admin sees "Mark as delivered" button in status box of auction show page"
- And there is a published auction that has a winner
- And there is a delivery URL specified
- And the vendor has not yet clicked the "I'm done" button
- When I go to the auction show page for that auction
- Then I should see the following status in the status box

> Work in progress
> [auction.winner.name] initiated a pull request on [auction.startWorkDate], and should complete work no later than [deadline].

- And I should see a "Mark as delivered" button

screen shot 2016-09-26 at 3 00 42 pm

- [x] **Scenario: Admin clicks "Mark as delivered" button and delivers auction"
- When I click on the "Mark as delivered" button
- Then the auction should be marked as done
- And the vendor should no longer be able to mark it as done

harrisj:
Fixes #1299

### DIFF
--- a/app/models/auction_parser.rb
+++ b/app/models/auction_parser.rb
@@ -29,7 +29,7 @@ class AuctionParser
       :issue_url,
       :notes,
       :purchase_card,
-      :status,
+      :delivery_status,
       :start_price,
       :summary,
       :title,
--- a/app/views/admin/auctions/_work_in_progress.html.erb
+++ b/app/views/admin/auctions/_work_in_progress.html.erb
@@ -1,4 +1,4 @@
 <%= link_to I18n.t('statuses.admin_auction_status_presenter.work_in_progress.actions.mark_delivered'),
-    auction_path(status.auction, auction: { status: 'pending_acceptance' }),
+    admin_auction_path(status.auction, auction: { delivery_status: 'pending_acceptance' }),
     method: :patch,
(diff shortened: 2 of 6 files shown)


app/models/auction_parser.rb
```diff
@@ -29,7 +29,7 @@ def auction_params
       :issue_url,
       :notes,
       :purchase_card,
-      :status,
+      :delivery_status,
```
jessieay: niiiice

app/views/admin/auctions/_work_in_progress.html.erb
```diff
@@ -0,0 +1,4 @@
+<%= link_to I18n.t('statuses.admin_auction_status_presenter.work_in_progress.actions.mark_delivered'),
+    admin_auction_path(status.auction, auction: { delivery_status: 'pending_acceptance' }),
```
jessieay: I like that you are re-using the default controller. I feel like we should also do this for some of the other update links that are updating a single attr (right now we have custom controllers for some of those)

PR #1311, opened 2016-09-28
Update Tock importer and fixtures to reflect API changes

jessieay:
* Tock changed their API
* This does not explain the 401 errors we are seeing on prod but are def
  a problem

### DIFF
--- a/app/services/tock_importer.rb
+++ b/app/services/tock_importer.rb
@@ -14,10 +14,6 @@ class TockImporter
   private
 
   def projects
-    @_projects ||= json["results"]
-  end
-
-  def json
     JSON.parse(data)
   end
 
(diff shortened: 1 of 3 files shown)


app/services/tock_importer.rb
```diff
@@ -14,10 +14,6 @@ def perform
   private
 
   def projects
-    @_projects ||= json["results"]
```
harrisj: So, why is this code being eliminated? It seems a little suspicious

jessieay: the API changed. If you look at the updates to the fixtures, the API now returns an array rather than a JSON object with a `results` key

harrisj: Got it

harrisj:
LGTM

PR #1313, opened 2016-09-28
Add regression test: rejected auction has no winning bidder

jessieay:
* We have an auction on staging like this
* Shouldn't happen in real life too much, but better to return this than
  to have the page error when an auction is in this state


harrisj:
Weird, but LGTM

PR #1314, opened 2016-09-28
Email admins when vendor starts work on auction

### LINKED ISSUE #1263: Admin Email Notification - Winning vendor has started work
As an admin of micropurchase, so that I don't have to go to the show page to check whether a winning vendor has started work, I want to also receive an email letting me know. This email should go to as many admins as possible.

- [x] **Scenario: System sends email notification to micropurchase admins that vendor has started work**
- Given there is a published auction
- And that auction has ended
- And that auction has a winner
- And the winner has started work [(entered a delivery URL and clicked "Start work")](https://github.com/18F/micropurchase/issues/924)
- Then the system should send an automated email to the all micropurchase admins 
- And the email content should be as follows

> From: 		micropurchase@gsa.gov
> To address: 	[micropurchase admins email address]
> Subject: 		Work started on auction “[auction.name]”.
> [auction.winner.name] initiated a pull request on [auction.startWorkDate] for “[auction.title]” and should complete their work no later than [auction.deliveryDeadline]. 
> You can view this auction (including a link to [auction.winner.name]’s pull request) by signing in and visiting: [auction url]

- And [auction.winner.name] should be the name of the winning vendor (or the github name, if there is no name for that vendor)
- And [auction url] should link to the admin show page for that auction - micropurchase.18f.gov/admin/auctions/[id]

harrisj:
Fixes #1263

### DIFF
--- a/app/mailers/admin_mailer.rb
+++ b/app/mailers/admin_mailer.rb
@@ -1,0 +1,17 @@
+class AdminMailer < ActionMailer::Base
+  ADMIN_EMAIL_ADDRESS = 'micropurchase@gsa.gov'.freeze
+
+  def vendor_started_work(auction:)
+    @auction = auction
+    winning_bid = WinningBid.new(@auction).find
+    winner = winning_bid.bidder
+    @winner_name = winner.name || winner.github_login
+
+    mail(
+      to: ADMIN_EMAIL_ADDRESS,
+      subject: I18n.t('mailers.admin_mailer.vendor_started_work.subject'),
+      from: SMTPCredentials.default_from,
+      reply_to: 'micropurchase@gsa.gov'
+    )
+  end
+end
--- a/app/services/winning_vendor_update_auction.rb
+++ b/app/services/winning_vendor_update_auction.rb
@@ -11,10 +11,12 @@ class WinningVendorUpdateAuction
     if user_is_winning_bidder? && status.present?
       auction.update(delivery_status: :pending_acceptance)
     elsif user_is_winning_bidder? && delivery_url.present?
-      auction.update(
+      update = auction.update(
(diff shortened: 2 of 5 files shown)


app/mailers/admin_mailer.rb
```diff
@@ -0,0 +1,17 @@
+class AdminMailer < ActionMailer::Base
+  ADMIN_EMAIL_ADDRESS = 'micropurchase@gsa.gov'.freeze
```
jessieay: nice! we should potentially use this constant in some of our other mailers, too, since I know we set this is the `reply-to` for all emails!

app/services/winning_vendor_update_auction.rb
```diff
@@ -11,10 +11,12 @@ def perform
     if user_is_winning_bidder? && status.present?
       auction.update(delivery_status: :pending_acceptance)
     elsif user_is_winning_bidder? && delivery_url.present?
-      auction.update(
+      update = auction.update(
         delivery_url: delivery_url,
         delivery_status: :work_in_progress
       )
+      AdminMailer.vendor_started_work(auction: auction).deliver_later
+      update
```
jessieay: are we returning this bc we want it to return `true`? if so, what about returning `true` here instead?

harrisj: Well, I think there are some cases where update might return false?

jessieay: good point -- what about putting the update below the mailer, then. since we are sending the email whether the update succeeds or not, seems like it would be OK to send the email then update the auction

harrisj: I guess I could do that

Issue #1032, opened 2016-08-10
Auction show page - style clean-up


mtorres253:
Let's do a few things to clean up the auction show page styles

this is for both micropurchase.18f.gov/auctions/[id]

and for  micropurchase.18f.gov/admin/auctions/[id]

- Replace ribbon with status label

- The status label colors should be the same as the colors on the auction preview tiles
![Uploading Screen Shot 2016-09-02 at 3.21.51 PM.png…]()

- Move edit button to top right corner (admins only)

- Get the layout as close as you can to the layout in the prototype

- Try to match font sizes, line spacing and colors

- Move from button nav to tab nav

https://pages.18f.gov/micropurchase/auctions/show/?auctionWorkflowState=draft-vendors

jessieay:
moving this to backlog until there are more details

PR #1315, opened 2016-09-28
Use tab subnav on auction#show

jessieay:
* Move from buttons to tabs
* This new layout requires that the header wrapper be outside of
  `usa-grid`, so had to move that outside of the layout and add to
  individual view templates
* Part of https://github.com/18F/micropurchase/issues/1032

## Before

screen shot 2016-09-28 at 11 50 27 am

## After
screen shot 2016-09-28 at 11 49 31 am


jessieay:
This will be a lot easier to review sans whitespace changes. You can do that via this URL: https://github.com/18F/micropurchase/pull/1315/files?w=1

harrisj:
Looks good to me!

Issue #1318, opened 2016-09-28
Fix time-related intermittenty failing test


jessieay:
```
(::) failed steps (::)
expected to find text "Estimated delivery date October 12, 2016 01:00:00 PM EDT" in "Estimated delivery date October 11, 2016 01:00:00 PM EDT" (RSpec::Expectations::ExpectationNotMetError)
./features/step_definitions/admin_auction_form_steps.rb:89:in `block (2 levels) in <top (required)>'
./features/step_definitions/admin_auction_form_steps.rb:86:in `/^I should see the updated estimate for the delivery deadline$/'
features/admin_creates_auction.feature:31:in `Then I should see the updated estimate for the delivery deadline'
Failing Scenarios:
cucumber features/admin_creates_auction.feature:22 # Scenario: Adding an auction with a non-default delivery timeline
```

 https://travis-ci.org/18F/micropurchase/builds/163524948

harrisj:
So, this is a weird one. The issue seems to be something to do with thread variables in Ruby, perhaps exacerbated by Capybara. Let me explain.

We are using the [business_time](https://github.com/bokmann/business_time) gem to do calculations like 5 business days from now. Since we want to not count Federal holidays, the gem allows us to use an initializer to load in the holidays we want to ignore (those are not counted as working days).

``` ruby
Holidays.between(Date.today, 2.years.from_now, :us, :observed).each do
  |holiday| BusinessTime::Config.holidays << holiday[:date]
end
```

However, whenever we get close to a holiday, the business_time gem sometimes seems to treat the holiday as a working day, causing a test to fail. This effect is seemingly random, which is _why_ the test fails, since the controller method we are testing sometimes doesn't treat the holiday as such, but the test assertion we compute independently does.

``` ruby
delivery_date = 5.business_days.after(@end_time)
formatted_delivery_date = DcTimePresenter.format(delivery_date)
 expect(page).to have_content(
    "Estimated delivery date #{formatted_delivery_date}"
  )
```

This is from the failing Cucumber step. This feature is testing the code that shows an "estimated delivery deadline" for an auction on the edit form. This code calls the [business_days_controller](https://github.com/18F/micropurchase/blob/develop/app/controllers/api/v0/business_days_controller.rb) backend with a date and time like `'2016-11-05'`, `'1:00 PM'` and the controller returns a JSON string with the estimated deadline.

[...]

PR #1319, opened 2016-09-28
Fix intermittently failing test

jessieay:
* Closes https://github.com/18F/micropurchase/issues/1318


jessieay:
Shoot, I now see that this did not fix it permanently. `develop` is also passing now (3:23pm PT). An hour ago, it was failing on `develop`

PR #1321, opened 2016-09-28
Fix up header styles for admin views

jessieay:
* Closes https://github.com/18F/micropurchase/issues/1032 at long last

Before:

screen shot 2016-09-28 at 4 33 54 pm

After:

screen shot 2016-09-28 at 4 33 06 pm


harrisj:
Nice!

PR #1324, opened 2016-09-29
Add an archived status to auction model

### LINKED ISSUE #1276: Admin can archive auction
So I have a situation where I have a draft auction that has C2 approval, but is not going forward. I have another situation where we will eventually go forward with the auction but the current C2 approval can't be used because the IAA expires this week and we don't have all the details yet for the auction.

As an admin user, I want to archive draft auctions that aren't going forward so they don't remain in a draft state forever.

**Context**
- And there is an unpublished auction

- [x] **Scenario: Admin archives draft unpublished auction with no bids**
- When I go to the auction page
- Then I should see a way to archive that auction

- [x] **Scenario: Admin cannot archive published auctions**
- When I go to the auction page
- Then I should NOT see a way to archive that auction

- [x] **Scenario: Admin sees confirmation before archiving an auction**
- And I am on the edit page for that auction
- When I click on the "Archive auction" link
- Then I should see a confirmation with the following message

> Are you sure you want to archive this auction?

- And I should see a "Confirm" button
- And I should see a "Cancel" button

- [x] **Scenario: Admin confirms that auction should be archived**
- When I click on the "Confirm" button
- Then the auction should be archived
- And I should navigate to the auction show page (micropurchase.18f.gov/admin/auctions/[id])
- And the status box copy should read:

> Archived
> This auction was archived on [date]

- [x] **Scenario: Admin cancels out of archive confirmation**
- When I click on the "Cancel" button
- Then the confirmation dialog should disappear
- And I should remain on the edit auction page

## Notes
Archive button is on the edit form. Easy way to implement this would be to make the archive a soft delete using something like the paranoia gem.

harrisj:
Fixes #1276

localhost_3000_admin_auctions_1_edit

Note: this does not implement a few tweaks to the admin display for archived auctions or data insights that are specified in later issues. Also, not sure how you want that styled

### DIFF
--- a/app/assets/javascripts/admin/archiveConfirm.js
+++ b/app/assets/javascripts/admin/archiveConfirm.js
@@ -1,0 +1,5 @@
+$(document).ready(function() {
+  $('#archive-auction-link').click(function() {
+    return confirm("Are you sure you want to archive this auction?");
+  })
+});
--- a/app/models/auction.rb
+++ b/app/models/auction.rb
@@ -30,7 +30,7 @@ class Auction < ActiveRecord::Base
     rejected: 2
   }
 
-  enum published: { unpublished: 0, published: 1 }
+  enum published: { unpublished: 0, published: 1, archived: 2 }
   enum purchase_card: { default: 0, other: 1 }
   enum type: { sealed_bid: 0, reverse: 1 }
 
@@ -56,6 +56,7 @@ class Auction < ActiveRecord::Base
   validate :publishing_auction, on: :update, if: :published_changed?
 
   def publishing_auction
+    return if archived?
     if published_was == 'unpublished' && purchase_card == 'default' && c2_status != 'budget_approved'
       errors.add(:c2_status, " is not budget approved.")
     end
--- a/app/models/concerns/auction_scopes.rb
+++ b/app/models/concerns/auction_scopes.rb
@@ -20,10 +20,8 @@ module AuctionScopes
     scope :pending_delivery, -> { where(delivery_status: delivery_statuses['pending_delivery']) }
     scope :not_paid, -> { where(paid_at: nil) }
     scope :paid, -> { where.not(paid_at: nil) }
-    scope :published, -> { where(published: publishes['published']) }
     scope :started_at_in_future, -> { where('started_at > ?', Time.current) }
     scope :started_at_in_past, -> { where('started_at < ?', Time.current) }
-    scope :unpublished, -> { where(published: [nil, '', publisheds['unpublished']]) }
     scope :with_bids, -> { includes(:bids) }
     scope :with_bids_and_bidders, -> { includes(:bids, :bidders) }
 
--- a/features/step_definitions/js_dialog_steps.rb
+++ b/features/step_definitions/js_dialog_steps.rb
@@ -2,3 +2,8 @@ When(/^I click OK on the javascript confirm dialog for a bid amount of (.+)$/) d
   text = "Are you sure you want to place a bid for #{amount}?"
   page.accept_alert(text)
 end
+
+When(/^I click OK on the javascript confirm dialog to archive the auction$/) do
+  text = 'Are you sure you want to archive this auction?'
+  page.accept_alert(text)
+end
--- a/app/assets/javascripts/application.js
+++ b/app/assets/javascripts/application.js
@@ -6,3 +6,4 @@
 //= require selectize
 //= require selectizer
 //= require bidConfirm.js
+//= require admin/archiveConfirm.js
--- a/app/assets/stylesheets/components/_auction-form.scss
+++ b/app/assets/stylesheets/components/_auction-form.scss
@@ -22,6 +22,10 @@
   label {
     clear: both;
   }
+
+  input[type="submit"] {
+    display: inline-block;
+  }
 }
 
 .auction_summary textarea {
--- a/app/controllers/admin/auctions_controller.rb
+++ b/app/controllers/admin/auctions_controller.rb
@@ -32,7 +32,12 @@ class Admin::AuctionsController < Admin::BaseController
 
   def update
     auction = Auction.find(params[:id])
-    update_auction = UpdateAuction.new(auction: auction, params: params, current_user: current_user)
+
+    update_auction = if ArchiveAuction.archive_submit?(params)
+                       ArchiveAuction.new(auction: auction)
+                     else
+                       UpdateAuction.new(auction: auction, params: params, current_user: current_user)
+                     end
 
     if update_auction.perform
       flash[:success] = I18n.t('controllers.admin.auctions.update.success')
--- a/app/models/admin_auction_status_presenter_factory.rb
+++ b/app/models/admin_auction_status_presenter_factory.rb
@@ -16,7 +16,9 @@ class AdminAuctionStatusPresenterFactory
   private
 
   def default_purchase_card_presenter
-    if auction.c2_status == 'not_requested'
+    if auction.archived?
+      AdminAuctionStatusPresenter::Archived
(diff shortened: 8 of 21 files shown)


db/schema.rb
```diff
@@ -43,6 +43,7 @@
     t.integer  "customer_id"
     t.integer  "c2_status",       default: 0,    null: false
     t.string   "token"
+    t.datetime "archived_at"
```
jessieay: I'm not sure we need to save this timestamp -- thinking behind it?

app/models/statistics/average_auction_length.rb
```diff
@@ -16,6 +16,6 @@ def total_time_length
   end
 
   def published_auctions
-    @_published_auctions ||= Auction.published
+    @_published_auctions ||= Auction.not_archived.published
```
jessieay: we should make sure that all of the statistics are not looking at archived auctions

app/models/concerns/auction_scopes.rb
```diff
@@ -18,12 +19,12 @@ module AuctionScopes
     scope :in_reverse_chron_order, -> { order('ended_at DESC') }
     scope :pending_acceptance, -> { where(delivery_status: delivery_statuses['pending_acceptance']) }
     scope :pending_delivery, -> { where(delivery_status: delivery_statuses['pending_delivery']) }
-    scope :not_paid, -> { where(paid_at: nil) }
-    scope :paid, -> { where.not(paid_at: nil) }
-    scope :published, -> { where(published: publishes['published']) }
+    scope :not_paid, -> { not_archived.where(paid_at: nil) }
```
jessieay: I don't think we would archive a paid auction -- this is for auctions we never have run

app/models/concerns/auction_scopes.rb
```diff
@@ -18,12 +19,12 @@ module AuctionScopes
...
+    scope :paid, -> { not_archived.where.not(paid_at: nil) }
+    # scope :published, -> { not_archived.where(published: publishes['published']) }
```
jessieay: in the grooming meeting we discussed only allowing the admin to archive unpublished actions -- this is helpful because it either means an auction was never publishd, or was published but then unpubished before it ran. either way, it could not have any bids. so I think the `published` scope will necessarily not include archived acutions

db/schema.rb
harrisj: I just assume it will come up

app/models/statistics/average_auction_length.rb
harrisj: Indeed, I noticed we don't seem to have tests for some of these?

app/models/concerns/auction_scopes.rb
harrisj: True, it looks like we can't archive a published auction

app/models/concerns/auction_scopes.rb
harrisj: These scopes are actually not used (why I commented them out and will delete) since the scopes are automagically defined elsewhere

db/schema.rb
jessieay: I would default to not adding a new DB column when we don't need it. If we added something like `papertrail`, it would keep track of all data changes in the case that we needed them later

app/models/statistics/average_auction_length.rb
jessieay: I think we do: `spec/models/statistics` ?

app/models/concerns/auction_scopes.rb
jessieay: cool! yeah, I think Rails enums give us these for free (?)

harrisj:
I do wonder since `archived` can only be applied to unpublished auctions if this should be an enum on `published` and not `delivery_status`

app/models/statistics/average_auction_length.rb
harrisj: It's okay, I don't need it now

harrisj:
I went ahead an made `archived` an enum on `published` and that allowed me to remove some other code I had added

app/assets/javascripts/admin/archiveConfirm.js
```diff
@@ -0,0 +1,5 @@
+$(document).ready(function() {
+    $('a#archive-auction-link').click(function() {
+        return confirm("Are you sure you want to archive this auction?");
```
jessieay: this looks good! I think generally I do 2 spaces rather than 4 in JS

app/controllers/admin/auction_archive_controller.rb
```diff
@@ -0,0 +1,17 @@
+class Admin::AuctionArchiveController < Admin::BaseController
+  def update
```
jessieay: thoughts on using the regular `admin/auctions#update` controller instead? I did that in my recent refactor PR and found it was helpful to reuse that rather than having a separate controller

app/models/auction.rb
```diff
@@ -56,6 +56,7 @@ class Auction < ActiveRecord::Base
   validate :publishing_auction, on: :update, if: :published_changed?
 
   def publishing_auction
+    return if archived?
```
jessieay: can we add a test for this in `spec/models/auction_spec` ?

app/models/concerns/auction_scopes.rb
```diff
@@ -20,10 +20,8 @@ module AuctionScopes
     scope :pending_delivery, -> { where(delivery_status: delivery_statuses['pending_delivery']) }
     scope :not_paid, -> { where(paid_at: nil) }
     scope :paid, -> { where.not(paid_at: nil) }
-    scope :published, -> { where(published: publishes['published']) }
```
jessieay: 👏

db/schema.rb
jessieay: Yeah we've had this issue on a few other things (he also had timestamps in copy for things like when an auction was published, which we do not record) and after talking to him I realized that he was not as concerned about the timestamp. 

We can keep if he thinks it is important, but each new db field is another thing to maintain so it would be really awesome to not add this if we dont need it

features/step_definitions/js_dialog_steps.rb
```diff
@@ -2,3 +2,8 @@
   text = "Are you sure you want to place a bid for #{amount}?"
   page.accept_alert(text)
 end
+
+When(/^I click OK on the javascript confirm dialog to archive the auction$/) do
+  text = 'Are you sure you want to archive this auction?'
+  page.accept_alert(text)
```
jessieay: nice!

app/models/auction.rb
```diff
@@ -30,7 +30,7 @@ class Auction < ActiveRecord::Base
     rejected: 2
   }
 
-  enum published: { unpublished: 0, published: 1 }
+  enum published: { unpublished: 0, published: 1, archived: 2 }
```
jessieay: cool I think this is a nice place to put it. I kind of thought we might make this a boolean field one day, but it works well as an enum now that we are using it this way!

db/schema.rb
harrisj: I understand, but I would like if we document these things in the issue formally since I missed the grooming and find myself implementing things we don't need :(

jessieay: sure - do you want to do that?

harrisj:
Tried a thing where it's an alternate submit of the form now.

localhost_3000_admin_auctions_1_edit

jessieay:
cool! what do you think of this new design (red button) @mtorres253 ?

mtorres253:
The button looks good for now

mtorres253:
We don't need the timestamp. Thanks for checking!

jessieay:
yeahhhhhhhh mergin'

PR #1327, opened 2016-09-29
Consolidate controllers

jessieay:
* Publish and unpublish auctions via `admin/auctions#update` rather than
  custom controllers
* This is pre-work for validating dates stuff. If we use a single
 controller, we can do the validations in once place


app/models/auction_parser.rb
```diff
@@ -23,13 +23,14 @@ def auction_params
       :billable_to,
       :c2_proposal_url,
       :customer_id,
+      :delivery_status,
```
harrisj: Ooops, you're right, should've realphabetized

harrisj:
LGTM

Issue #1194, opened 2016-09-08
Design cleanup: Auctions home


andrewmaier:
- See [this vignette](https://docs.google.com/document/d/1lWPYtQLdyZ42ENPIMxITcAuALXr2ADswu0XUCB7hYb4/edit#heading=h.553wlv2hs1ya) and [the prototype's homepage](https://pages.18f.gov/micropurchase/).

- “Auctions” title should be bigger, lighter weight (font-weight: 300) Change font to meriweather
- Coming soon badge, should use lowercase on “soon,” have rounded edges.
- Pagination should also better reflect the prototype styles:
![image](https://cloud.githubusercontent.com/assets/2970/18362725/f3812dbc-75d4-11e6-9d1a-905fff556e8c.png)

mtorres253:
We decided to leave it as "all" (rather than open and coming soon) in a previous grooming session. Let's revisit

jessieay:
had a ? for @mtorres253 on this in slack: https://18f.slack.com/archives/micropurchase-dev/p1475016909001041

mtorres253:
Per discussion with @jessieay we are getting rid of the "all" tab in the "Auctions" section of the admin view. We now have just "Needs attention" and "Closed". The "all" view will be collapsed with /auctions, and edit button will be viewable by admins only

PR #1331, opened 2016-09-29
Remove `admin/auctions#index`

jessieay:
* Include `edit` link on home page instead
* Addresses https://github.com/18F/micropurchase/issues/1194#issuecomment-250616246


harrisj:
LGTM

PR #1333, opened 2016-09-29
[PRODUCTION FIX] Rename Budget approved C2 presenter

jessieay:
* Was named `Approved` before, but we changed this state to be called
  "Budget approved"
* Prod drafts page is broken because the presenter is not found
  undefined method on nil error)


harrisj:
LGTM

PR #1335, opened 2016-09-30
Deploy (#1 9/30/2016)

adelevie:
Happy end of the fiscal year!!!


harrisj:
Wiating for Travis then will merge

harrisj:
The TZ bug is the only failing test. Time to see what is going on but will merge.

Issue #1316, opened 2016-09-28
Re-organize and rename mailers


jessieay:
All mailers to the winning bidder should be in `WinnerMailer` https://github.com/18F/micropurchase/pull/1284/files#r80791563

We should also rename the `auction_paid_winning_vendor_notification` to indicate that it is for the default purchase card (the "other purchase card" mailer has that in the name)

PR #1336, opened 2016-09-30
Move and rename winning bidder emails

jessieay:
* All emails to winning bidder are now in `WinningBidderMailer`
* Since that is the case, we no longer need "winning bidder" in the name
  of the mailers
* Closes https://github.com/18F/micropurchase/issues/1316


jessieay:
Seeing this rubocop failure

> Use next to skip iteration.

do you agree with this? I do not but am open to complying.

harrisj:
It's fine. Good work!

PR #1338, opened 2016-09-30
Validate auction dates when publishing

jessieay:
* Start date is in future
* End date is after start date
* Also updated presenter factory to show "ready to publish" message for
  any unpublished auction because I realized that sometimes an auction
  is no longer in the future because it's been a draft for a while, but
  we still want to show the "ready to publish" message in that case.
* Closes https://github.com/18F/micropurchase/issues/1168

### DIFF
--- a/app/models/auction.rb
+++ b/app/models/auction.rb
@@ -78,20 +78,9 @@ class Auction < ActiveRecord::Base
     end
   end
 
-  def publishing_auction
-    if published_was == 'unpublished' && purchase_card == 'default' && c2_status != 'budget_approved'
-      errors.add(:c2_status, "is not budget approved.")
-    end
-
-    if published_was == 'unpubished' && (started_at < Time.current || started_at > ended_at)
-      errors.add(:started_at, "is invalid")
-    end
-  end
-
   def lowest_bids
     bids.select { |b| b.amount == lowest_amount }.sort_by(&:created_at)
   end
-
 
   def lowest_amount
     bids.sort_by(&:amount).first.try(:amount)
--- a/app/models/auction.rb
+++ b/app/models/auction.rb
@@ -39,12 +39,10 @@ class Auction < ActiveRecord::Base
 
   validate :user_is_contracting_officer_if_above_micropurchase
   validates :delivery_due_at, presence: true
-  validates :description, presence: true, if: :published?
   validates :ended_at, presence: true
   validates :purchase_card, presence: true
   validates :start_price, presence: true
   validates :started_at, presence: true
-  validates :summary, presence: true, if: :published?
   validates :title, presence: true
   validates :user, presence: true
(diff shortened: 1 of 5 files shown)


app/models/auction.rb
```diff
@@ -53,13 +51,11 @@ class Auction < ActiveRecord::Base
...
 
-  def publishing_auction
-    if published_was == 'unpublished' && purchase_card == 'default' && c2_status != 'budget_approved'
-      errors.add(:c2_status, " is not budget approved.")
-    end
-  end
+  validates :summary, presence: true, if: :published?
+  validates :description, presence: true, if: :published?
+  validates_with PublishedAuctionValidator, on: :update, if: :published_changed?
+  # validate :publishing_auction, on: :update, if: :published_changed?
```
harrisj: Should we remove this line?

app/models/auction.rb
```diff
@@ -90,4 +78,23 @@ def user_is_contracting_officer_if_above_micropurchase
...
+    if published_was == 'unpubished' && (started_at < Time.current || started_at > ended_at)
+      errors.add(:started_at, "is invalid")
+    end
+  end
+
+  def lowest_bids
+    bids.select { |b| b.amount == lowest_amount }.sort_by(&:created_at)
+  end
+
+
```
harrisj: Can we get rid of this line?

app/models/auction.rb
```diff
@@ -90,4 +78,23 @@ def user_is_contracting_officer_if_above_micropurchase
       )
     end
   end
+
+  def publishing_auction
```
harrisj: We can get rid of this function now, right?

app/validators/published_auction_validator.rb
```diff
@@ -0,0 +1,35 @@
+class PublishedAuctionValidator < ActiveModel::Validator
+  def validate(auction)
+    if auction.published_was == 'unpublished'
```
harrisj: Skip all these if `auction.published` is set to `:archived`

app/models/auction.rb
jessieay: yes!

app/models/auction.rb
jessieay: Auction model cleaned up at 5517e34

app/validators/published_auction_validator.rb
jessieay: oook will write a test for that -- good catch!

jessieay: Added test and updated logic at e5c4882

app/models/auction.rb
jessieay: done!

jessieay:
OK @harrisj I think I've made all the requested changes. LMK what you think!

PR #1339, opened 2016-09-30
Always use day count form input for delivery deadline

jessieay:
* We used to use this just not Auctions#new form, but now we are using
  it in the edit form and on sidebar when auction is ready to publish,
  too
* This ensures that delivery deadline is always after end date
* Related to https://github.com/18F/micropurchase/issues/1168


harrisj:
Only failing test is my nemesis, so I am going to merge

PR #1342, opened 2016-09-30
C2 status string refactor

jessieay:
* This is for the status we show for draft auctions
* We were using methods on C2 presenters, but this works just as well

### DIFF
--- a/app/view_models/admin/draft_list_item.rb
+++ b/app/view_models/admin/draft_list_item.rb
@@ -15,7 +15,7 @@ class Admin::DraftListItem < Admin::BaseViewModel
 
   def c2_proposal_status
     if auction.purchase_card == 'default'
-      c2_proposal_status_presenter.status
+      I18n.t("drafts.c2_status.#{auction.c2_status}.status")
     else
       'N/A'
     end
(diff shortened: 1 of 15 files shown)


app/view_models/admin/draft_list_item.rb
```diff
@@ -15,7 +15,7 @@ def id
 
   def c2_proposal_status
     if auction.purchase_card == 'default'
-      c2_proposal_status_presenter.status
+      I18n.t("drafts.c2_status.#{auction.c2_status}.status")
```
harrisj: I was wondering why we didn't do this earlier. I had assumed we were being explicit for a reason instead of doing conventions vs. configuration

harrisj: I guess in many cases we are also doing interpolation in there too

jessieay: I am not sure why -- I think that using the C2 presenters made sense when all we had was C2 presenters. Also, I believe this was implemented before we were using I18n, so interpolation would not have been an option since we were just returning strings.

harrisj:
Only failing test is that one we need to resolve, so merging

Issue #1317, opened 2016-09-28
Update namespacing for purchase-card-specific presenter


jessieay:
`AdminAuctionStatusPresenter::Accepted` is only used for the default purchase card. We should include that in the name of the presenter like we do for `AdminAuctionStatusPresenter::OtherPcard::Accepted` -- same goes for any other presenters that should be namespaced based on the purchase card

PR #1345, opened 2016-09-30
Include pcard namespace for "accepted" presenter

jessieay:
* To be consistent with "other pcard" presenter
* Closes https://github.com/18F/micropurchase/issues/1317


harrisj:
LGTM!

PR #1346, opened 2016-09-30
Style tweaks to header

jessieay:
* Make it match prototype more
* Change "sign in" copy
* Closes https://github.com/18F/micropurchase/issues/1191

Before:

screen shot 2016-09-30 at 4 25 24 pm

After:

screen shot 2016-09-30 at 4 35 52 pm


harrisj:
LGTM

PR #1347, opened 2016-10-01
Improve onboarding docs

jessieay:
* We should also have a policy if assigning someone to respond to NR
  alerts each week so they don't just hang around.


harrisj:
LGTM!

PR #1349, opened 2016-10-03
Admins receive email when vendors finish work

### LINKED ISSUE #1266: Admin Email Notification - Winning vendor is done with work
As an admin of micropurchase, so that I don't have to go to the show page to check whether a winning vendor has started work, I want the system to notify me when a vendor has completed work via email. This email should go to as many admins as possible.

- [x] **Scenario: System sends email notification to micropurchase admins that vendor has completed work**
- Given there is a published auction
- And that auction has ended
- And that auction has a winner
- And the winner has completed work [(clicked the "I'm done" button)](https://github.com/18F/micropurchase/issues/1000)
- Then the system should send an automated email to the all micropurchase admins
- And the email content should be as follows

> From: 		micropurchase@gsa.gov
> To address: 	[micropurchase admin email alias]
> Subject: 		[auction.winner.name] delivered on auction “[auction.name]”.
> Howdy,
> [auction.winner.name] recently finished working on “[auction.name]”. You can view their work here: [pull request URL] 
> Please review [auction.winner.name]’s work with the customer, and either accept or reject it by signing in and visiting: [auction.url]

harrisj:
Fixes #1266

### DIFF
--- a/app/mailers/admin_mailer.rb
+++ b/app/mailers/admin_mailer.rb
@@ -14,4 +14,18 @@ class AdminMailer < ActionMailer::Base
       reply_to: 'micropurchase@gsa.gov'
     )
   end
+
+  def vendor_finished_work(auction:)
+    @auction = auction
+    winning_bid = WinningBid.new(@auction).find
+    winner = winning_bid.bidder
+    @winner_name = winner.name || winner.github_login
+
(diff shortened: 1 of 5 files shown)


adelevie:
LGTM 🍧

PR #1351, opened 2016-10-04
Archiving an auction should cancel in C2

### LINKED ISSUE #1303: C2 request is canceled for archived or rejected auctions
As a C2 approver, I want to not have active C2 requests for auctions that have been archived, so C2 requests should be canceled when auctions are archived

- [x] **Scenario: C2 request is cancelled when an auction is archived**
- And there is an unpublished auction that has a C2 request that is either pending (response has been received from C2), or approved (the request has been marked as approved in C2)
- When I archive that auction (#1276)
- Then any C2 request associated with that auction is canceled (in C2)
- And the C2 status should be set to "Canceled" in the micropurchase system
- And the status message on the show page  (micropurchase.18f.gov/admin/auctions/[id]) should read as specified in #1276 

- [x] **Scenario: C2 request is cancelled when an auction is rejected**
- And there is a published auction that has ended
- When I reject work for that auction (https://github.com/18F/micropurchase/issues/1015)
- Then any C2 request associated with that auction is canceled (in C2)
- And the C2 status should be set to "Canceled" in the micropurchase system
- And the status message on the show page  (micropurchase.18f.gov/admin/auctions/[id]) should be as documented here: https://github.com/18F/micropurchase/issues/1341

## Notes
 Is C2 cancelled a new C2 status? Maybe?

harrisj:
Fixes part of #1303

~~Do not merge yet, I still need to add the c2_status presenter!~~ (no need for new status messages actually)

This seems like it should work, but I still need to test out the flow manually

### DIFF
--- a/app/models/c2_cancel_attributes.rb
+++ b/app/models/c2_cancel_attributes.rb
@@ -1,0 +1,32 @@
+class C2CancelAttributes
+  def initialize(auction)
+    @auction = auction
+  end
+
+  def to_h
+    {
+      status: 'cancelled',
+      gsa18f_procurement: {
+        additional_info: additional_info
+      }
+    }
+  end
+
+  private
+
+  attr_accessor :auction
+
+  def additional_info
+    "Reason cancelled: #{reason_cancelled}"
+  end
+
+  def reason_cancelled
+    if auction.archived?
+      'the auction was archived'
+    elsif auction.rejected?
+      'the auction was rejected'
+    else
+      'unknown'
+    end
+  end
+end
--- a/app/models/c2_cancel_attributes.rb
+++ b/app/models/c2_cancel_attributes.rb
@@ -5,7 +5,7 @@ class C2CancelAttributes
 
   def to_h
     {
-      status: 'cancelled',
+      status: 'canceled',
       gsa18f_procurement: {
         additional_info: additional_info
       }
@@ -17,10 +17,10 @@ class C2CancelAttributes
   attr_accessor :auction
 
   def additional_info
-    "Reason cancelled: #{reason_cancelled}"
+    "Reason canceled: #{reason_canceled}"
   end
 
-  def reason_cancelled
+  def reason_canceled
     if auction.archived?
       'the auction was archived'
     elsif auction.rejected?
--- a/app/services/archive_auction.rb
+++ b/app/services/archive_auction.rb
@@ -8,6 +8,7 @@ class ArchiveAuction
   def perform
     if auction.unpublished?
       auction.published = :archived
+      update_c2_status
       auction.save
     else
       false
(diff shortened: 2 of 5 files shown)


app/models/c2_cancel_attributes.rb
```diff
@@ -0,0 +1,32 @@
+class C2CancelAttributes
+  def initialize(auction)
+    @auction = auction
+  end
+
+  def to_h
+    {
+      status: 'cancelled',
```
jessieay: I believe C2 spells this with one "l" (may be important for enum-related stuff)

harrisj: Ha, the source seems to be `canceled` so you are right. Filed a PR to fix the docs

jessieay: yeah it was "cancelled" at one time but the 18F content guides say to spell it "canceled" so we made the switch on C2 at some pt. Guess we forgot to update the docs!

app/services/archive_auction.rb
```diff
@@ -17,4 +18,13 @@ def perform
   def self.archive_submit?(params)
     params.key?(:archive_auction)
   end
+
+  private
+
+  def update_c2_status
+    return unless auction.c2_proposal_url.present?
```
jessieay: oof I am not a fan of this style (I know many are). I feel like this is so much clearer:

```
  if auction.c2_proposal_url.present?
    UpdateC2ProposalJob.perform_later(auction.id, 'C2CancelAttributes')
    auction.c2_status = :c2_canceled
end
```

what do you think?

app/models/c2_cancel_attributes.rb
```diff
@@ -0,0 +1,32 @@
...
+
+  def additional_info
+    "Reason canceled: #{reason_canceled}"
+  end
+
+  def reason_canceled
+    if auction.archived?
+      'the auction was archived'
+    elsif auction.rejected?
+      'the auction was rejected'
```
jessieay: should we also call this where we reject auctions, then? (great idea!)

jessieay:
This looks great! Two comments from me

app/services/archive_auction.rb
harrisj: Fixed

harrisj:
Thanks @jessieay!

jessieay:
This looks ready to merge. Did you test manually in your local env, @harrisj ? I think it would make sense to do that first in case tweaks are needed.

adelevie:
LGTM 🍏

jessieay:
Re-posting my earlier comment:

> Did you test manually in your local env, @harrisj ? I think it would make sense to do that first in case tweaks are needed.

If it works locally and on staging, let's to a prod deploy!

PR #1352, opened 2016-10-05
Rejecting auction cancels c2 proposal

### LINKED ISSUE #1303: C2 request is canceled for archived or rejected auctions

harrisj:
Merge #1351 first

Fixes #1303

### DIFF
--- a/app/controllers/admin/auction_rejections_controller.rb
+++ b/app/controllers/admin/auction_rejections_controller.rb
@@ -1,8 +1,7 @@
 class Admin::AuctionRejectionsController < Admin::BaseController
   def update
     auction = Auction.find(params[:id])
-    auction.update(delivery_status: :rejected, rejected_at: Time.current)
-    WinningBidderMailer.auction_rejected(auction: auction).deliver_later
+    RejectAuction.new(auction: auction).perform
     redirect_to admin_auction_path(auction)
   end
 end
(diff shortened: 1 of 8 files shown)


adelevie:
LGTM

PR #1364, opened 2016-10-13
Creating auction should redirect to show page

### LINKED ISSUE #1357: When new auction is successfully created, admin should go to show page
Right now, when I click the "Create" button on the https://micropurchase.18f.gov/admin/auctions/new page, and my auction is successfully created, I am taken to the "Needs Attention" page. I should be taken to the show page for that auction.

See https://github.com/18F/micropurchase/issues/931 and https://github.com/18F/micropurchase/issues/930

harrisj:
Fixes #1357

### DIFF
--- a/features/step_definitions/navigation_steps.rb
+++ b/features/step_definitions/navigation_steps.rb
@@ -106,6 +106,10 @@ Then(/^I should be on my profile page$/) do
   expect(page.current_path).to eq(profile_path)
 end
 
+Then(/^I should be on the admin auction page for that auction$/) do
+  expect(page.current_path).to eq(admin_auction_path(@auction))
+end
+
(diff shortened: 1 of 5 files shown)


features/step_definitions/navigation_steps.rb
```diff
@@ -106,6 +106,10 @@
   expect(page.current_path).to eq(profile_path)
 end
 
+Then(/^I should be on the admin auction page for that auction$/) do
```
jessieay: I am pretty sure we already have this step defined

harrisj: Actually, this is correct. IT's the "check I'm in the right place" equivalent to the "open the URL" for the admin auction show page

PR #1374, opened 2016-10-18
Add new /account pages

### LINKED ISSUE #1201: Vendor account page - includes bid history
As a vendor on micro-purchase platform, I want to go to one place to access information about my account and my bid history, so the current bid history and profile pages should be collapsed into 1 page

**Context**

- [x] **Scenario: Vendor navigates to /account/profile page**
- When I have navigated to micropurchase.18f.gov/account/profile
- Then I should see a page titled "My Account"
- And I should see a sub navigation under the title
- And the sub navigation should contain a link labeled "Profile"
- And the sub navigation should contain a link labeled "Bids placed"
- And "Profile" should be selected
- And I should see beneath the sub navigation the content that is currently on the /profile page

- [ ] **Scenario: Vendor navigates to /account/bids_placed**
- When I navigate to /account/bids_placed
- Then I should see a page titled "My Account"
- And I should see a sub navigation under the title
- And the sub navigation should contain a link labeled "Profile"
- And the sub navigation should contain a link labeled "Bids placed"
- And "Bids placed" should be selected

- [ ] **Scenario: Vendor navigates to /account/bids_placed and sees bid history**
- When I navigate to /account/bids_placed
- And I have a history of bids
- Then I should see a table
- And the table should contain a column labeled `<Column name>`
- And that column should display `<Column data>`

| Column name | Column data |
| --- | --- |
| Auction | The name of the auction that links to the auction show page (micropurchase.18f.gov/auctions/[id]) |
| Status | The Phase of that auction - Open or Closed |
| Number of Bids | The number of bids by that vendor on that auction |
| Lowest Bid | The lowest bid amount for that vendor |
| Won? | Yes or No |

- [ ] **Scenario: Vendor navigates to /account/bids_placed, and does NOT have a bid history**
- When I navigate to /account/bids_placed
- And I do NOT have a history of bids
- Then I should see the following message instead of the bid history table

> This page is blank because you’ve yet to bid on an auction. Peruse our list of coming soon and open auctions to find one that interests you. 

screen shot 2016-09-20 at 1 37 50 pm

- And "list of coming soon and open auctions" should be linked to micropurchase.18f.gov

harrisj:
Fixes #1201 

This updates a few existing controller methods:

1. `profile` is now `/account/profile` with a new look and tabbed subnavs
2. `my-bids` is now `/account/bids_placed` with a differently formatted table that summarizes bidding details for auctions rather than listing all bids.

Profile page
18f_-_micro-purchase

User has not bid
banners_and_alerts_and_18f_-_micro-purchase

User has bid
banners_and_alerts_and_18f_-_micro-purchase


spec/controllers/users_controller_spec.rb
```diff
@@ -27,6 +27,34 @@
...
+        create(:bid, bidder: other_person, auction: auction)
+        get :bids_placed, { }, user_id: current_bidder.id
+        view_model = assigns(:view_model)
+        expect(view_model).to_not be_nil
+        expect(view_model.auctions).to be_empty
+      end
+    end
+  end
+
+
```
jessieay: couple extra newlines :)

spec/controllers/users_controller_spec.rb
```diff
@@ -27,6 +27,34 @@
...
   end
 
+  describe '#bids_placed' do
+    context 'when logged in' do
+      it 'should assign auctions that current user have bidded on, presented' do
+        current_bidder = create(:user, sam_status: :sam_accepted)
+        auction = create(:auction)
+        bid = create(:bid, bidder: current_bidder, auction: auction)
+        expect(bid).to be_valid
+        get :bids_placed, { }, user_id: current_bidder.id
```
jessieay: I think it would be helpful here to have newlines between setup, exercise, and expectations (like you did in spec/view_models/user_auction_view_model_spec.rb)

app/controllers/users_controller.rb
```diff
@@ -15,4 +15,8 @@ def update
       render :edit
     end
   end
+
+  def bids_placed
```
jessieay: I am not line love with having non-RESTful routes like this. Why not map the same route we were using before (`bids#index`) to the path name that we want? As a rails dev, that routing makes more sense to me

harrisj: I just am implementing what was in the issue. I think my problem with just using `/bids` is that this is not a list of all bids, but just those scoped to the user. I personally would prefer `/account/bids` to `bids_placed` though

spec/controllers/users_controller_spec.rb
harrisj: Got it

spec/controllers/users_controller_spec.rb
harrisj: Thanks!

app/controllers/users_controller.rb
jessieay: Yeah I think the URL is fine, but the actual rails route could be more REST-ful. (so you'd map `bids_placed` to `bids#index` or `account/bids#index` in config/routes.rb)

jessieay: I think the standardized naming conventions are really helpful from a developer perspective.

harrisj: So, you think I should move this back to the `bids_controller` then and map `/account/bids` to the `bids#index` method?

jessieay: yeah I think that's as great idea! then you can still keep the same URL per the feature request

config/routes.rb
```diff
@@ -56,9 +56,9 @@
     get 'receipt', to: 'receipts#show'
   end
 
-  resources :bids, only: [:index]
   resources :users, only: [:update]
-  get 'profile', to: 'users#edit'
+  get 'account/profile', to: 'users#edit', as: 'profile'
+  get 'account/bids', to: 'bids#index', as: 'bids_placed'
```
jessieay: 👏  😂  (o' joy)

app/views/users/_subnav.html.erb
```diff
@@ -0,0 +1,8 @@
+<div class="user-subnav">
+  <%= link_to I18n.t('labels.vendor.account.tabs.profile'),
+      profile_path,
+      class: "nav-auction #{ 'active' if active == :profile }" %>
```
jessieay: I might like to see this logic in the view model in future iterations.

app/view_models/user_auction_view_model.rb
```diff
@@ -0,0 +1,59 @@
...
+    BiddingStatusPresenterFactory.new(auction).create.label
+  end
+
+  def user_bid_count
+    user_bids.count
+  end
+
+  def user_won_label
+    if user_won?
+      'Yes'
```
jessieay: would be good to put this in I18n when we get a chance

PR #1375, opened 2016-10-19
Don't display "Bids placed" tab to admins

### LINKED ISSUE #1330: Admin account page - Just shows profile information
As an admin, I won't be bidding on any auctions, so I should not see page for my bids.

- [ ] **Scenario: Admin does not see subnav on /account/profile page**
- When I am on the micropurchase site
- And I navigate to /account/profile
- Then I should just see my profile information (no subnav)

harrisj:
Fixes #1330

**Merge after #1374**


app/controllers/users_controller.rb
```diff
@@ -15,4 +15,8 @@ def update
       render :edit
     end
   end
+
+  def bids_placed
```
jessieay: I think this is an unused method that got in here via merge

features/admin_views_profile.feature
```diff
@@ -0,0 +1,13 @@
...
+  So the current bid history and profile pages should be collapsed into 1 page
+
+  Scenario: Vendor navigates to /account/profile page
+    Given I am an administrator
+    And I sign in
+    When I visit my profile page
+    Then I should see a page header labeled "My Account"
+    And the "Profile" subnav should be selected
+    And I should see a profile form with my info
+    And I should not see a "Bids placed" subnav
```
jessieay: what about I18n? I feel like we were moving toward referencing I18n strings in feature specs, right?

harrisj:
These are dependent on #1374 being merged first

harrisj:
Done, I just added an explicit view model for the Subnav that is instantiated by the other view models (it seemed cleaner to put that logic in a single view model rather than split among the two existing ones)

app/views/users/_bids_table_none.html.erb
```diff
@@ -0,0 +1,2 @@
+<%= t('labels.vendor.account.bids_placed.no_bids_html',
+    index_url: Url.new(link_text: 'current and upcoming auctions', path_name: 'root')) %>
```
jessieay: I actually think this is unused -- remove?

app/views/users/bids_placed.html.erb
```diff
@@ -0,0 +1,15 @@
+<div class="user-profile">
+  <div class="user-header-wrapper">
+    <div class="usa-grid">
+      <div class="user-header-title">
+        <h1><%= t 'labels.vendor.account.title' %></h1>
+      </div>
+      <%= render partial: 'users/subnav', locals: {active: :bids_placed} %>
```
jessieay: +1 to myself :)

app/views/bids/index.html.erb
```diff
@@ -4,7 +4,7 @@
       <div class="user-header-title">
         <h1><%= t 'labels.vendor.account.title' %></h1>
       </div>
-      <%= render partial: 'users/subnav', locals: {active: :bids_placed} %>
+      <%= render partial: 'users/subnav', locals: {subnav: @view_model.subnav_view_model} %>
```
jessieay: spaces inside curly braces, please!

jessieay:
Just a couple more comments -- there's an unused partial in there!

app/views/bids/index.html.erb
jessieay: it is? it's what we've done in the rest of this project. I am fine with changing but it is pretty common to enforce this in Ruby projs that I've seen.

jessieay:
This looks awesome. Will merge once 🍏

PR #1376, opened 2016-10-19
Change header profile link

### LINKED ISSUE #1193: Design cleanup: Header nav has link to "Account" instead of links to "Profile" and "My Bids"
Topnav for both vendors and admins should have a link labeled "Account" instead of the current "Profile" and "My Bids" links.

Clicking on the "Account" link in the header goes to /account/profile. Details of what is on that page are in #1201 and #1330

harrisj:
**Merge after #1374**

Fixes #1193


jessieay:
I can't tell if it's a GH lag or if you need to rebase but this is still showing the changes from #1374 even though that was merged

harrisj:
GH lag. I rebased

app/controllers/users_controller.rb
```diff
@@ -15,4 +15,8 @@ def update
       render :edit
     end
   end
+
+  def bids_placed
```
jessieay: I think this is unused?

app/views/users/bids_placed.html.erb
```diff
@@ -0,0 +1,15 @@
...
+    <div class="usa-grid">
+      <div class="user-header-title">
+        <h1><%= t 'labels.vendor.account.title' %></h1>
+      </div>
+      <%= render partial: 'users/subnav', locals: {active: :bids_placed} %>
+    </div>
+  </div>
+  <div class="user-body-wrapper">
+    <div class="usa-grid">
+      <%= render partial: @view_model.bids_table_partial, locals: {view_model: @view_model} %>
```
jessieay: I think we are doing "space within curly braces" style in this proj. Surprised code climate didnt pick it up!

app/views/users/bids_placed.html.erb
```diff
@@ -0,0 +1,15 @@
+<div class="user-profile">
+  <div class="user-header-wrapper">
+    <div class="usa-grid">
+      <div class="user-header-title">
+        <h1><%= t 'labels.vendor.account.title' %></h1>
+      </div>
+      <%= render partial: 'users/subnav', locals: {active: :bids_placed} %>
```
jessieay: spaces within  curly braces, please!

app/views/users/_bids_table_none.html.erb
```diff
@@ -0,0 +1,2 @@
+<%= t('labels.vendor.account.bids_placed.no_bids_html',
+    index_url: Url.new(link_text: 'current and upcoming auctions', path_name: 'root')) %>
```
jessieay: I might expect to find this link in the view model rather than here (and then we can call the method that contains this logic in the view). Would be easier to I18n-ify the text if we do that, too.

PR #1399, opened 2016-10-28
Now default P-Card auctions can be marked as paid

### LINKED ISSUE #1371: Admin can manually mark an auction as paid
As an admin, I want to manually mark an auction as paid if the C2 integration fails
- [x] **Scenario: Admin sees "Mark Auction as Paid" button for auction that has been paid**
- And there is an auction with a winner
- And that winner has been paid via 18F p-card
- And the system has not yet recognized that the vendor was paid
- When I go to the auction show page (https://micropurchase.18f.gov/admin/auctions/[id])
- Then I should see in the auction status box a button labeled "Mark as Paid"
- [x] **Scenario: Admin manually approves C2 Request**
- And there is an auction using the 18F p-card that has been paid
- When I click on the "Mark as Paid" in the auction status box
- Then the auction should be marked as paid
- And the system should stop checking to see whether that auction has been paid
- And all existing jobs should be canceled

harrisj:
Fixes #1371 

localhost_3000_admin_auctions_11

Note than when an default auction is marked as paid, it sets the paid_at time to `Time.now`, not the time that the auction might've actually been paid (because C2 API would be down in this circumstance)


adelevie:
:shaved_ice:

PR #1401, opened 2016-10-29
Fixing auction header and account header styles

mtorres253:
Made it look more like the prototype. Will do more work on it later


mtorres253:
Will work on locally and reopen later

PR #1414, opened 2016-10-31
Add Merriweather to fonts.scss

### LINKED ISSUE #1394: Two 404s for font assets on production
# These are causing some issues with the compliance scans and should be fixed because the look and feel of the site is affected

harrisj:
Fixes #1394 

At some point, I would guess we can replace many of these files with using the USWDS Rails gem, right?


jessieay:
yeah! we are using the gem. https://github.com/18F/micropurchase/blob/develop/Gemfile#L38

I think we made a fix to the gem that makes fonts work but we'd need to update the version of the gem in this project. I can look into that if you like since I am more familiar with the gem.

PR #1429, opened 2016-11-08
Add a missed_delivery status for auctions

### LINKED ISSUE #1411: Admin can mark an auction as not delivered
So that we can keep track of auctions that were not delivered vs. code that was rejected, as an admin I should be able to mark an auction as not delivered (as opposed to rejected)

- [x] **Scenario: admin sees a way to mark an auction as not delivered**
- And there is an auction that has a winner
- And the delivery deadline for that auction has passed
- And the winner has NOT yet submitted a delivery URL

> Auction is past due
> [auction.winner] has not submitted a delivery URL for this auction.

- And I should see a button labeled "Mark as not delivered"

- [x] **Scenario: admin marks an auction as not delivered**
- When I click on the "Mark as not delivered" button
- Then the auction should go into a "Not Delivered" state
- And I should see the following message

> Not delivered
> This auction has been marked as not delivered

harrisj:
Fixes #1411

This allows us to explicitly mark some auctions where delivery deadline was missed

### DIFF
--- a/app/models/admin_auction_status_presenter_factory.rb
+++ b/app/models/admin_auction_status_presenter_factory.rb
@@ -32,8 +32,12 @@ class AdminAuctionStatusPresenterFactory
       AdminAuctionStatusPresenter::Available
     elsif won? && auction.pending_delivery?
       AdminAuctionStatusPresenter::WorkNotStarted
+    elsif overdue_delivery?
+      AdminAuctionStatusPresenter::OverdueDelivery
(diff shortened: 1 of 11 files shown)


adelevie:
LGTM, merging when :green_apple:

harrisj:
So, there was another issue that was revealed by this apparently. Fixed now

adelevie:
yay! merging when :green_heart:

PR #1449, opened 2016-11-16
Fix a broken link_to that was rendering wrong

### LINKED ISSUE #1443: User object being instantiated in Closed auctions view for undelivered auctions
localhost_3000_admin_auctions_closed

harrisj:
Fixes #1443

### DIFF
--- a/app/views/admin/auctions/closed/_delivery_missed.html.erb
+++ b/app/views/admin/auctions/closed/_delivery_missed.html.erb
@@ -11,7 +11,7 @@
     <tr>
       <td><%= link_to(auction.title, admin_auction_path(auction.id)) %></td>
       <td><%= auction.ended_at %></td>
-      <td><%= link_to(auction.winning_bidder, auction.winning_bidder.name) %></td>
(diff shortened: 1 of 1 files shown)


adelevie:
merging when :green:

baccigalupi:
Sorry, didn't realize these were connected. @adelevie knows the coding standards better, so you should go with his pass.

harrisj:
No need to apologize! Our cucumber tests aren't so great at looking at page details (Capybara isn't so great either), so it's a bit annoying to test at the feature level, especially since that missed it here

baccigalupi:
@harrisj So one of the use cases for view models is to ensure that there is no logic in your view. That would include deriving a url. I know this isn't a typical Rails way, but the benefit is that you could create a unit test on the view model for this kind of a bug.

harrisj:
I hear you on the View Models, and we have used them for some things like conditional branching, but I wasn't sure about using them for URL construction. But we could make this change

harrisj:
I think this change to use ViewModels for app construction should wait for future refactoring

PR #1452, opened 2016-11-16
Remove SAM status flashes from homepage

### LINKED ISSUE #1241: Remove DUNS number validation flash message
So that I only see messages I need to see when i need to see them, as a vendor, I should only be reminded to supply a valid Duns number only when I am on the show page for an active auction and I'm qualified to bid

So basically, let's remove the DUNS number flash message for vendors on micropurchase.18f.gov

screen shot 2016-11-07 at 3 15 20 pm

harrisj:
Fixes #1241 

These are now on the auction show page if the auction is available and on the user profile edit page. In addition, I changed the `user_controllers#update` to return to the edit form on update so that I can show an updated flash there. For instance

18f_-_micro-purchase

now leads to 

18f_-_micro-purchase

I also see it when the auction is open for bidding

18f_micro-purchase_-_proactive_solution-oriented_framework

I could probably use some styling advice here

### DIFF
--- a/app/controllers/auctions_controller.rb
+++ b/app/controllers/auctions_controller.rb
@@ -7,7 +7,6 @@ class AuctionsController < ApplicationController
       current_user: current_user
     )
     @auctions = paginated_auctions
-    @auction_collection.sam_status_message_for(flash)
   end
 
   def show
@@ -17,6 +16,7 @@ class AuctionsController < ApplicationController
       current_user: current_user,
       bid_error: flash[:bid_error]
     )
+    @view_model.sam_status_message_for(flash)
   end
 
   def update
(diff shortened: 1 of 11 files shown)


app/controllers/auctions_controller.rb
```diff
@@ -17,6 +16,7 @@ def show
       current_user: current_user,
       bid_error: flash[:bid_error]
```
baccigalupi: This is a question not about this set of commits, but I have been wondering about storing messages in flash and not just dropping them down to the typical error, notice, warning. Can you explain how this is being used?

harrisj:
I think I will need someone to take this one over and fix the styling somewhat

harrisj:
I need someone to take this over

PR #1453, opened 2016-11-17
Update en.yml

mtorres253:
Adding language to the payment confirmation email we send to the vendor so they know to click on the confirmation link (rather than just reply to the email)


harrisj:
LGTM

PR #1455, opened 2016-11-18
Ruboclean

baccigalupi:
This tries to take as much of rubocop's advice as possible just to reduce noise:

* Uses rubocop's autocorrect capabilities for simplest stuff 
* Then does a pass at correcting a bunch more easy fixes manually
* Finally does one extraction from complex method to class

This does not tackle the simplification of status in the admin bid status presenter factory object. At a quick glance this feels like real complexity both in the product and the code. Although the individual presenter classes are ideal and the polymorphic process around getting to those classes is also great.

I am wondering if:
1. It might be time for an event log for bids to have more visibility over the history of a bid. Sometimes it is easier to convey the status of a thing if you can see its history (with a good ux). This is of course a product @mtorres253 and design question. On a code level separating a single state field into a series of logs means more information to work with to make the right decision. When you overwrite status repeatedly, you are losing information.
2. In the code there are many different statuses that are conflated. Those two complex methods might be better understood if they took apart the idea of status, and grouped the presenter choices by type.

### DIFF
--- a/app/models/statistics/total_savings.rb
+++ b/app/models/statistics/total_savings.rb
@@ -1,6 +1,6 @@
 class Statistics::TotalSavings
   def to_s
-    if completed_auctions.count > 0
+    if completed_auctions.count.positive?
       Currency.new(total_starting_price - total_winning_bid_amount).to_s
     else
       'n/a'
--- a/app/presenters/admin_auction_status_presenter/available.rb
+++ b/app/presenters/admin_auction_status_presenter/available.rb
@@ -4,7 +4,7 @@ class AdminAuctionStatusPresenter::Available < AdminAuctionStatusPresenter::Base
   end
 
   def body
-    if total_bids > 0
+    if total_bids.positive?
       I18n.t(
         'statuses.bid_status_presenter.available.admin.has_bids',
         end_date: end_date,
--- a/app/presenters/average.rb
+++ b/app/presenters/average.rb
@@ -8,7 +8,7 @@ class Average
   end
 
   def to_s
-    if count > 0
+    if count.positive?
       calculate
     else
       'n/a'
--- a/app/presenters/bid_status_presenter/over/vendor/winner/accepted_pending_payment_url.rb
+++ b/app/presenters/bid_status_presenter/over/vendor/winner/accepted_pending_payment_url.rb
@@ -6,7 +6,7 @@ class BidStatusPresenter::Over::Vendor::Winner::AcceptedPendingPaymentUrl < BidS
   def body
     I18n.t(
       'statuses.bid_status_presenter.over.winner.accepted_pending_payment_url.body',
-      delivery_url: auction.delivery_url,
+      delivery_url: auction.delivery_url
     )
   end
 end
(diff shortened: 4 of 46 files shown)


mtorres253:
Thanks @baccigalupi let's talk about this. I think it can be simplified and we have talked about implementing log functionality but decided it wasn't time yet. I agree that the statuses are a bit complex as we are mixing things like C2 status with auction status. We could talk about the problem and see if logging would help.

harrisj:
Thank you for doing this. I will admit that I don't agree with all of Rubocop's suggestions myself (`positive?` doesn't seem more clear than `> 0`), but sometimes the needs of the many outweigh the needs of the few. I do want to respond to your two points suggested here though:

1. We have talked about having an event log for the various states an auction goes through. Right now for instance, we have some auxiliary fields to capture information related to certain actions (like the datetime when an auction was rejected), but we are far short of anything resembling a true audit trail. We have discussed addressing this in a future release (I think the first priority was getting feature complete) by doing one of two possible things:
      1. Creating a has_many relationship of `auction_events` to each Auction that can capture an event type, datetime, and who performed the auction. These could be used to create a timeline of auction events as well.
      2. Using the [papertrail](https://github.com/airblade/paper_trail) gem to track changes to the Auction object and then reconstructing the timeline from changes to the model.
2. I know that we discussed and rejected using a state machine for determining when an auction is available for bidding vs. closed. But I will admit that I am a bit bothered that we do essentially have state machine-like flows for C2 preauthorizing auctions before it is open and for accepting/rejecting/paying the auction after it is closed, except the current state is often determined by looking at one or more distinct and related enum fields. It makes for some confusing situations. I wonder if we should simplify by sketching out a state-machine like flow for auction delivery and approval.

baccigalupi:
@harrisj A state machine is a good abstraction. The state machine gems are really problematic though. That and single table inheritance are the problems that projects get into that they never seem to get back out of. 

That said, I think there are a couple different state types in this big system. The C2 being one of them. But there is a huge and winding branch for what happens when an auction is won, vs. not won that is separate from the not-yet-available -> open -> closed flow. Each state flow is likely simple on its own, but gets unreasonably complicated when branched with other flows. So, I think more than a state machine abstraction, the app needs to separate the different states that are masquerading as a single status.

The 1.ii events is what I was thinking about. I wouldn't do the papertrail backfill though. Lost data is lost data.

harrisj:
Do you think we should just not enforce a class style at all?

baccigalupi:
Not sure, if the dev team feels OK with enforcing nested except in cases like the credentials where it does not work, then having a standard is probably good. Also, I have been thinking about that commit and that we can change the name of the class. With Rails I try to avoid classes with the same name, because Rails autoloading. With domain concepts like Bid and Auction, you can build pluralized namespaces, but have to be careful to keep the classes inside named differently. Will Rails didn't do the autoloading!

baccigalupi:
I just dropped this back to the original set of commits, without the change to nested classes.

harrisj:
Um okay... I mean, I have no strong opinions on this rule (unlike TABS vs. SPACES, where one side is clearly horribly wrong). I just was worried that you would've been forced to correct a bunch of files all over the codebase with this change.

Thank you for fixing my Swagger hack

harrisj:
LGTM. Merging

PR #1457, opened 2016-11-21
Fix bug with cloudfoundry credentials

harrisj:
Not sure why this is also picking up some recent change to develop, but here are the two things I found with the clean-up-credentials on staging

1. Rails autoloading assumed the reference to `CloudFoundry` was to a global CloudFoundry module already loaded and not to `Credentials::CloudFoundry` so it crashed when calling `CloudFoundry.new`. This wasn't fixed by simply changing code to `Credentials::CloudFoundry.new`. I had to do a few things to explicitly load and reference the CloudFoundry subclasses first.
2. For some reason, there was a _spec in the `app/services` directory which would crash when trying to require `rails_helper`


adelevie:
*@baccigalupi (sorry for the typo, Kane!)

PR #1458, opened 2016-11-21
Clean up credentials

adelevie:
Re-PRing @baccigalupi's credentials PR, but with fixes from @harrisj merged in.

PR #1461, opened 2016-11-26
Remove an errant spec in the app/services directory

harrisj:
This is causing deploys of `develop` to fail so this needs to be removed. I already identified this issue in PR #1457 which was approved and merged to fix #1458


harrisj:
Apologies for being bold, but I am going to merge this rather than waiting, since it was approved in #1457

PR #1475, opened 2016-12-09
Use Docker in Travis

adelevie:
Achieves better parity between local/test and CI environments. Also greatly simplifies the .travis.yml file (e.g. no need to comprehend complicated phantom installation bash)


afeld:
You should write a blog post about this 😃

PR #1487, opened 2017-01-04
Revert Revert (aka re-introduce) "Clean up credentials"

adelevie:
Reverts the reversion at 18F/micropurchase#1456


jessieay:
why was this reverted before?

PR #1489, opened 2017-01-09
Only existing admin users can log in with Login.gov

jessieay:
* Adds specs and logic to ensure that only admins can log in with
  login.gov
* Show error flash when non-admin user tries to sign in

### DIFF
--- a/app/models/admins.rb
+++ b/app/models/admins.rb
@@ -1,7 +1,7 @@
 class Admins
-  def verify?(uid)
-    return false if uid.nil?
-    github_ids.include?(uid.to_s)
+  def verify?(github_id)
+    return false if github_id.nil?
+    github_ids.include?(github_id.to_s)
   end
 
   def github_ids
--- a/app/models/user.rb
+++ b/app/models/user.rb
@@ -11,8 +11,13 @@ class User < ActiveRecord::Base
   enum sam_status: { duns_blank: 0, sam_accepted: 1, sam_rejected: 2, sam_pending: 3 }
 
   def self.from_saml_omniauth(auth)
-    where(uid: auth.uid).first_or_create do |user|
-      user.assign_from_auth(auth)
+    return if auth.uid.blank?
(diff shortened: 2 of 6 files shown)


jessieay:
cc @nickbristow @adelevie

app/models/user.rb
```diff
@@ -11,8 +11,13 @@ class User < ActiveRecord::Base
   enum sam_status: { duns_blank: 0, sam_accepted: 1, sam_rejected: 2, sam_pending: 3 }
 
   def self.from_saml_omniauth(auth)
```
jessieay: this method should potentially be renamed or moved into its own service class. thoughts?

app/models/admins.rb
```diff
@@ -1,7 +1,7 @@
 class Admins
-  def verify?(uid)
-    return false if uid.nil?
-    github_ids.include?(uid.to_s)
+  def verify?(github_id)
```
nickbristow: 👍

spec/requests/sso_spec.rb
```diff
@@ -5,24 +5,43 @@
...
+      asserted_attributes = saml_response.attributes.attributes.keys.map(&:to_sym)
+      expect(asserted_attributes).to match_array([:uid, :email])
+
+      post '/auth/saml/callback', SAMLResponse: saml_idp_resp
+
+      expect(response).to redirect_to('http://www.example.com/')
+    end
+  end
 
+  context 'user is not an admin' do
```
nickbristow: I think this is flipped, looks like this test `user is an admin`

app/models/user.rb
nickbristow: I like having it here, but I see the logic in moving it off of the model since it doesn't effect the objects but rather supports the model.

spec/requests/sso_spec.rb
jessieay: Right you are! Fixed at 0f60d57

PR #1493, opened 2017-01-11
Remove first / last name request from SAML oauth

jessieay:
* We get this from GitHub already
* LOA1 users will not necessarily have a name in IDP app
* Easier to just not worry about it


jessieay:
cc @nickbristow 🐶

PR #1494, opened 2017-01-11
Move states into separate tables

adelevie:
Working with @baccigalupi to pay down some technical debt by moving many of the groupings of fields on `Auction` into their own state objects. The disambiguation will make the code easier to reason about, and easier to modify as state needs change.

### DIFF
--- a/app/services/update_auction.rb
+++ b/app/services/update_auction.rb
@@ -7,6 +7,7 @@ class UpdateAuction
 
   def perform
     assign_attributes
+    create_auction_states
     update_auction_ended_job
     perform_accepted_auction_tasks
     perform_rejected_auction_tasks
@@ -16,6 +17,20 @@ class UpdateAuction
   private
 
   attr_reader :auction, :params, :current_user
+
+  def create_auction_states
+    create_published_state
+    # add more state creation here, as needed
+  end
+
+  def create_published_state
+    if parser.publishing?
+      auction_state = AuctionState.new(auction_id: auction.id,
+                                       name: 'published',
+                                       state_value: 'published')
+      auction_state.save
+    end
+  end
 
   def assign_attributes
     auction.assign_attributes(parsed_attributes)
@@ -71,8 +86,12 @@ class UpdateAuction
     parsed_attributes[:status] == 'rejected'
   end
 
+  def parser
+    @_parser ||= AuctionParser.new(params, user)
+  end
+
   def parsed_attributes
-    @_parsed_attributes ||= AuctionParser.new(params, user).attributes
+    parser.attributes
   end
 
   def user
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -1,6 +1,8 @@
 db:
    image: postgres:9
 web:
+    tty: true
+    stdin_open: true
     build: .
     env_file: .env
     environment:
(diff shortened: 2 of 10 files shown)


app/services/update_auction.rb
```diff
@@ -71,8 +86,12 @@ def auction_rejected?
     parsed_attributes[:status] == 'rejected'
   end
 
+  def parser
```
adelevie: Just noticed I need to be smarter about the memoization. I could fix this by memoizing `attributes` when `AuctionParser` is instantiated.

docker-compose.yml
```diff
@@ -1,6 +1,8 @@
 db:
    image: postgres:9
 web:
+    tty: true
```
adelevie: Useful for `pry`ing in using `docker-compose` and `docker attach`.

spec/services/update_auction_spec.rb
```diff
@@ -2,6 +2,43 @@
...
+            current_user: auction.user
+          ).perform
+        end.to change { AuctionState.count }.by(1)
+
+        auction.reload
+
+        # do we want to low-level test for the published state object, like this?
+        published_state = auction.states.find {|state| state.name == 'published'}
+        expect(published_state.state_value).to eq('published')
+
```
baccigalupi: ^^^ that seems like the important part to test. You should probably also test what happens when an auction moves from 'archived' to 'published' in that case it will already have a record, so changing the count of records should not happen.

I forgot to mention that you can put a constraint in the database on that not happening by adding a uniq index on `['auction_id', 'name']`. Then if a unique index error is thrown you revert back to updating. Solves some obscure bugs that Rails generates via their expectation that there is always just one server running in one thread.

spec/services/update_auction_spec.rb
```diff
@@ -2,6 +2,43 @@
...
+        # and/or this?:
+        # expect(auction.published?).to be true
```
baccigalupi: This doesn't seem important to me. It seems like that is the interface expansion that I was complaining about. Special methods for states instead of just setting state and reading state. Since the states are distributed now, *we will need a nice way to read states.* `auction.state_for('published') # "archived" ` ???

spec/services/update_auction_spec.rb
```diff
@@ -2,6 +2,43 @@
...
+
+        # and/or this?:
+        # published = AuctionPublishedState.new(auction).perform
+        # expect(published).to be true
+
```
baccigalupi: I would only test this as the return value if you already had plans to use it somewhere. Otherwise the maintenance is overhead.

spec/services/update_auction_spec.rb
```diff
@@ -2,6 +2,43 @@
...
+        # and/or this?:
+        # published = FindAuctionState.new(auction, name: 'published').perform
+        # expect(published).to be true
+
```
baccigalupi: Future class? I don't have a problem with a separate object for getting state ... probably good. But we are going to need an easy way to read state from an auction that is less wordy.

adelevie:
An idea for a state builder:

```
ChangeState.new(auction, published: :unpublished).perform
```

app/services/update_auction.rb
```diff
@@ -6,17 +6,36 @@ def initialize(auction:, params:, current_user:)
...
+      auction.save
+    end
   end
 
   private
 
   attr_reader :auction, :params, :current_user
 
+  def create_auction_states
+    create_published_state
```
adelevie: Let's consider changing the published state name to something like `visibility`

PR #1495, opened 2017-01-11
Updated admins sign in page to submit login.gov request

nickbristow:
Added Nick Bristow to admins
  skip authenticity token before actions on saml requests
  Added new route to handle GET callback
  Updated location of saml sign in redirect in sso_spec

### DIFF
--- a/app/controllers/saml_authentications_controller.rb
+++ b/app/controllers/saml_authentications_controller.rb
@@ -1,14 +1,16 @@
 class SamlAuthenticationsController < ApplicationController
+  skip_before_action :verify_authenticity_token
+
   def create
     user = User.from_saml_omniauth(request.env['omniauth.auth'])
 
     if user
       session[:user_id] = user.id
-      redirect_to admin_path, notice: t('omniauth_callbacks.success')
+      redirect_to admin_auctions_needs_attention_path, notice: t('omniauth_callbacks.success')
     else
       redirect_to(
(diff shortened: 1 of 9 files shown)


app/controllers/saml_authentications_controller.rb
```diff
@@ -1,14 +1,16 @@
...
 
     if user
       session[:user_id] = user.id
-      redirect_to admin_path, notice: t('omniauth_callbacks.success')
+      redirect_to admin_auctions_needs_attention_path, notice: t('omniauth_callbacks.success')
     else
       redirect_to(
         root_path,
-        error: t('omniauth_callbacks.failure', reason: 'no admin account found.')
+        notice: t('omniauth_callbacks.failure', reason: 'no admin account found')
```
jessieay: maybe as part of this PR we should also write a test where a non-admin user tries to log in with login.gov and the expectation can be that they see this message.

config/routes.rb
```diff
@@ -8,7 +8,7 @@
   get '/auth/github/callback', to: 'authentications#create'
   get '/logout', to: 'authentications#destroy'
 
-  post 'auth/saml/callback', to: 'saml_authentications#create'
+  match 'auth/saml/callback', to: 'saml_authentications#create', via: [:get, :post]
```
jessieay: 🤷‍♂️

PR #1496, opened 2017-01-12
Added single logout link for saml auth sessions

nickbristow:
Logout path is different for users who auth with saml


jessieay:
Closing this in favor of https://github.com/18F/micropurchase/pull/1497

Note that 1497 does not include a feature spec for logging out an admin. I think we should still aim to write one, but in order to get that to pass I had to change the URL in `config/secrets.yml` to `idp_slo_url: http://idp.example.com/auth/saml/logout`, which then broke our request specs. I couldn't figure out why after about an hour so my calculation is that we should move on and tackle figuring this out when there is value in doing so.

PR #1497, opened 2017-01-13
Add single logout link for saml auth sessions

jessieay:
* Logout path is different for users who auth with saml


jessieay:
We should manually check that everything is still good here before merging. Perhaps @nickbristow can do it?

PR #1500, opened 2017-01-17
Added login.gov svg

nickbristow:
18f_micro-purchase_-_admin_login
  Added login logo to login.gov link


jessieay:
@nickbristow these specs are now failing.

Also, I think this looks better but is not strictly necessary for the integration.

app/views/admin/sign_ins/show.html.erb
```diff
@@ -8,10 +8,13 @@
         '/auth/github',
         class: 'usa-button usa-button usa-button-outline' %>
     </p>
-    <div class="mb1 right-align login_gov">
-      <%= link_to 'Authorize with Login.gov',
-        '/auth/saml',
-        class: 'usa-button usa-button usa-button-outline' %>
-    </div>
+    <p>
+      <a href='/auth/saml' class='usa-button usa-button usa-button-outline'>
```
jessieay: can we make this a `link_to` rather than href? that would be more railsy. Just change what you have to

```
<%= link_to '/auth/saml`, class: 'usa-button usa-button usa-button-outline' do %>
```
and then include the image tag within.

nickbristow: Donesies

PR #1502, opened 2017-01-18
Sources github refs via https

amoose:
Sources github repos via https and will remove warnings such as this:
```
The git source `git://github.com/18f/c2-api-client-ruby.git` uses the `git` protocol, which transmits data without encryption. Disable this warning with `bundle config git.allow_insecure true`, or switch to the `https` protocol to keep your data secure.
```

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -1,4 +1,5 @@
 source 'https://rubygems.org'
+git_source(:github) { |repo_name| "https://github.com/#{repo_name}.git" }
 
 ruby '2.3.3'
 


adelevie:
Good catch!

amoose:
Something in docker-compose is not detecting the checked out repo 😕 
```
The command "docker-compose up -d" exited with 0.
1.27s$ docker-compose run web bundle exec rake
The git source https://github.com/18f/c2-api-client-ruby.git is not yet checked out. Please run `bundle install` before trying to start your application
```

PR #1505, opened 2017-02-01
Migrate micropurchase prod to govcloud

jcscottiii:
https://github.com/18F/micropurchase/pull/1504 can be merged by this one needs to wait until the DNS changes are made to point micropurchase.18f.gov to prod.

Similar to https://github.com/18F/micropurchase/pull/1504

Removes the E/W credentials entirely.
Uses the GovCloud Production credentials now


jessieay:
Hello! I just noticed that staging is now running at https://micropurchase-staging.app.cloud.gov -- this is breaking the Login.gov integration. When is the app expected to be running on the previous URL?

jcscottiii:
@jessieay it's using the CDN broker to point all incoming traffic from `https://micropurchase-staging.18f.gov` to the new internal app url `https://micropurchase-staging.app.cloud.gov`. You won't see the extra route for `18f.gov` attached to the app like you did on E/W cloud.gov. You can run `cf logs` on the app to see traffic too :)

jessieay:
should we update the login.gov integration to use the new app url (https://micropurchase-staging.app.cloud.gov) or is it going to revert back to the previous URL?

jcscottiii:
there are a few urls:
- `micropurchase-staging.18f.gov`(was the public facing url pointing to e/w app but that url points to the govcloud app, you should use this url)
- `micropurchase-staging.app.cloud.gov` (the internal url for the govcloud app)
-`micropurchase-staging.apps.cloud.gov` (e/w app. if you’re using this url, you should move to one of the other two urls because this is going away)

jcscottiii:
also i may be missing something completely as an outsider working on this app. so let's chat today if i'm missing something on how login.gov is using micropurchase :)

jcscottiii:
Just to update everyone. There's a problem when you explicitly try to hit `https://micropurchase-staging.18f.gov/admin`. Instead of redirecting to `https://micropurchase-staging.18f.gov/admin/sign_in`, it redirects to `https://micropurchase-staging.app.cloud.gov/admin/sign_in`

Here's a snapshot from the Chrome Developer Console when trying to hit `/admin`

image

@jessieay is looking into what's happening.

Also, this domain redirection only happens for this URI. not for others that i'm aware of.

PR #1506, opened 2017-02-08
Update admins.yml

drlucas:


mtorres253:
Hi @drlucas the admin role is limited to TTS employees only.

PR #1512, opened 2017-03-26
Remove quotes from example .env file

divergentdave:
It seems the values in the .env file shouldn't be quoted. (ref. https://docs.docker.com/compose/env-file/, "There is no special handling of quotation marks.") Without this PR, I get an error that `''development''` is not a recognized environment when starting the app, and with the PR, everything works.


adelevie:
:tada:

